PR #36, opened 2013-07-30
Avoid multi-threaded auto-commit

mahendra:
Note: This is done on top of #33 

This is done using shared memory objects provided by multiprocessing.

Design
* We start a commit process which waits on a <tt>multiprocessing.Queue</tt> with <tt>timeout = auto_commit_every_t</tt>
* In the main proc, when <tt>_auto_commit</tt> triggers a commit, we send a message on the queue
* The commit proc wakes up and does a commit
* If timeout is not None, then the commit proc will wake up after the specified timeout and to the commit.
* Data is shared between both procs using shared memory.
* We avoid the use of Locks completely

Also fix a bug in auto-commit for - auto_commit_every_n

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -1,11 +1,10 @@
 import logging
 import socket
 import struct
-from threading import local
 
 log = logging.getLogger("kafka")
 
-class KafkaConnection(local):
+class KafkaConnection(object):
     """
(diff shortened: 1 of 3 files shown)


dpkp:
Closing inactive PR.

PR #153, opened 2014-04-03
Threaded producer

snaury:
Turned out to be a little more involved than I anticipated, since KafkaConnection had a major bug with copy() and threads (due to subclassing threading.local). As a side effect reinit() is not mandatory anymore, client will automatically reconnect when it is used for the first time after a copy().

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -59,6 +59,9 @@ class KafkaConnection(local):
 
         self.reinit()
 
+    def __getnewargs__(self):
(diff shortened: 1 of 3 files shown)


sontek:
@snaury This doesn't merge cleanly currently.  Could you try to fix the conflicts so it can be ready for merge?

dpkp:
Thanks snaury!  vshlapakov got this to merge in PR 330.

PR #162, opened 2014-04-30
Offset enhancement for consumer

yongkun:
Offset enhancement for consumer. User can pass offset as a parameter to consumer, with the following values:  
Previous (-1): position stores in zookeeper last time; 
CurrentBeginning (-2): current beginning offset (may not be 0 due to TTL); 
PreviousOrCurrentBeginning (-3): Get previous offset firstly, if not available, use current beginning; 
Latest (-4): Start from latest, like tail;
Any other value >= 0: actual offset, raise exception if the offset is invalid.  
Yongkun Wang and KiranRaj Mariyappa

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -269,13 +269,18 @@ class KafkaClient(object):
 
             self.topic_partitions[topic] = []
             for partition, meta in partitions.items():
-                self.topic_partitions[topic].append(partition)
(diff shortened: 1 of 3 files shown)


dpkp:
I believe this is addressed in 0.9.4 / #296 with SimpleConsumer.reset_partition_offset

PR #164, opened 2014-05-06
Support for consumer-aware OffsetFetchRequest and OffsetCommitRequest

toddpalino:
The OffsetFetchRequest and OffsetCommitRequest requests to the broker, which are used for broker storage of consumer offsets (to replace Zookeeper storage of offsets) require a ConsumerMetadataRequest. Similar to the topic metadata request, this returns the coordinator for a given consumer group, which is the broker that you must send the offset fetch and commit requests to. If you do not do this, you will more often than not get an error code 16, which is NotCoordinatorForConsumer.

I've added support for the ConsumerMetadataRequest, as well as a _send_consumer_aware_request method to support getting the coordinator and then sending a request to that broker. I have updated the OffsetFetchRequest and OffsetCommitRequest to use this method instead of the  _send_broker_aware_request method. I've also added error handling for the metadata request (in the case that it fails).

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -9,7 +9,7 @@ from kafka.common import (ErrorMapping, ErrorStrings, TopicAndPartition,
                           ConnectionError, FailedPayloadsError,
                           BrokerResponseError, PartitionUnavailableError,
                           LeaderUnavailableError, CoordinatorUnavailableError,
-                          KafkaUnavailableError)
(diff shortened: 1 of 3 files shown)


toddpalino:
Closing this. It's already fixed via other commits. There's still some issues against trunk, but it should be handled separately

dpkp:
thanks, todd.

PR #179, opened 2014-06-25
Adding codec abstraction for compatibility

jerluc:
The `Codec` class has been added to support Snappy-Xerial compatibility in the Producer. Without this, Java-based consumers will be unable to decode messages produced by non-Xerial Snappy producers.


dpkp:
Closing - by default all snappy encoded messages are now xerial-compatible, so this should no longer be an issue.

PR #180, opened 2014-07-08
Added pass through parameters to the multiprocess consumer.

hasScott:
Added passthrough parameters so that the MultiProcess consumer's child SimpleConsumers can have their fetch/buffer settings changed from the defaults.


dpkp:
closing in favor of #255 and #336

PR #252, opened 2014-10-03
Allow MultiProcessConsumer to start each child consumer from current offset.

mahall:
Fixes mumrah/kafka-python#173.


sontek:
@mahall Were you still working on the test?

dpkp:
fixed via PR #356 -- thanks for helping on this issue!

PR #281, opened 2015-01-02
Added keys to compressed messages (both gzip and snappy).

meandthewallaby:
The patch applied in the PR #268 only allowed for uncompressed messages to be assigned a key. The key was ignored when using CODEC_GZIP or CODEC_SNAPPY.


wizzat:
Nice catch!

PR #283, opened 2015-01-05
Add retry and queue size on async producer

ashangit:
Add retry on async producer in case of send failure. Default set to 3
retry with a back retry set to 300 ms

Add capabilities to limit the size of the async queue in order to avoid
having too much memory consumption by this queue if the async process
is not able to send messages fastly.

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -4,9 +4,9 @@ import logging
 import time
 
 try:
-    from queue import Empty
+    from queue import Empty, Full
 except ImportError:
-    from Queue import Empty
+    from Queue import Empty, Full
 from collections import defaultdict
 from multiprocessing import Queue, Process
 
@@ -22,11 +22,17 @@ log = logging.getLogger("kafka")
 BATCH_SEND_DEFAULT_INTERVAL = 20
 BATCH_SEND_MSG_COUNT = 20
 
+BATCH_SEND_QUEUE_BUFFERING_MAX_MESSAGES = 0
+BATCH_SEND_QUEUE_MAX_WAIT = -1
+
+BATCH_SEND_MAX_RETRY = 3
+BATCH_SEND_RETRY_BACKOFF_MS = 300
+
 STOP_ASYNC_PRODUCER = -1
 
 
 def _send_upstream(queue, client, codec, batch_time, batch_size,
-                   req_acks, ack_timeout):
+                   req_acks, ack_timeout, batch_send_max_retry, batch_send_retry_backoff_ms):
     """
     Listen on the queue for a specified number of messages or till
     a specified timeout and send them upstream to the brokers in one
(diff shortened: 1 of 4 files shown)


kafka/producer/base.py
```diff
@@ -73,12 +79,16 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
...
+        for i in range(batch_send_max_retry):
+            try:
+                client.send_produce_request(reqs,
+                                            acks=req_acks,
+                                            timeout=ack_timeout)
+            except Exception:
+                log.exception("Unable to send message - retry {0}".format(i))
+                time.sleep(float(batch_send_retry_backoff_ms) / float(1000))
+                continue
+            break
```
wizzat: Shouldn't this break be in the try?

kafka/producer/base.py
```diff
@@ -188,7 +211,12 @@ def _send_messages(self, topic, partition, *msg, **kwargs):
 
         if self.async:
             for m in msg:
-                self.queue.put((TopicAndPartition(topic, partition), m, key))
+                try:
+                    self.queue.put((TopicAndPartition(topic, partition), m, key), block=True,
+                                   timeout=self.batch_send_queue_max_wait)
+                except Full:
+                    log.exception('Queue full, failed to put message "{0}" in the async queue after {1} ms'.format(
+                        m, self.batch_send_queue_max_wait))
```
wizzat: The timeout is in seconds, not milliseconds.

kafka/producer/base.py
ashangit: In the try or out the try do the same.
We continue if we face an exception or we break if no exception.

kafka/producer/base.py
ashangit: In fact the timeout on the message insertion in the async queue is in seconds.
Only the backoff for retry is in milliseconds.

wizzat: Are you sure this is doing what you think it is then?

ashangit: Here are the behaviour expected:
If batch_send_queue_buffering_max_messages and batch_send_queue_max_wait are not set then the queue is an unlimited sized queue and put in the queue never wait or fail for Full exception.

If only batch_send_queue_max_wait is set, it won't change anything as the queue is unlimited in size.

If only batch_send_queue_buffering_max_messages is set to a value then the queue is limited in size. Put of messages will raise a Full exception if the queue is full as the default value of batch_send_queue_max_wait will induce no wait despite the parameter block set to True.

If both parameter are set the queue will be limited in size. Put of messages will wait the time set in batch_send_queue_max_wait (in second). If the queue is still full after this time an exception Full will be raised otherwise the put will succeed.

Let me know if you think that there is an issue on the code or if you think that this implementation is not appropriate.

toli:
What's the status of this PR? it's been 4+ months, same with https://github.com/mumrah/kafka-python/pull/304
We are facing a similar problem at at work, and i'd prefer not to have custom patches and just use vanilla upstream code...

@dpkp - you mentioned something about Rdio having their own, perhaps you can get that committed to upstream? or get this patch in?

dpkp:
I'd like to get #331 merged for this issue instead. If you have a second to review, would appreciate more eyeballs and/or testing.

dpkp:
#331 is merged and supports this via `async_queue_maxsize` and `async_queue_put_timeout` configuration parameters to the producer

PR #286, opened 2015-01-09
KafkaConsumer should be in __all__.

Dinoshauer:
Hey, just noticed that the new ``KafkaConsumer``is not in ``__all__`` making it hard to import. This PR fixes it :)


dpkp:
Thanks!  TravisCI failure seems like a flapper -- I kicked the test, and will merge when it passes.

Dinoshauer:
Grand! I was wondering if there was any estimated date on when the new version with this consumer is going to be released?

dpkp:
Explicit imports should work without `__all__`:
`from kafka import KafkaConsumer`
as well as
`from kafka.consumer.kafka import KafkaConsumer` 

This PR will include the KafkaConsumer in `from kafka import *` .  I'm not a big fan of `*` imports, but I acknowledge that many folks find it useful.

dpkp:
Would love to cut a new release as soon as possible. there's lots of new bits since 0.9.2

PR #288, opened 2015-01-09
Randomize start by default for SimpleProducer

alexcb:
random_start=False is dangerous. Any client that initializes a
SimpleProducer and performs only a single publish batch will always go
to partition 0. A common use-case for this is command line utilities,
or web-servers which are unable to cache the SimpleProducer instance
between calls. This would result in partition 0 becoming overloaded, and all other partitions getting starved.


dpkp:
Agree.

PR #289, opened 2015-01-10
Use reflection to avoid multiple errno definitions

alexcb:
I was searching through modifications between 0.9.0 and 0.9.1 in order to upgrade our deployed software and came across the same bug that @dpkp discovered with the check_error() method:

https://github.com/mumrah/kafka-python/commit/0d57c2718fcf3819f2c18911126f245e9e9ce3e0#diff-9e97caff580d80c9a0eb85fe22ffb3d1R184

However the fix involved creating a NoError exception which seems a bit akward. Here's a alternative version.

Cheers,
Alex

### DIFF
--- a/kafka/common.py
+++ b/kafka/common.py
@@ -1,3 +1,5 @@
+import inspect
+import sys
 from collections import namedtuple
 
 ###############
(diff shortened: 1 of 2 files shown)


dpkp:
I like your solution better!

Issue #263, opened 2014-11-06
OffsetOutOfRange Error


jimmyislive:
I initially posted this to the kafka users group, but they suggested to post it here as I am using the python client:

I recently received a OffsetOutOfRange error in production while it was functioning properly for like 4-5 days.

In the kafka broker log:

[2014-11-03 21:39:25,658] ERROR [KafkaApi-8] Error when processing fetch request for partition [activity.stream,5] offset 7475239 from consumer with correlation id 69 (kafka.server.KafkaApis)
kafka.common.OffsetOutOfRangeException: Request for offset 7475239 but we only have log segments in the range 8014610 to 10227768.

And on the client side I saw:

Nov 03 21:39:25 INFO kafka consumer.py: Commit offset 10227769 in SimpleConsumer: group=ActivityStream, topic=activity.stream, partition=5
...
...
Nov 03 21:39:26 ERROR demux.consumer_stream consumer_stream.py: OffsetOutOfRangeError(FetchResponse(topic='activity.stream', partition=5, error=1, highwaterMark=-1, messages=<generator object _decode_message_set_iter at 0x3dcf870>),)

Why would the kafka client commit offset 10227769, and then, 1 second later, turn around and ask kafka for the offset 7475239?

We are using a fork with some minor modifications of the kafka-python client.

realmeh:
MultiProcessConsumer really need offsets fix in OffsetOutOfRangeError problem, hopes

dpkp:
not sure what the cause is, but #296 and #356 include some possible fixes to underlying issues. the fork you are using has a lot of tweaks in it -- consumer rebalancing issues could be a cause (I see you are using mahendra's zookeeper patch!).  if you can reproduce w/ a released kafka-python client, please re-open.

PR #296, opened 2015-01-16
Validate consumer offset in SimpleConsumer 

ecanzonieri:
I slightly changed the SimpleConsumer interface adding the default option use_latest_offsets. Resetting the offsets automatically is especially useful in MultiprocessConsumer, where an explicit call to the method seek is not possible. 

I manually tested it in my kafka cluster, but there are no integration tests for this feature, yet.
I think that tuning "log.retention.bytes" I should be able to implement two tests. 
Fix for #263


jobevers:
:+1:

ecanzonieri:
I changed the code to be more like KafkaConsumer. It now uses the keyword argument auto_offset_reset, which can be either "smallest" or "largest" (default), and falls back to the old behavior otherwise.

farvour:
Can we get this merged in?

dpkp:
LGTM

PR #302, opened 2015-01-22
Add failed_payloads attribute to FailedPayloadsError

jobevers:
The logs and stack trace can get quite large if a payload is large, or there are a lot of them.  Moved it  to be its own attribute.

### DIFF
--- a/kafka/common.py
+++ b/kafka/common.py
@@ -161,7 +161,9 @@ class KafkaTimeoutError(KafkaError):
 
 
 class FailedPayloadsError(KafkaError):
-    pass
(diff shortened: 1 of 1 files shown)


kafka/common.py
```diff
@@ -161,7 +161,9 @@ class KafkaTimeoutError(KafkaError):
 
 
 class FailedPayloadsError(KafkaError):
-    pass
+    def __init__(self, failed_payloads, *args):
+        KafkaError.__init__(self, *args)
```
dpkp: pls use `super` here

dpkp:
Merged w/ change to super()

PR #303, opened 2015-01-23
Pass through keyword arguments to send_produce_request

jobevers:
Its currently not possible to set the `fail_on_error` keyword argument in `client.send_produce_request` from the Producer.


dpkp:
Sorry to take a while to get to this one. I opened issue #395 for this issue. I dont want to merge this PR as is because I would prefer not to expose the internal client api kwargs directly to the consumer, and also it makes unit testing a bit more difficult.

Instead I put up #396 to support fail_on_error configuration. Because this affects the I intend to merge that and include in the forthcoming 0.9.4 release. If you want/need better callback support (the other internal producer kwarg), I think it is worth opening a separate issue.

Issue #297, opened 2015-01-19
async producer has unbounded internal queue


milyenpabo:
I'm using the provided example code for SimpleProducer to send out 1 kB messages in an infinite loop, as fast as possible. What I see is that I can pass ~60k messages/sec to the kafka-python library, although only ~1k messages are actually sent out towards the kafka brokers. In the meantime, resident memory usage of the producer is going up steadily and without an upper bound.

I'm new to kafka-python, and I'm not familiar with its iternals, but it seems as if the communication between my client code and kafka-python was async (regardless of instantiating the SimpleProcuder with async=False), and messages were queued internally by the kafka-python library in an unlimited queue.

Is that intutition right? Is there a way to use the library in a kind of "blocking" mode where memory usage is limited?

I'm using kafka-python v0.9.2 with two v0.8.2 kafka brokers, all nodes are connected with a 10 Gbps network.

dpkp:
this sounds odd -- the SimpleProducer should operate in a sync / blocking mode by default.  can you paste a test case to reproduce?

toidi:
I confirm that memory leak in producer with `batch_send=True`.

toidi:
I guess this happens because growing queue.
We have to add queue maxsize parameter.

milyenpabo:
Thanks for the quick patch!

Though it might be that from a client point of view, blocking on send would be more desirable than raising an error. What do you think?

toidi:
Kafka is about a big data.

If you can not send your data to kafka it is better to spill data to disk and try to resend later.
In my opinion there are more cases where no time to wait.

dpkp:
there is currently no bounds on the async producer's internal queue size.  setting either `async=True` or `batch_send=True` uses async mode and will not block but instead add messages to the internal queue.  If you add messages to the producer's queue faster than it can send them over the network to kafka then you will see memory increasing as the unsent queue grows.

honestly I would not recommend the current async producer code for production use -- it needs to be fixed for several issues (this one included).  at my company we wrote our own async producer class that wraps the sync simpleproducer.  I will see if we can get that merged back upstream.

wizzat:
FWIW, async and multiprocessing both have a host of issues. I'm still
firmly convinced that the multiprocessing consumer is worse for performance
in every way possible.

dpkp:
[...]

PR #304, opened 2015-01-23
Add producer batch send queue size limit

toidi:
As noted in #297 unlimited queue size can cause memory leak.

My patches add ability to specify queue size.
By default queue size is unlimited.

If queue become full next send message will raise an `BatchQueueOverfilledError`.

### DIFF
--- a/kafka/common.py
+++ b/kafka/common.py
@@ -200,6 +200,10 @@ class KafkaConfigurationError(KafkaError):
     pass
 
 
+class BatchQueueOverfilledError(KafkaError):
+    pass
+
+
 def _iter_broker_errors():
     for name, obj in inspect.getmembers(sys.modules[__name__]):
         if inspect.isclass(obj) and issubclass(obj, BrokerResponseError) and obj != BrokerResponseError:
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -4,15 +4,16 @@ import logging
 import time
 
 try:
-    from queue import Empty
+    from queue import Empty, Full
 except ImportError:
-    from Queue import Empty
+    from Queue import Empty, Full
 from collections import defaultdict
 from multiprocessing import Queue, Process
 
 import six
 
 from kafka.common import (
+    BatchQueueOverfilledError,
     ProduceRequest, TopicAndPartition, UnsupportedCodecError
 )
 from kafka.protocol import CODEC_NONE, ALL_CODECS, create_message_set
@@ -21,6 +22,7 @@ log = logging.getLogger("kafka")
 
 BATCH_SEND_DEFAULT_INTERVAL = 20
 BATCH_SEND_MSG_COUNT = 20
+BATCH_SEND_QUEUE_MAXSIZE = 0
 
 STOP_ASYNC_PRODUCER = -1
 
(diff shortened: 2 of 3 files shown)


kafka/producer/base.py
```diff
@@ -139,7 +143,8 @@ def __init__(self, client, async=False,
             log.warning("async producer does not guarantee message delivery!")
             log.warning("Current implementation does not retry Failed messages")
             log.warning("Use at your own risk! (or help improve with a PR!)")
-            self.queue = Queue()  # Messages are sent through this queue
+            # Messages are sent through this queue
+            self.queue = Queue(maxsize=batch_send_queue_maxsize)
```
dpkp: this is used both in batch mode as well as in simple async mode -- the variable name probably should not be specific to batch mode

test/test_producer.py
```diff
@@ -25,3 +26,16 @@ def test_producer_message_types(self):
             # This should not raise an exception
             producer.send_messages(topic, partition, m)
 
+    @patch('kafka.producer.base.Process')
+    def test_producer_batch_send_queue_overfilled(self, process_mock):
```
dpkp: should also test `async=True` mode

toidi:
@dpkp how producer should behave with async mode enabled?

dpkp:
should be the same behavior, i just think that the max queue size variable should probably be labeled as an `async_` kwarg and applied in the same way whether async mode is enabled directly via `async=True` or indirectly via `batch_send=True`

kafka/common.py
```diff
@@ -200,6 +200,10 @@ class KafkaConfigurationError(KafkaError):
     pass
 
 
+class BatchQueueOverfilledError(KafkaError):
```
dpkp: How about `AsyncQueueFull`

kafka/producer/base.py
```diff
@@ -188,7 +194,13 @@ def _send_messages(self, topic, partition, *msg, **kwargs):
 
         if self.async:
             for m in msg:
-                self.queue.put((TopicAndPartition(topic, partition), m, key))
+                try:
+                    item = (TopicAndPartition(topic, partition), m, key)
+                    self.queue.put_nowait(item)
+                except Full:
+                    raise BatchQueueOverfilledError(
+                        'Producer batch send queue overfilled. '
```
dpkp: s/batch/async/

dpkp:
consider whether PR 283 is a more robust approach (includes backoff + retry)

rogaha:
+1

rogaha:
What is the status of this PR? Is there anything I can help you guys with in order to get this merged?

rogaha:
ping @dpkp

toli:
What's the status of this PR? it's been 4+ months, same with https://github.com/mumrah/kafka-python/pull/283
We are facing a similar problem at at work, and i'd prefer not to have custom patches and just use vanilla upstream code...

@dpkp - you mentioned something about Rdio having their own, perhaps you can get that committed to upstream? or get this patch in?

dpkp:
I'd like to get #331 merged for this issue instead. If you have a second to review, would appreciate more eyeballs and/or testing.

dpkp:
#331 is merged and supports this via `async_queue_maxsize` and `async_queue_put_timeout` configuration parameters to the producer

PR #309, opened 2015-02-02
fix circle reference

zhaopengzp:
circle reference between SimpleConsumer and ReentrantTimer .
objgraph-icqmvd


wizzat:
Nice, thanks!

PR #316, opened 2015-02-09
Allow better usage of docs locally.

sontek:
This will default to the rtd theme and introduces
tox -e docs for generating docs locally.


sontek:
Tests are failing because we haven't merged in:

https://github.com/mumrah/kafka-python/pull/314

For the pypy fix.

sontek:
You can see how this looks here:

http://sontekkafka-python.readthedocs.org/en/add_docs_to_tox/apidoc/modules.html

sontek:
This also fixes: #318

http://sontekkafka-python.readthedocs.org/en/add_docs_to_tox/apidoc/kafka.html#module-kafka.client

Issue #323, opened 2015-02-16
Due to their daemonized nature, producers are never destroyed


chmduquesne:
Hi,

I am writing a Flask app to push/pull messages to Kafka.

If you are not familiar with Flask, it is threaded by default, and objects are usually only valid during one request.

In my app:
- a post to `/<topic>/[<key>/]` will push the posted content on kafka to the topic `<topic>` (possibly with the key `<key>` if provided)
- a get on `/<topic>/[<group>/]` will retrieve the last 20 messages on kafka for the topic `<topic>` (possibly from the group `<group>`, otherwise with the default group name `flasfka`) if provided.

The relevant code is here: https://github.com/chmduquesne/kafka-http/blob/master/flasfka/api.py (this url may change, I will update the issue if it does)

This code will create a new producer for each request, but it seems like the kafka objects continue running forever even after the request is over. The problem is obvious when looking at the logs of the application. I launched it, then ran:

    seq 1 20 | xargs -I % -n 1 curl -X POST --data-binary "test%" http://127.0.0.1:5000/my-topic/
    watch -n 5 curl -s http://127.0.0.1:5000/my-topic/

Here is what happens when I kill the application:

     * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
    No handlers could be found for logger "kafka"
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
    127.0.0.1 - - [16/Feb/2015 14:09:09] "POST /my-topic/ HTTP/1.1" 204 -
[...]

PR #324, opened 2015-02-17
Properly destroying the objects that contain daemonized threads

chmduquesne:
Consumer, Producer and ReentrantTimer all create daemonized threads. Problem: when destroying these objects, the daemonized threads are not destroyed and the threads live forever with no mean to reach them.

This created issue #323, which I could deal with by using stop() manually, but there is no reason to force the user to force the user to do that when it could be dealt with automatically.

This pull request fixes this.

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -127,6 +127,7 @@ class Producer(object):
         self.async = async
         self.req_acks = req_acks
         self.ack_timeout = ack_timeout
+        self.stopped = False
 
         if codec is None:
             codec = CODEC_NONE
@@ -212,3 +213,8 @@ class Producer(object):
 
             if self.proc.is_alive():
                 self.proc.terminate()
+        self.stopped = True
+
+    def __del__(self):
+        if not self.stopped:
+            self.stop()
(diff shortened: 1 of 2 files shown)


dpkp:
can you submit pep8 fixes / ae0017cb0172b633a56084be6e4779fc0bfedac2 separately?

chmduquesne:
Yes.

Actually I did this commit to trigger a travis build on my clone and check
that the build was passing before doing a pull request (I don't know how
else to do).

I will remove the pep8ification.

chmduquesne:
I have removed the pep8 fix, and rebased my patch onto master, so that this can be merged with a fast forward.

Travis reports an error on this pull request, but I don't understand why: on my clone, [it does not complain](https://travis-ci.org/chmduquesne/kafka-python/builds/52819116). Can you relaunch the [incriminated build](https://travis-ci.org/mumrah/kafka-python/jobs/52819175)?

dpkp:
not sure what is going on w/ tests -- `test_batched_simple_producer__triggers_by_message` may need some investigation

Issue #322, opened 2015-02-11
SimpleConsumer and KafkaConsumer may fetch and return committed messages if compression is enabled


ecanzonieri:
Looking at the Fetch API on the [Kafka documentation](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchAPI) it seems that when compression is enabled fetch request may return messages with offsets smaller than the FetchOffset in the request.
It is up to the client library to filter those messages out.
It looks like both SimpleConsumer and KafkaConsumer do not check if the offset in the message is actually higher than the one in the request. So they may return duplicate messages.

dpkp:
good catch.

dpkp:
This is fixed in master and will go out in the next release.

PR #328, opened 2015-02-23
Fixes consumer/kafka and consumer/simple to only yield messages if the m...

wkiser:
Addresses issue #322 
Fixes consumer/kafka and consumer/simple to only yield messages if the message's offset is greater than or equal to the consumer offset.


dpkp:
wkiser: looks like you pushed a few additional unrelated commits to this branch `73ad381` and `0ed3795` -- can you remove?

not sure what happened on tests initially.  i restarted the ones that failed...

wkiser:
Sorry about that. I was doing some version hacking to get this package working with our build system. Didn't realize that I had pushed those commits.

PR #329, opened 2015-02-25
Correct message keys for async batching mode

vshlapakov:
There're some possible changes relative to #325.

As create_message_set() method is called only from base Producer class and from _send_upstream(), I've changed it a little bit, so messages sent as a list of (msg, key) tuples. I've added some basic tests also.

Take a look please, probably it can be improved.
(This PR is squashed to the other branch #326.)

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -62,7 +62,7 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
             # Adjust the timeout to match the remaining period
             count -= 1
             timeout = send_at - time.time()
-            msgset[topic_partition].append(msg)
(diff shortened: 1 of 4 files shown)


dpkp:
This seems like a reasonable approach.  Another possibility would be to use a namedtuple.  Will take a closer look in the next few days.  Would like to get this issue fixed before next release!

vshlapakov:
@dpkp Indeed, it's an important fix, I can confirm that enabled Kafka log compaction will delete all messages with same keys at the topic, and it's an irreversible loss of data. I can change it with `namedtuple` (though don't see a big difference), just let me know if it's necessarily.

PR #331, opened 2015-02-25
Async producer retries for failed messages

vshlapakov:
Another approach to add retries for failed messages (for asynchronous mode only).

1. I've extended `ProduceRequest` a little bit by adding complete retries count to the fields.
It's backward compatible, you can miss this param when creating initial ProduceRequest, it's 0 initially.
(It's not very clean solution in terms of code, but storing retries count per request sounds reasonable.)

2. I've added several retry params for producers also:
  * `batch_retry_backoff_ms` - backoff timeout to wait before next retry (300 by default);
  * `batch_retries_limit` - upper limit of retries per request 
  (`5` by default, `None` means infinite retries, `0` means no retries).

3. I've used `FailedPayloadsException` from #55 to get failed requests and increment retries for it, or drop it depending on settings. 

What do you think about the approach? I'm going to add some additional tests for it asap.


dpkp:
There are several exceptions that could be raised within the send_produce_request() call chain besides FailedPayloadsError:

`NotLeaderForPartitionError` -- metadata is likely out of date and should be refreshed before retry
`LeaderNotAvailableError` -- leader election is underway, backoff + refresh metadata before retry
`UnknownTopicOrPartitionError` -- broker is neither a leader or a replica, possible that a metadata refresh and retry would fix, but also possible that topic just doesnt exist and auto-topic-creation is not enabled.
`ConnectionError` -- socket / connection error, backoff + refresh metadata + retry
`KafkaUnavailableError` -- all broker nodes are offline...  backoff
`RequestTimedOutError` -- broker was unable to get all ACKs specified by the Produce request w/i the required timeout.  If the user is ok w/ possible duplicate messages, a retry is fine.
Then also a bunch of message related errors: `InvalidMessage`, `InvalidMessageSize`, `MessageSizeTooLarge`  -- which generally should not be retried

vshlapakov:
Thanks a lot for the clarifications, it seems that we need more smart retrying logic, I'll think about it. So far I've updated the code a little bit to retry all failed requests without exclusions.

dpkp:
I dont think we can do blanket retries.  Many of the errors require resyncing cluster metadata before retry.  The error handling logic should handle that.

vshlapakov:
Sorry, removed/restored the branch by accident, I'm on this issue.

vshlapakov:
As this PR is directly related with async producer and retries, I also added an ability to limit async queue size (used some commits from #283 & #304 - look like abandoned).

vshlapakov:
@dpkp I'm still having some odd errors from Travis, but could you check the approach please? I'm not sure if we're fine with having a named tuple as an argument for a producer.

kafka/producer/base.py
```diff
@@ -71,12 +84,48 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
...
+
+        reqs_to_retry, error_type = [], None
         try:
             client.send_produce_request(reqs,
                                         acks=req_acks,
                                         timeout=ack_timeout)
-        except Exception:
-            log.exception("Unable to send message")
+
+        except FailedPayloadsError as ex:
```
dpkp: now that #366 is merged, we can pass `fail_on_error=False` to `client.send_produce_request` and get a list of responses/errors. This would allow us to loop through the responses and only retry the reqs that failed. Currently this exception would cause all reqs in the batch to retry, even if only a single req caused the error.

kafka/producer/base.py
```diff
@@ -71,12 +84,48 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
...
+        # doing backoff before next retry
+        if error_type in RETRY_BACKOFF_ERROR_TYPES and retry_options.backoff_ms:
+            log.info("Doing backoff for %s(ms)." % retry_options.backoff_ms)
+            time.sleep(float(retry_options.backoff_ms) / 1000)
+
+        # refresh topic metadata before next retry
+        if error_type in RETRY_REFRESH_ERROR_TYPES:
+            client.load_metadata_for_topics()
+
+        reqs = [req._replace(retries=req.retries+1)
```
dpkp: this seems a little clunky -- couldn't we track the retries count in an internal dict with the namedtuple as the key?

kafka/producer/base.py
```diff
@@ -71,12 +84,48 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
...
+
+        except Exception as ex:
+            error_type = type(ex)
+            if type(ex) in RETRY_ERROR_TYPES:
+                reqs_to_retry = reqs
+
+        finally:
+            reqs = []
+
+        if not reqs_to_retry or retry_options.limit == 0:
```
dpkp: I dont think we want a retry_options check here to prevent us from refreshing metadata on a refresh_error

kafka/producer/base.py
```diff
@@ -71,12 +84,48 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
...
+            log.info("Doing backoff for %s(ms)." % retry_options.backoff_ms)
+            time.sleep(float(retry_options.backoff_ms) / 1000)
+
+        # refresh topic metadata before next retry
+        if error_type in RETRY_REFRESH_ERROR_TYPES:
+            client.load_metadata_for_topics()
+
+        reqs = [req._replace(retries=req.retries+1)
+                for req in reqs_to_retry
+                if not retry_options.limit or
```
dpkp: should a user be able to configure unlimited retries ?

kafka/producer/base.py
vshlapakov: Yeah, that sounds good, I'll fix it.

kafka/producer/base.py
```diff
@@ -111,12 +160,16 @@ def __init__(self, client, async=False,
                  codec=None,
                  batch_send=False,
                  batch_send_every_n=BATCH_SEND_MSG_COUNT,
-                 batch_send_every_t=BATCH_SEND_DEFAULT_INTERVAL):
+                 batch_send_every_t=BATCH_SEND_DEFAULT_INTERVAL,
+                 async_retry_options=ASYNC_RETRY_OPTIONS,
```
dpkp: i think separate kwargs for each option (limit, backoff_ms, and retry_on_timeouts) might be a bit easier for users

kafka/producer/base.py
vshlapakov: Not sure, probably it's an over-feature, I can't imagine good use-cases.

kafka/producer/base.py
vshlapakov: Right, agree.

kafka/producer/base.py
```diff
@@ -200,7 +253,16 @@ def _send_messages(self, topic, partition, *msg, **kwargs):
...
             for m in msg:
-                self.queue.put((TopicAndPartition(topic, partition), m, key))
+                try:
+                    item = (TopicAndPartition(topic, partition), m, key)
+                    if self.async_queue_put_timeout == 0:
+                        self.queue.put_nowait(item)
+                    else:
+                        self.queue.put(item, True, self.async_queue_put_timeout)
+                except Full:
+                    raise AsyncProducerQueueFull(
```
dpkp: we'll need to have a way to let the user know which messages were successfully added to the queue, and which were not (because we loop `for m in msg` it is possible that the first n messages are added, but msg n+1 raises the exception.

kafka/producer/base.py
vshlapakov: I don't get used to use so many args in a constructor, it scares me a little bit:)
But well, yes, it should be easier for a finite user.

kafka/producer/base.py
vshlapakov: Ok, thanks for pointing me, I missed that PR!

kafka/producer/base.py
vshlapakov: Good catch! My proposal is tracking index `n` and return `msg[n+1:]` as `failed_msgs` property's value of `AsyncProducerQueueFull` exception.

kafka/producer/base.py
dpkp: i agree that lots of kwargs gets messy, but a typical user will generally only want/need to set a small set of the all available kwargs. One way to make large kwargs lists easier to maintain is to keep a module or class level dictionary of defaults, and just put `__init__(client, **kwargs)`. Then in the constructor you could loop through the dictionary of defaults, checking for an explicit init value in `kwargs` . This is how the KafkaConsumer constructor manages kwargs options.

kafka/producer/base.py
dpkp: I think that would work

kafka/producer/base.py
vshlapakov: That's a good approach. Ok, sounds reasonable!

vshlapakov:
@dpkp Thanks for all your comments, they're extremely helpful, I'll try to fix everything asap.

dpkp:
thank you for digging into this! I think it's getting really close

test/test_producer.py
```diff
@@ -40,3 +51,147 @@ def partitions(topic):
...
+        producer = Producer(MagicMock(), batch_send=True,
+                            async_queue_maxsize=queue_size)
+
+        topic = b'test-topic'
+        partition = 0
+        message = b'test-message'
+
+        with self.assertRaises(AsyncProducerQueueFull):
+            message_list = [message] * (queue_size + 1)
+            producer.send_messages(topic, partition, *message_list)
```
rogaha: Here you could check the length of the queue, make sure to mock the ``_send_upstream`` method to avoid draining the queue before the ``assert`` is executed.

test/test_producer.py
```diff
@@ -40,3 +51,147 @@ def partitions(topic):
...
+        producer = Producer(MagicMock(), async=True,
+                            async_queue_maxsize=queue_size)
+
+        topic = b'test-topic'
+        partition = 0
+        message = b'test-message'
+
+        with self.assertRaises(AsyncProducerQueueFull):
+            message_list = [message] * (queue_size + 1)
+            producer.send_messages(topic, partition, *message_list)
```
rogaha: same here

rogaha:
@vshlapakov I had two minor comments, once you address @dpkp comments, LGTM.

rogaha:
@dpkp I tested this PR against our current code and it worked! :+1:

vshlapakov:
@rogaha Thanks a lot for the feedback and comments! This PR works, but it should be tuned a bit, I'm going to work on it tomorrow.

rogaha:
Welcome @vshlapakov! Thank you for submitting this PR, can't wait to see that merged to upstream! :)

test/test_producer.py
toli: @vshlapakov you can see how we did that in our fork at https://github.com/docker-hub/kafka-python/blob/master/test/test_producer.py
It seemed like verifying that 2 messages went through called for a race condition, so we mocked-out the `_send_upstream` call and drained the queue so the thread wouldn't hang around. 
That may help in case you start seeing spurious errors.

vshlapakov: Yeah, makes sense, thanks for the hint, I'll add it! @toli @rogaha

kafka/producer/base.py
vshlapakov: Mm, these namedtuples contain lists of messages, and ```tuples can be used as a dicti keys if they contain only strings, numbers, or tuples```, so thinking about other approach; it's the last issue that I should fix so far.

vshlapakov: As `payloads` is used as an iterable, `ProduceRequest` could be created with `payloads=tuple(messages)`, it should work completely the same (though it seems like a cheat).

vshlapakov:
Well, I think I've fixed all main issues, take a look please!

rogaha:
LGTM

toli:
LGTM, thanks Viktor!

dpkp:
Added my few remaining comments: need to check response.error, should update params in docstrings, and should add a few tests on response errors. Looking great!

vshlapakov:
Great, thanks for meaningful comments, I'll finish it!

dpkp:
@vshlapakov -- any progress? I can jump in on this and close out if needed. Let me know!

vshlapakov:
My apologies @dpkp !! I was really going to finish it, but had a lot of important events last week. I'll have some free time at the beginning of the next week, but if you'd like to close it sooner, you're welcome and sorry for the delay.

vshlapakov:
Rebased and prepared a draft for `response.error` check.

dpkp:
awesome! will take a look shortly!

dpkp:
There are a handful of issues remaining -- need to reenable the test_switch_leader_async test in test/test_failover_integration.py and then fix it along w/ the lint issues failing travis build.

But you've put in a ton of work and I think it is basically there, so I am going to merge and push a few fixes directly. Thanks so much for working on this one!

vshlapakov:
Cool, thank you @dpkp! If I have some additional time, I'll fix other issues (like additional tests and docstrings) as well in a separate PR.

dpkp:
Thanks again for your help on this. I put together some edits in #388 (merged yesterday).

PR #334, opened 2015-03-03
Eagerly concat received bytes

pkoch:
This should reduce the amount of memory required for large packets. We're effectively side stepping keeping all the intermediary fragments around, which had us require 10% more the memory.

Couldn't test on neither pypy nor py34.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -79,7 +79,7 @@ class KafkaConnection(local):
 
     def _read_bytes(self, num_bytes):
         bytes_left = num_bytes
-        responses = []
+        result_buffer = six.BytesIO()
 
         log.debug("About to read %d bytes from Kafka", num_bytes)
 
@@ -104,9 +104,9 @@ class KafkaConnection(local):
 
             bytes_left -= len(data)
             log.debug("Read %d/%d bytes from Kafka", num_bytes - bytes_left, num_bytes)
-            responses.append(data)
+            result_buffer.write(data)
 
-        return b''.join(responses)
+        return result_buffer.getvalue()
 
     ##################
     #   Public API   #


pkoch:
Here's my science, for reference: https://gist.github.com/pkoch/05e854fb805c21b0b2c2

dpkp:
my quick testing shows similar improvement on py34, but substantially more memory use on pypy...

dpkp:
should _read_bytes() just return a BytesIO object and we refactor downstream consumers?  it would be nice to have some protocol decode/encode performance metrics.

kafka/conn.py
```diff
@@ -79,7 +79,7 @@ def _raise_connection_error(self):
 
     def _read_bytes(self, num_bytes):
         bytes_left = num_bytes
-        responses = []
+        result_buffer = six.BytesIO()
```
dpkp: is `six` needed here?  why not just use `io.BytesIO` which I believe is supported as of python 2.6 ?

pkoch:
Increased mem usage can be explained if pypy decided to delay the collection of the already appended chunks. And whatever other optimizations they're doing, it's hard to tell from the outside.

Did you exercise it enough for the JIT to kick in? I can't recall the actual number of passages for it to kick in (I think it was a big prime? Not sure).

kafka/conn.py
pkoch: Oh, right, we can totally do that. I thought we wanted to support elder pythons.

Now that I think of it, we might be able to use `memoryview` and such fast shenanigans since our floor version is 2.6. Let me get back to you on this.

pkoch: This is what I will be going for: http://stackoverflow.com/a/15964489 We'll have to use `buffer` instead of `memoryview`. I'll try to explore that avenue.

dpkp: that looks very promising!

dpkp:
the protocol stack has been re-written using BytesIO, so I'm closing this PR

pkoch:
Great! Thanks for the investment. Sorry I never got around to it!

Issue #344, opened 2015-03-11
kafkaconsumer is limited by 32int correlation_id


richchang0:
when the connection id hits 2,147,483,648 (one over int32 MAX), then the driver can't encode that message, and hits an error

dpkp:
thanks for the bug report and PR!  merged #355 and will be in 0.9.4

PR #345, opened 2015-03-11
rollover correlation id. addresses #344

richchang0:
when the correlation id hits 2,147,483,648 (one over int32 MAX), then the driver can't encode that message, and hits an error.

So in this changed I've done two things.

1. change the encoder to use an unsigned number from a signed for the correlation_id
2. removed itertools and replaced it with a normal python integer, with every _next() call converting the number to a ctype uint_32 to allow for an overflow to roll over back down to 0.

Not sure if this is the better design choice than just doing a check of itertools.count to see if it hit the intmax, then resetting it.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -2,11 +2,11 @@ import binascii
 import collections
 import copy
 import functools
-import itertools
 import logging
 import time
 import kafka.common
 
+from ctypes import c_uint32
 from kafka.common import (TopicAndPartition, BrokerMetadata,
                           ConnectionError, FailedPayloadsError,
                           KafkaTimeoutError, KafkaUnavailableError,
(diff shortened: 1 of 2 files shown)


dpkp:
looks great -- can you add a test ?

kafka/protocol.py
```diff
@@ -52,7 +52,7 @@ def _encode_message_header(cls, client_id, correlation_id, request_key):
         """
         Encode the common request envelope
         """
-        return struct.pack('>hhih%ds' % len(client_id),
+        return struct.pack('>hhIh%ds' % len(client_id),
```
dpkp: the kafka protocol defines this field as a signed int32.  i think it's better to stick with that.

kafka/client.py
```diff
@@ -101,7 +101,8 @@ def _next_id(self):
         """
         Generate a new correlation id
         """
-        return next(KafkaClient.ID_GEN)
```
dpkp: would it be simpler to just add a modulo here?  (`% 2**31`)

richchang0: ah yeah I think that's much simpler.

dpkp:
take a look at pr #355

dpkp:
merged #355 -- thanks for looking into this!

PR #351, opened 2015-03-22
Fixes an error when pulling a topic without partitions yet, behavior is to skip/ignore.

redsand:
Fixes an error when pulling a topic without partitions yet, behavior is to skip/ignore.


redsand:
Looks like the Travis CI build failed due to an unrelated issue.

dpkp:
`None` partition ids was a bug in the underlying client.  but it is already fixed in master and will be included in the next release.  see https://github.com/mumrah/kafka-python/commit/c80fbd14 .  Please reopen if you get the same error on master!

Issue #347, opened 2015-03-18
Auto commit with MultiProcessConsumer doesn't result in resume from last committed offset


alisaifee:
I've added a test case demonstrating what seemed like the expected behaviour to me [here](https://github.com/alisaifee/kafka-python/blob/master/test/test_consumer_integration.py#L330)

dpkp:
should be fixed now -- thanks for the test!

PR #352, opened 2015-03-23
Allow consumers to control start from beginning behavior

alisaifee:
Clients of the multi-process consumer currently have no way of enforcing that the consumer resumes from the last committed offset.  `SimpleConsumer` automatically receives this behaviour when `auto_commit` is enabled and can explicitly request the offset using `fetch_last_known_offsets`. 

To ensure backward compatibility a `start_from_beginning` parameter has been added to the base consumer which defaults to `True`. 

Reference: #347

### DIFF
--- a/kafka/consumer/base.py
+++ b/kafka/consumer/base.py
@@ -40,7 +40,8 @@ class Consumer(object):
     """
     def __init__(self, client, group, topic, partitions=None, auto_commit=True,
                  auto_commit_every_n=AUTO_COMMIT_MSG_COUNT,
-                 auto_commit_every_t=AUTO_COMMIT_INTERVAL):
+                 auto_commit_every_t=AUTO_COMMIT_INTERVAL,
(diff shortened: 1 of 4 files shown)


alisaifee:
The failing test is unrelated. Theres a new PR #353 which pretty much does the same thing (with inverted logic for the default value of the kwarg for `SimpleConsumer`).  I'm ok with closing this if #353 is preferred.

vshlapakov:
Oops, my apologies @alisaifee, I missed your PR (searched by word `MultiProcessConsumer` and found earlier hanging PR #252). They both do practically the same and I'll be fine too if your PR is preferred, it's be just great to add this functionality asap.

dpkp:
I think there's a better approach to this -- see #356 . I cherry-picked your test commits :)

alisaifee:
:+1: 
closing this PR.
Thanks!

PR #355, opened 2015-03-29
correlation_id modulo

dpkp:
Rollover KafkaClient correlation ids at 2**31 to keep within int32 protocol encoding
Fixes Issue #344


richchang0:
ah simpler and faster. very nice. sorry i wasn't able to get to this myself.

Issue #173, opened 2014-05-16
MultiProcessConsumer disregards topic offsets between subsequent runs


macmarcin:
Not sure if this works as designed but, every time I invoke MultiProcesConsumer on same topic/partition (new execution thread), the consumer disregards the consumer offset in Zookeeper and reads all the messages that are available. SimpleProcessConsumer only reads messages that have not been read. Shouldn't MultiProcessConsumer behave the same way?

Code snippet:
<pre><code>
brokers=["localhost:9092","localhost:9093","localhost:9094"]
kafka = KafkaClient(brokers)
consumer = MultiProcessConsumer(kafka, "marcin-group", "test_1", num_procs=2)
logging.debug('*** all offsets=%s' %consumer.offsets)
consumed_count=0
for message in consumer:
    print(message)
    consumed_count += 1

logging.debug('*** all offsets=%s ...after consuming all messages, number of messages consumed=%s' % (consumer.offsets, consumed_count)
----------------------------------------
ZK info:
get /kafka/consumers/marcin-group/offsets/test_1/0
200
cZxid = 0x1a2f0
--------
Debug snippets from the log file:
DEBUG:root:*** all offsets={0: 200}
.....
DEBUG:kafka:Commit offset 100 in SimpleConsumer: group=marcin-group, topic=test_1, partition=0
.....
DEBUG:kafka:Commit offset 200 in SimpleConsumer: group=marcin-group, topic=test_1, partition=0
.....
DEBUG:root:*** all offsets={0: 200} ...after consuming all messages, number of messages consumed=200
</code></pre>

dpkp:
i think the best solution here is to fetch committed offsets in the base consumer on __init__.  currently it is only done when autocommit is configured, but any consumer that uses a pre-existing group should use those offsets by default.

dpkp:
See PR #356 -- I recall now that the reason SimpleConsumer only fetches commits when auto_commit=True was because everyone was using server v0.8.0 and it did not have support for offset commit/fetch. We should change to check whether client.group is None.

vshlapakov:
Sounds good to me, it's simpler to reset offsets only when required (minority of cases), and resuming offsets should be the default behavior.

dpkp:
#356 merged -- this should be fixed now and will be available in 0.9.4 release

PR #356, opened 2015-03-30
fetch commit offsets in base consumer unless group is None

dpkp:
Fix for Issue #173 (and #347) w/ included test case from @alisaifee

### DIFF
--- a/kafka/consumer/base.py
+++ b/kafka/consumer/base.py
@@ -83,12 +83,13 @@ class Consumer(object):
         if not partitions:
             partitions = self.client.get_partition_ids_for_topic(self.topic)
 
-        for partition in partitions:
-            (resp,) = self.client.send_offset_fetch_request(
-                self.group,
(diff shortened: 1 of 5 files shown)


kafka/consumer/base.py
```diff
@@ -68,29 +68,42 @@ def __init__(self, client, group, topic, partitions=None, auto_commit=True,
...
+
     def fetch_last_known_offsets(self, partitions=None):
+        if self.group is None:
+            raise ValueError('KafkaClient.group must not be None')
+
         if not partitions:
             partitions = self.client.get_partition_ids_for_topic(self.topic)
 
-        def get_or_init_offset(resp):
+        for partition in partitions:
```
ecanzonieri: Would it be possible to send the offset requests for all partitions at once?

dpkp: good call -- adding a commit for that

Issue #280, opened 2014-12-25
bug in kafka/consumer/simple.py L255


tyronecai:
https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/simple.py#L255

there is a bug when get_partition_info is True, so should change to:

if get_partition_info:
    return None, None
else:
    return None

dpkp:
agree design is off, but at the interface should behave as documented...

dpkp:
looking a bit closer -- the current interface is documented like so:

```
        If no messages can be fetched, returns None.
        If get_partition_info is None, it defaults to self.partition_info
        If get_partition_info is True, returns (partition, message)
        If get_partition_info is False, returns message
```

which is how it works (If no messages can be fetched, returns None. -- regardless of the get_partition_info config).  that may be a bad design, but I dont think there's a bug here.

PR #358, opened 2015-03-31
Honor get_partition_info in consumer._get_message on timeout / queue empty

dpkp:
Fix issue #280


dpkp:
further review, the code works as documented:
```
 If no messages can be fetched, returns None.
```

PR #361, opened 2015-03-31
Make external API consistently support python3 strings for topic.

kecaps:
I noticed that python3 string topics were supported in SimpleProducer.send_messages but not elsewhere in the API. This change makes python3 string topics supported throughout the public API.

This required changes in some testing code to ensure that bytes were used when building internal structs for return and sending to kafka service.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -258,12 +258,14 @@ class KafkaClient(object):
         self.topic_partitions.clear()
 
     def has_metadata_for_topic(self, topic):
+        topic = kafka_bytestring(topic)
         return (
           topic in self.topic_partitions
           and len(self.topic_partitions[topic]) > 0
         )
 
     def get_partition_ids_for_topic(self, topic):
+        topic = kafka_bytestring(topic)
         if topic not in self.topic_partitions:
             return []
 
@@ -312,6 +314,7 @@ class KafkaClient(object):
         Partition-level errors will also not be raised here
         (a single partition w/o a leader, for example)
         """
+        topics = [kafka_bytestring(t) for t in topics]
         resp = self.send_metadata_request(topics)
 
         log.debug("Broker metadata: %s", resp.brokers)
--- a/kafka/consumer/base.py
+++ b/kafka/consumer/base.py
@@ -10,7 +10,7 @@ from kafka.common import (
     UnknownTopicOrPartitionError, check_error
 )
 
-from kafka.util import ReentrantTimer
+from kafka.util import kafka_bytestring, ReentrantTimer
 
 log = logging.getLogger("kafka")
 
@@ -44,8 +44,8 @@ class Consumer(object):
                  auto_commit_every_t=AUTO_COMMIT_INTERVAL):
 
         self.client = client
-        self.topic = topic
-        self.group = group
+        self.topic = kafka_bytestring(topic)
+        self.group = None if group is None else kafka_bytestring(group)
         self.client.load_metadata_for_topics(topic)
         self.offsets = {}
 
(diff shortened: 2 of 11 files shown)


kafka/client.py
```diff
@@ -258,15 +258,19 @@ def reset_all_metadata(self):
         self.topic_partitions.clear()
 
     def has_metadata_for_topic(self, topic):
```
dpkp: why not simply set `topic = kafka_bytestring(topic)` ?

kafka/client.py
```diff
@@ -258,15 +258,19 @@ def reset_all_metadata(self):
...
-          and len(self.topic_partitions[topic]) > 0
-        )
+        if not topic in self.topic_partitions:
+            bytes_topic = kafka_bytestring(topic)
+            if (bytes_topic is topic) or (not bytes_topic in self.topic_partitions):
+                return False
+            topic = bytes_topic
+        return len(self.topic_partitions[topic]) > 0
 
     def get_partition_ids_for_topic(self, topic):
```
dpkp: same - `topic = kafka_bytestring(topic)` seems like a simpler approach here

kafka/consumer/base.py
```diff
@@ -10,7 +10,7 @@
     UnknownTopicOrPartitionError, check_error
 )
 
-from kafka.util import ReentrantTimer
+from kafka.util import (kafka_bytestring, ReentrantTimer)
```
dpkp: parens not needed here

kafka/producer/base.py
```diff
@@ -170,6 +170,8 @@ def send_messages(self, topic, partition, *msg):
 
         All messages produced via this method will set the message 'key' to Null
         """
+        if not isinstance(topic, six.binary_type):
```
dpkp: why not use kakfa_bytestring here?

kafka/producer/keyed.py
```diff
@@ -57,10 +58,14 @@ def _next_partition(self, topic, key):
         return partitioner.partition(key)
 
     def send_messages(self,topic,key,*msg):
+        if not isinstance(topic, six.binary_type):
```
dpkp: kafka_bytestring ?

kafka/producer/keyed.py
```diff
@@ -57,10 +58,14 @@ def _next_partition(self, topic, key):
...
+            topic = topic.encode('utf-8')
         partition = self._next_partition(topic, key)
         return self._send_messages(topic, partition, *msg,key=key)
 
     def send(self, topic, key, msg):
+        if not isinstance(topic, six.binary_type):
```
dpkp: kafka_bytestring ?

kafka/util.py
```diff
@@ -95,6 +95,8 @@ def kafka_bytestring(s):
         return s
     if isinstance(s, six.string_types):
         return s.encode('utf-8')
+    if s is None:
```
dpkp: why do we want to support `None` here?

kafka/client.py
kecaps: I did it this way as a premature optimization because I saw has_metadata_for_topic and get_partition_ids_for_topic were on the callpath of send_messages which is probably the most used method and always calls these metadata methods with a byte string. But on closer inspection, it should only call them once on initialization, so its probably clearer to use a consistent pattern.

kafka/util.py
kecaps: group was None in some circumstances. Can catch that before calling kafka_bytestring(group).

dpkp:
thanks -- test error looks like a travis-ci failure.  do you mind squashing your second commit or git commit --amend ?

dpkp:
thanks!

Issue #301, opened 2015-01-22
Raising a FailedPayloadsError loses response information for other payloads


jobevers:
In `client._send_broker_aware_request` (https://github.com/mumrah/kafka-python/blob/master/kafka/client.py#L199), if failed_payloads exists, raising an Exception will cause the lose of the other response information (in `acc`)

Is there a reason for raising the Exception?  Would it be acceptable to return that information along with the responses instead?

dpkp:
agree -- the current error processing dates back to sept/oct 2013 and could use some love.  have time to submit a PR?

jobevers:
It would be a backwards compatible change - I can try to take a stab at a PR if you're okay with that.

dpkp:
please do - thanks!

dpkp:
see #366 -- FailedPayloads() objects are now returned in the list of responses rather than raised immediately.

PR #366, opened 2015-04-06
Refactor client request/response handling to support better retries

dpkp:
Return a list of responses and include individual (unraised) FailedPayloadsError instances rather
than always raising a FailedPayloadsError. This should allow producers to determine which payloads succeeded and which failed, specifically.


dpkp:
Fixes Issue #301

PR #369, opened 2015-04-15
Fix wrong documention

akshayn1107:
get_messages actually blocks for the full time if block is True regardless of whether or not count messages has been fetched.


dpkp:
Wont merge. The documentation accurately reflects the intended interface.

If get_messages isn't working as documented, then there's probably a bug. I can't replicate this issue locally; can you post some code or steps to replicate?

dpkp:
closing the PR -- please open an issue on this instead.

PR #376, opened 2015-05-14
Fix calling of _offsets

hellais:
Previously you would see this error:
```
self.offsets._fetch[(topic, partition)],
AttributeError: 'function' object has no attribute '_fetch'
```


dpkp:
Good catch, but I think it should be `self._offsets.fetch(....)` (drop the _ before fetch)

hellais:
ah yes, sorry. Fixed.

dpkp:
commit --amend and push -f to update the PR?

PR #384, opened 2015-05-29
Async producer does not silently throw away messages if Kafka is unavailable

imbusy:
The idea is that once we find out we can't deliver a message using the asynchronous method, the producer should throw an exception when trying to enqueue new messages. The async producer constantly checks if Kafka is available again and, when it is, sends the enqueued messages and signals the main thread that it's safe to add messages again.

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -82,20 +82,20 @@ def _send_upstream(queue, client, codec, batch_time, batch_size,
                 
                 # Set connection state as available
                 connection_state_lock.acquire()
-                connection_exc = None
+                connection_exc[0] = None
                 connection_state_lock.release()
             except Exception as e:
                 log.exception("Unable to send message")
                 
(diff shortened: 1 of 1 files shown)


dpkp:
Hi imbusy -- thanks for the PR!

The async producer is actually in the middle of a fairly substantial refactor here: https://github.com/mumrah/kafka-python/pull/331

So I won't merge this PR as is, but perhaps you can take a look at PR 331 and if you believe this issue remains, can you rebase against it?

Thanks!

imbusy:
Thanks @dpkp . I'll have a look at #331

dpkp:
#331 is merged and more logging added in #388 .  The behavior you describe can be achieved by setting a limit on the internal queue size and a timeout on putting messages to the queue when it is full.  See `async_queue_maxsize` and `async_queue_put_timeout` configuration parameters to the producer.

Issue #386, opened 2015-06-04
the client seems cannot handle server failure properly


reAsOn2010:
Environment:

* single instance Zookeeper (version: 3.4.6)
* a Kafka cluster with four broker on one machine, running in supervisor  (version: 0.8.2.0)
* Java version "1.8.0_45"
* topic info:

```
Topic:timeline	PartitionCount:20	ReplicationFactor:2	Configs:
	Topic: timeline	Partition: 0	Leader: 0	Replicas: 0,1	Isr: 0,1
	Topic: timeline	Partition: 1	Leader: 1	Replicas: 1,2	Isr: 2,1
	Topic: timeline	Partition: 2	Leader: 2	Replicas: 2,3	Isr: 2,3
	Topic: timeline	Partition: 3	Leader: 3	Replicas: 3,0	Isr: 3,0
	Topic: timeline	Partition: 4	Leader: 0	Replicas: 0,1	Isr: 0,1
	<and more...>
```

I tried the latest code on branch master, to test cluster failover. It seems that SimpleConsumer exited due to an AttributeError when a broker was shutdown. Below is The stack:

	[E 150604 11:24:47 /Users/leon/zhihu/kafka-python/kafka/conn.py:107 leon-MacBook-Pro:22207] Unable to receive data from Kafka
	Traceback (most recent call last):
	  File "/Users/leon/zhihu/kafka-python/kafka/conn.py", line 104, in _read_bytes
	    raise socket.error("Not enough data to read message -- did server kill socket?")
	error: Not enough data to read message -- did server kill socket?
	[W 150604 11:24:47 /Users/leon/zhihu/kafka-python/kafka/client.py:204 leon-MacBook-Pro:22207] Could not receive response to request [00000084000100000000004a000c6b61666b612d707974686f6effffffff0000ea600000000100000001000874696d656c696e650000000500000012000000000000000900001000000000020000000000000028000010000000000e0000000000000009000010000000000a00000000000000070000100000000006000000000000000900001000] from server <KafkaConnection host=kids.aws.dev port=9094>: Kafka @ kids.aws.dev:9094 went away
	Traceback (most recent call last):
	  File "zhihu_logger/consumer.py", line 23, in <module>
	    main()
	  File "zhihu_logger/consumer.py", line 19, in main
	    for m in consumer:
	  File "/Users/leon/zhihu/kafka-python/kafka/consumer/simple.py", line 311, in __iter__
	    message = self.get_message(True, timeout)
[...]

PR #392, opened 2015-06-08
try to fix uncaught FailedPayloadsError in consumer

reAsOn2010:
about #386. It‘s a not that elegant fix, but works...

### DIFF
--- a/kafka/consumer/simple.py
+++ b/kafka/consumer/simple.py
@@ -19,7 +19,7 @@ from kafka.common import (
     FetchRequest, OffsetRequest,
     ConsumerFetchSizeTooSmall, ConsumerNoMoreData,
     UnknownTopicOrPartitionError, NotLeaderForPartitionError,
-    OffsetOutOfRangeError, check_error
+    OffsetOutOfRangeError, FailedPayloadsError, check_error
 )
 from .base import (
     Consumer,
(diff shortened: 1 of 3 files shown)


reAsOn2010:
And this fix not solve the second problem about NotLeaderForPartitionError after the failed broker recover

kafka/consumer/simple.py
```diff
@@ -341,6 +341,8 @@ def _fetch(self):
             for resp in responses:
 
                 try:
+                    if isinstance(resp, Exception):
+                        raise resp
```
dpkp: this check can probably move into the kafka.utils.check_error function

kafka/consumer/simple.py
```diff
@@ -353,6 +355,13 @@ def _fetch(self):
                     # Retry this partition
                     retry_partitions[resp.partition] = partitions[resp.partition]
                     continue
+                except FailedPayloadsError:
```
dpkp: would do `except FailedPayloadsError as e` here

kafka/consumer/simple.py
```diff
@@ -353,6 +355,13 @@ def _fetch(self):
...
+                    log.warning("Failed payloads of %s"
+                                "Resetting partition offset...",
+                                resp.failed_payloads)
+                    # Retry this partition
+                    retry_partitions[resp.failed_payloads.partition] = partitions[resp.failed_payloads.partition]
```
dpkp: .failed_payloads has been changed to .payload .  suggest using `e.payload.partition`

dpkp:
thanks for the PR! I added a few comments, and also would love if you could take a stab at adding a test for this.

reAsOn2010:
... why I mentioned others' commits, or I make a new PR?

Issue #408, opened 2015-06-16
Null payload for Kafka log compaction deletion feature


vshlapakov:
According to [the official doc](https://cwiki.apache.org/confluence/display/KAFKA/Log+Compaction) Kafka log compaction could be used to delete messages from the log:
```Compaction also allows from deletes. A message with a key and a null payload will be treated as a delete from the log. This delete marker will cause any prior message with that key to be removed (as would any new message with that key), but delete markers are special in they will themselves be cleaned out of the log after a period of time.```

But now there's a [limitation](https://github.com/mumrah/kafka-python/blob/master/kafka/producer/base.py#L358) in the code that doesn't allow to pass null payloads, so the deletion feature doesn't work at all. Was it done by purpose for some reasons or I can prepare a short PR to fix it?

vshlapakov:
Tested; Kafka deletion feature works with this patch, take a look please!

dpkp:
PR welcome. Can you also take a look at https://github.com/mumrah/kafka-python/issues/124 while you're addressing None values?

PR #409, opened 2015-06-16
Allow null payload for deletion feature

vshlapakov:
PR for #408.

[Review on Reviewable](https://reviewable.io/reviews/mumrah/kafka-python/409)

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -354,9 +354,15 @@ class Producer(object):
         if not isinstance(msg, (list, tuple)):
             raise TypeError("msg is not a list or tuple!")
 
-        # Raise TypeError if any message is not encoded as bytes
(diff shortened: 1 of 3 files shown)


dpkp:
Should we only allow None values if a key is provided? Also I think it is worth adding a comment w/ a link to relevant apache kafka protocol docs.

Shouldn't we add a test or two as well?

vshlapakov:
Yeah, good catch: I think that message with no key and no payload doesn't make sense at all. I'll add a comment with a link and some tests.

vshlapakov:
@dpkp Check it please, do we need to add something else here?

vshlapakov:
The PR is waiting for code review, check it please!

vshlapakov:
Thanks! :dancers:

PR #424, opened 2015-07-07
Add metrics to KafkaClient.

ucarion:
I am trying to collect data about how long I'm waiting on Kafka; I want to answer questions like "how much latency is there for getting messages out of Kafka?". To this end, I've put together this PR, which adds an option to enable metrics collection in `KafkaClient`.

This uses [scales](https://github.com/Cue/scales), and would let me write stuff like this:

```python

consumer = KafkaConsumer('foobar',
                         bootstrap_servers=['localhost:9092'],
                         group_id='kafka_test',
                         enable_metrics=True)

# interesting logic would go here ...
for _ in xrange(10):
    for message in consumer.fetch_messages():
        process(message)

# outputs the 95th percentile of how long it takes to perform send_fetch_request
print consumer.metrics.fetch_request_timer['95percentile']
```

I've only added support for metrics in KafkaConsumer, but I can add other consumers if you agree with this design.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -26,9 +26,8 @@ def time_metric(metric_name):
             start_time = time.time()
             ret = fn(self, *args, **kwargs)
 
-            if self.metrics:
-                metric = getattr(self.metrics, metric_name)
-                metric.addValue(time.time() - start_time)
+            if self.metrics_responder:
+                self.metrics_responder(metric_name, time.time() - start_time)
 
             return ret
         return wrapper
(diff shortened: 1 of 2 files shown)


kafka/client.py
```diff
@@ -48,6 +50,20 @@ def __init__(self, hosts, client_id=CLIENT_ID,
     #   Private API  #
     ##################
 
+    def time_metric(metric_name):
+        def decorator(fn):
+            def wrapper(self, *args, **kwargs):
```
patricklucas: Check out [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps).

ecanzonieri:
It may also be useful to get errors count as part of these metrics

kafka/consumer/kafka.py
```diff
@@ -38,6 +38,8 @@
     'auto_commit_interval_messages': None,
     'consumer_timeout_ms': -1,
 
+    'metrics_responder': None,
```
dpkp: the upstream client configuration here is `metric.reporters`. I try to stay consistent, so perhaps `metric_reporters` and support a list of reporter interfaces?

kafka/consumer/kafka.py
```diff
@@ -135,10 +137,13 @@ def configure(self, **configs):
                 'bootstrap_servers required to configure KafkaConsumer'
             )
 
+        self.metrics_responder = self._config['metrics_responder']
```
dpkp: similarly would like to name this `metric_reporters` and make it a list

dpkp:
Metrics support would be great -- if you have time, take a look at my comments re: consistency w/ upstream client approach. Otherwise I'll probably put this off for a bit and clean up w/ a larger 0.9 client refactor.

PR #432, opened 2015-07-21
Use debug logging level for metadata request

ecanzonieri:


zackdever:
Yes please. We're having to use a log filter for `Updating topic metadata`.

Issue #434, opened 2015-07-24
Producer cleanup logic is not handling 'stopped' flag correctly


toli:
The logic in the `producer.cleanup` section is backwards - we are only calling `stop()` on an object that's already stopped:
```
def cleanup(obj):
     if obj.stopped:
        obj.stop()
```
This should be `if not obj.stopped` instead.

This came up in our case where we were using `kafka-python` producers under `gunicorn`, and when I was load-testing and sending over 10k messages, `gunicorn` would stop/recreate the process and some messages were lost since `gunicorn` try to shutdown the process and cleanup logic wouldn't get executed.

The `if` statement change fixes it.

PR #435, opened 2015-07-24
Reworked the if statement logic to only call stop() on not-stopped producer objects

toli:
Reworked the if statement logic to only call stop() on not-stopped objects.
Fixing #434 
added tests

Ironically, this was one of the only pieces of code not "covered" by unit tests.

### DIFF
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -322,7 +322,7 @@ class Producer(object):
             self.thread.start()
 
             def cleanup(obj):
-                if obj.stopped:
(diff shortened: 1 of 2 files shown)


rogaha:
:+1: great job catching that @toli!

tleach:
I was reading through the `atexit` cleanup logic, as it's somewhat important to me that this runs and we don't drop messages on restart. I thought this logic seemed wrong. Great job on fixing it!

@mumrah can this be merged soon?

PR #441, opened 2015-07-31
Include LICENSE in PyPI sdist

koobs:
Allows downstream OS packagers and porters to include the LICENSE file in package installations


dpkp:
Thanks!

PR #448, opened 2015-08-27
match SimpleProducer sendmessages with key docs and code

c19:
http://kafka-python.readthedocs.org/en/latest/usage.html#keyed-messages

producer.send_messages(b'my-topic', b'key1', b'some message')
originally will two None key messages (None, 'key1') (None, 'some message')
fixed send with key would be:
producer.send_messages(b'my-topic', b'some message', key=b'key1')

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -66,8 +66,8 @@ Keyed messages
 
     # HashedPartitioner is default (currently uses python hash())
     producer = KeyedProducer(kafka)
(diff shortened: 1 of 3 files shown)


dpkp:
I think maybe the documentation is just confusing? the docstring for SimpleProducer.send_messages does say: `All messages produced via this method will set the message 'key' to Null`

For keyed messages you might try the `KeyedProducer` class.

c19:
Yes, I figured it out..

PR #456, opened 2015-09-18
Support retry semantics in MultiProcessConsumer

duggan:
Have found these changes useful in situations where the client is being run in a daemon, where temporary connectivity loss is more likely over long running periods, and for recovering from broker restarts / unavailability.

Changes best viewed with the `?w=1` flag in the URL, as most of it is indentation.


dpkp:
Thanks!

PR #457, opened 2015-09-18
Consumers get_messages: allow blocking until some messages are received

saaros:
Modified MultiProcessConsumer's and SimpleConsumer's `block` argument to
allow integer value which defines the number of messages to block for.

This allows callers to ask for a high number of messages and block only
until some of them are received.  Otherwise callers would have to request
messages one by one or block for some time.


dpkp:
Thanks for the PR!

PR #463, opened 2015-10-14
Fixes silly counting bug in exception message

bschopman:


dpkp:
Can you explain? request_tries holds messages that have been popped off the queue but that have not been successfully delivered yet. Don't we need to include both those unsent messages and the messages still on the queue in the unsent count?

bschopman:
My bad, the variable name confused me.

PR #464, opened 2015-10-14
Prevents crashing communication thread of async producer

bschopman:
If an uncaught exception occurs in _send_messages() the thread sending
data to Kafka (asynchronously) will crash and the queue will never be
emptied. To reproduce:
1) Run an Async producer.
2) Kill the Kafka server.
3) Restart the Kafka server.
The communication thread dies shortly after step 2. After step 3 the
communication does not resume without this commit.

The change in `KafkaClient::_send_broker_aware_request()` prevents an
exception from being thrown when it's called by
`KafkaClient::send_produce_request()` with `fail_on_error=False`.
Other exceptions besides `KafkaUnavailableError` that can be thrown include:
`LeaderNotAvailableError`, `UnknownTopicOrPartitionError` and 
`UnknownTopicOrPartitionError`.

The try/except in `KafkaClient::_send_broker_unaware_request()` prevents
an exception to crash the communication thread.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -7,7 +7,7 @@ import time
 import kafka.common
 from kafka.common import (TopicAndPartition, BrokerMetadata,
                           ConnectionError, FailedPayloadsError,
-                          KafkaTimeoutError, KafkaUnavailableError,
+                          KafkaTimeoutError, KafkaUnavailableError, KafkaError,
                           LeaderNotAvailableError, UnknownTopicOrPartitionError,
                           NotLeaderForPartitionError, ReplicaNotAvailableError)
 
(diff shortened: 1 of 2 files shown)


bschopman:
All the checks failed, but the error doesn't ring a bell. What can I do to fix this?
    The command "if [ -n "$UNIT_AND_LINT_ONLY" ]; then tox -e lint,`./travis_selector.sh $TRAVIS_PYTHON_VERSION`; else tox -e `./travis_selector.sh $TRAVIS_PYTHON_VERSION`; fi" exited with 1.

bschopman:
This morning I've installed all dependencies. `tox -e lint` passes, but `tox -e py27` has been running for 6 hours now. I see that it communicates with a Kafka server that I'm running, but it hasn't printed more than:

```
GLOB sdist-make: /home/bschopman/git/kafka-python/setup.py
py27 inst-nodeps: /home/bschopman/git/kafka-python/.tox/dist/kafka-python-0.9.5-dev.zip
py27 runtests: commands[0] | nosetests -v -x --with-id --id-file=/home/bschopman/git/kafka-python/.tox/py27/.noseids --with-timer --timer-top-n 10 --with-coverage --cover-erase --cover-package kafka
```

PR #467, opened 2015-10-22
Prevents crashing communication thread of async producer

bschopman:
If an uncaught exception occurs in _send_messages() the thread sending
data to Kafka (asynchronously) will crash and the queue will never be
emptied. To reproduce:
1) Run an Async producer.
2) Kill the Kafka server.
3) Restart the Kafka server.
The communication thread dies shortly after step 2. After step 3 the
communication does not resume without this commit.

The changes in both files prevent an Exception from being thrown through
to do main communication process, which could cause the crash.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -478,7 +478,11 @@ class KafkaClient(object):
         else:
             decoder = KafkaProtocol.decode_produce_response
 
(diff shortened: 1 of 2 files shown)


dpkp:
Nice -- would love to get a test coverage for this issue

PR #473, opened 2015-11-11
Use unblocking io for broker aware requests

ecanzonieri:
Send broker aware requests in parallel and use select to read from available sockets. 
This is a good improvement when using large clusters and reading from multiple partitions/topics which span on many brokers. 
Do not expect performance improvement when reading from a single partition.


dpkp:
Nice!

PR #478, opened 2015-11-30
SSL-support added

geogeurt:
tested with Python 2.6 and 2.7


dpkp:
Have you taken a look at PR #474 ? Is there a reason to prefer this approach instead? If so, can you take a look at test failures?

Also, as mentioned in #474, I would like to defer SSL support for a larger 0.9 client release, so going to defer merges on this for a little bit.

dpkp:
closing in favor of #621

PR #505, opened 2016-01-11
Add event registrar and events for async producer

thieman:
@dpkp We ended up building this in our fork and, though it's not exactly ready to be merged, I thought it'd be valuable to throw it out here so we could talk about the use case.

Our issue was that the driver, and the async producer in particular, does a lot of fairly complex stuff which we would really like to be able to keep track of in our monitoring system. Basic metrics like throughput, connection attempts, failures and retries, etc. would greatly help us understand what's going on and alert if we have a problem in our cluster. To do this, we need a way to get information out of the driver.

The approach here is to add a barebones event handling system and emit events at all the critical points in the async producer's operation. If no handlers are registered, this should incur basically zero overhead, but it allows the enduser the option of tying their own events (in our use case, metrics emission) into the producer's workflow.

### DIFF
--- a/kafka/util.py
+++ b/kafka/util.py
@@ -151,3 +151,18 @@ class ReentrantTimer(object):
 
     def __del__(self):
         self.stop()
+
+class EventRegistrar(object):
+    """
+    Handles registration of callable event handlers which are
+    executed in sequence when events are emitted.
+    """
+    def __init__(self):
+        self.handlers = collections.defaultdict(list)
+
+    def register(self, event, handler):
+        self.handlers[event].append(handler)
+
+    def emit(self, event, *args, **kwargs):
+        for handler in self.handlers[event]:
+            handler(*args, **kwargs)
--- a/kafka/producer/base.py
+++ b/kafka/producer/base.py
@@ -20,6 +20,7 @@ from kafka.common import (
     RequestTimedOutError, AsyncProducerQueueFull, UnknownError,
     RETRY_ERROR_TYPES, RETRY_BACKOFF_ERROR_TYPES, RETRY_REFRESH_ERROR_TYPES
 )
+from kafka.util import EventRegistrar
 
 from kafka.protocol import CODEC_NONE, ALL_CODECS, create_message_set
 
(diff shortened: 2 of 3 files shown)


dpkp:
Awesome - will take a look. Thanks for the PR!

kafka/util.py
```diff
@@ -151,3 +151,18 @@ def stop(self):
...
+    """
+    def __init__(self):
+        self.handlers = collections.defaultdict(list)
+
+    def register(self, event, handler):
+        self.handlers[event].append(handler)
+
+    def emit(self, event, *args, **kwargs):
+        for handler in self.handlers[event]:
+            handler(*args, **kwargs)
```
dpkp: we should wrap in try / except to prevent an errant handler from crashing the producer thread

dpkp:
This is fairly similar to the way the new java client handles metrics, and I think kafka-python should attempt to emulate that behavior.

The emit interface will work great for folks that have an external aggregation system like statsite / statsd; I suppose we could also use it to pipe into an internal aggregator if that's needed.

The other general comment is that I'd like to favor composable interfaces, using configuration parameters that are similar -- if not identical -- to the official java client. In this case we can probably address that later, but at the expense of likely breaking inheritance interfaces you may be relying on right now.

Finally, I should comment that I expect to transition the async producer to use the new async client and emulate the official java producer interface more closely (return futures, accumulate batches more flexibly). If/when that happens, the internal interfaces are likely to change and may break your internal fork!

thieman:
> Finally, I should comment that I expect to transition the async producer to use the new async client and emulate the official java producer interface more closely (return futures, accumulate batches more flexibly).

I'm actually quite interested in this, as we've been digging more and more into the driver and our current thinking is that we need to make quite a few more improvements to make it production-ready for us. We might be able to help out with your efforts if you point us in the right direction.

dpkp:
Awesome! The general plan is to expose the same interface as the KafkaProducer from the latest java release. In order to do that I had planned to reimplement the Accumulator and Sender classes from the java source tree. Prior to that it may be useful to simply re-package the current SimpleProducer w/ a KafkaProducer api, without support for most configuration options and likely not returning Futures just yet.

jeffwidman:
@thieman it looks like your original intent was to support metrics, but that has since been added to kafka-python in #637... so should we close this or do you need additional functionality that this provides?

thieman:
@jeffwidman I'm no longer able to actively participate in this project, please carry on as you see fit

Issue #506, opened 2016-01-11
New client imports are extremely confusing


thieman:
I've been playing around with current master and I recently just got bitten hard by this recent change to the module init:

```
from kafka.client import KafkaClient as SimpleClient
from kafka.client_async import KafkaClient
```

I realize that this will probably be a little less difficult to get hurt by when the full release is out and the changelog is updated (and hopefully major version bumped), but this is very counterintuitive. I have `from kafka import KafkaClient` in my application code, and in any normal situation I'd be able to grep for `class KafkaClient` in the library and find the class I want.

I'm wondering if this change is necessary and if it's worth the obfuscation. If there is going to be a backwards-incompatible change either way, would it make more sense to import them as `SimpleClient` and `AsyncClient`?

dpkp:
Agree that it is a bit confusing currently, but the intent is to deprecate the old "KafkaClient" interface completely. I'm sorry this caused you pain! We should probably just rename the class to SimpleClient internally.

I considered `AsyncClient`, but I dont like labelling it as asynchronous specifically. In fact, I'd like to add a synchronous send() to make it easier to use in the repl. The bits that were the most difficult to maintain, and that are not coming back, are the send_broker_aware / send_broker_unaware, and the different methods for each api call with the funky 'payloads' interface.

Were you calling api methods directly, or just importing the client to instantiate a SimpleConsumer / SimpleProducer ?

thieman:
I was importing the `kafka.client.KafkaClient` and instantiating it, which failed once I upgraded because the constructors of the sync and async clients have different signatures.

dpkp:
Ok -- and you were calling client methods directly via your application code? Which methods?

thieman:
I haven't tried to update my application to use the async client yet, I just updated my import to point to the old client so I could go merrily on my way. :) I'd need to dig into the OO structure a bit more to figure out what we're doing between grabbing the client and calling `send_messages` on the producer.

dpkp:
Right on. I ask because most common use case is just passing client to
SimpleProducer or SimpleConsumer.

A more complicated use case would be calling client methods directly. If
you are doing that, please let me know.

For new KafkaConsumer and KafkaProducer interfaces, you do not need to
import a client at all. The consumer/producer manages client instances
directly.

dpkp:
[...]

PR #507, opened 2016-01-13
Add DeprecationWarnings to legacy classes

dpkp:
Address concerns in #506


thieman:
:+1: this is the right approach to the problem

PR #513, opened 2016-01-23
Update usage.rst

giaosudau:
kafka.common.KafkaConfigurationError: Unknown configuration key(s): ['enable_auto_commit']
raise OffsetOutOfRangeError('Cannot reset partition offsets without a '
                                            'valid auto_offset_reset setting '
                                            '(largest|smallest)')

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -20,8 +20,8 @@ KafkaConsumer
                                               message.offset, message.key,
                                               message.value))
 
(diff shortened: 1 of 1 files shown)


dpkp:
Thanks for the PR, but the interface has changed in master and the docs reflect those changes - if you are using the pypi release, consult readthedocs for the version you have installed.

PR #531, opened 2016-02-03
Fix Exception type assertions for python3

sondinhinfectious:
Building off master and then setting up a KafkaConsumer raises AssertionErrors, as f.exceptions.args[0] is different in python3. ConnectionError also wasn't set up for python3.


dpkp:
This is not correct (see failing tests). Can you open an issue with details on what you think needs changing and why?

PR #537, opened 2016-02-08
Update index.rst

kyrsideris:
TopicPartition leaves inside kafka.common instead of kafka


dpkp:
Thanks! Actually, I had intended to make TopicPartition importable from the top-level module but apparently did not update the code. So in this case I'm going to update the code to be consistent with the docs, rather than vice-versa.

PR #540, opened 2016-02-09
producer re-request metadata when waiting on topic partitions

### LINKED ISSUE #539: sporadic timeouts producing to multiple auto-created topics
seeing this:
```
ERROR : kafka.cluster : Topic 5f2dc140e6e84fff8dc1acca1e186509 is not available during auto-create initialization
DEBUG : kafka.cluster : Updated cluster metadata version 3 to Cluster(brokers: 1, topics: 1, groups: 0)
```

looks like `KafkaProducer._wait_on_metadata` doesn't retry metadata request https://github.com/dpkp/kafka-python/blob/master/kafka/producer/kafka.py#L459-L468.

aisch:
closes #539

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -466,6 +466,8 @@ class KafkaProducer(object):
             partitions = self._metadata.partitions_for_topic(topic)
             if partitions:
                 return partitions
+            if future.is_done:
+                future = request_update(self, event)
         else:
             raise Errors.KafkaTimeoutError(
                 "Failed to update metadata after %s secs.", max_wait)


dpkp:
I believe the problem in #539 is a bit deeper than this. The problem is that the metadata response handler only fails on errors if a single topic was requested. If the metadata request included more than 1 topic, errors are essentially ignored. My initial thought is that this underlying problem should be fixed first.

dpkp:
I went back and did a bit more refactoring to _wait_on_metadata as well as the threading locks around metadata updates. I believe this is fixed now in master.

Issue #572, opened 2016-03-01
Log warnings about high cpu usage when number of consumers is higher than partition count


stefanth:
I have two consumers for a topic that only has one partition. When the idle consumer is running it prints out a constant stream of:
"_poll called with a non-zero timeout and sleep=False -- but there was nothing to do. This can cause high CPU usage during idle."

This is similar to issue #494 that had been fixed but it seems that commit b8c2097 broke it again. There is no way to pass the sleep=True parameter to the client object when calling the poll method of the KafkaConsumer object (in group.py). I tried changing line 465 of group.py to:
self._client.poll(timeout_ms, sleep=True)

and then everything worked correctly again. I don't see the reason for commit b8c2097 so I don't know if this would break anything else.

akheron:
I have the same issue.

dpkp:
Can you post code to reproduce + broker version? My quick tests locally do not reproduce this issue.

akheron:
Here's a simple program that reproduces the issue when you run two instances in parallel: https://gist.github.com/akheron/890d0506b803abb3130b

Kafka version 2.11-0.9.0.0.

dpkp:
ok - I was testing the iterator, not `consumer.poll`. I can reproduce this. Will fix.

akheron:
Works like a charm, thanks!

kashifusmani:
I am on version 1.0.0, what version is the fix released in?

kashifusmani:
nvm, found it, 1.0.2

kashifusmani:
I still see the message with 1.0.2. Is the message still supposed to get printed?

Issue #494, opened 2016-01-05
High CPU load when using Group Coordination


CVTJNII:
Using kafka.KafkaConsumer in multiple processes with the same group ID I'm noticing high CPU load when idle.  I'm running the kafka-python 0.9 branch (commit 31c3d59ee3507a65533a3db58dd0fa59d9925b11) against Kafka 0.9.0.0.  Group coordination is working as expected, with each message being processed by one consumer in the group, however (n-1)/n consumers in the group are running at 100% CPU.  With n=1 CPU load is low as expected, however with n>1 I see n-1 processes running 100% CPU as if they were busy-spinning.

I'm seeing this with the following, stripped-down consumer worker function:

```python
def consumer_worker(kafka_uri, topic, group_id, heartbeat, stats_array):
    consumer = kafka.KafkaConsumer(topic, group_id=group_id, bootstrap_servers=[kafka_uri])
    for message in consumer:
        print "consumer_worker iterating"
```

The print shows that the for isn't busy-spinning.  I have not performed any further profiling at this point.

CVTJNII:
More details:

This is seen when the consumer count exceeds the partition count.  I was testing about with a single partition test topic, which is where the 1 in (n-1) appears to be coming from.  If I have more partitions than consumers things behave as expected.

I haven't proven that the worker with no subscriptions is the one spinning the CPU yet, but after enabling logging I do notice that workers without partitions throw this error:

Task <kafka.coordinator.consumer.AutoCommitTask object at 0x7f787c0a8390> failed: future failed without an exception

dpkp:
Thabks

dpkp:
Thanks Tom - I'll investigate. Seems likely that there is a bug in handling idle consumers.

CVTJNII:
Here is the cProfiler output of the 'for message in consumer' block:
[profiler.txt](https://github.com/dpkp/kafka-python/files/79331/profiler.txt)

CVTJNII:
[...]

PR #577, opened 2016-03-09
Bug fix: KafkaConsumer.position()

stefanth:
Method KafkaConsumer.position() was not refreshing position correctly
when calling self._update_fetch_positions().

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -477,7 +477,7 @@ class KafkaConsumer(six.Iterator):
         assert self._subscription.is_assigned(partition), 'Partition is not assigned'
         offset = self._subscription.assignment[partition].position
         if offset is None:
-            self._update_fetch_positions(partition)
+            self._update_fetch_positions([partition])
             offset = self._subscription.assignment[partition].position
         return offset
 


kafka/consumer/group.py
```diff
@@ -462,7 +462,7 @@ def _poll_once(self, timeout_ms):
             return records
 
         self._fetcher.init_fetches()
-        self._client.poll(timeout_ms)
+        self._client.poll(timeout_ms, sleep=True)
```
dpkp: This is a substantial change - why? The current consumer design expects poll() to return immediately if there is nothing to block on.

stefanth: When a consumer is running that doesn't currently have anything to do it will print out a constant stream of when poll is called in a loop (which I assume is the normal method of consuming endlessly):
"_poll called with a non-zero timeout and sleep=False -- but there was nothing to do. This can cause high CPU usage during idle."

This is similar to issue #494 that had been fixed but it seems that commit b8c2097 broke it again. I feel that it is unneeded work for every client to check on the partition assignments before every call to poll to make sure that it is non-empty. For the client, there is little difference between receiving no data because there are no partitions currently assigned and receiving no data because the topic has no new messages. So allowing poll to sleep if a timeout is specified helps simplying the usage of the library.

dpkp: Thanks for the thoughts. Can you remove from this PR? This is a separate issue (and also I believe the incorrect fix for the bug you described).

stefanth:
You are right, it is a different issue that accidentally got added to the pull request because it was tracking my master. I have removed it. The poll() sleep issue is in issue #572.

PR #578, opened 2016-03-09
Bug fix: KafkaConsumer.position()

stefanth:
After manually assigning a TopicPartition and seeking it to the end, I didn't find
any way to obtain the offset. The reason was this bug in the position method. Now
the position method correctly refreshed the position after seek_to_end() is called.


dpkp:
thanks!

PR #589, opened 2016-03-14
Refactor NodeNotReadyError handling

dpkp:
This alters the behavior of some methods that previously raised NodeNotReadyError and now return a failed Future(). Also remove ready() check from coordinator_unknown() and let NodeNotReadyError propagate to heartbeat errback, which will remove the 'coordinator_unknown' log warning (but not ready exception will remain).


zackdever:
Super weird. The failures are from socket.socket not being mocked. On OS X the mock works and I can't reproduce the failure, but I can reproduce the Travis failures in an Ubuntu VM. Will dig some more tomorrow.

PR #615, opened 2016-03-30
Kafka IPv6 Support.

### LINKED ISSUE #616: IPv6 Kafka connections fails
Kafka broker IPv6 addresses are not parsed correctly to support "[x:y::z]:port" syntax.    Even when using hostname/fqdn, connections still use IPv4 sockets.

TimEvens:
IPv6 address without port can be defined as the IPv6 address.  If the address
is a hostname or if a port is included, then the address MUST be wrapped
in brackets [] (E.g. [somehost]:1234 or [fd00:1001::2]:1234).

Fixes  #616
Tests included.


dpkp:
very nice -- thanks!

PR #637, opened 2016-04-08
Metrics java port

zackdever:
Ports the java metrics functionality, but only has one example instrumentation to keep size down.

The only thing that isn't a straight port is the DictReporter, which I made in place of the JMX reporter since there is no obvious equivalent in python. I don't have a good feel yet for how that will be used, so I would consider it a rough draft towards the standard python reporter.

One thing to keep an eye on is performance hits by the use of locks.

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -12,6 +12,7 @@ from kafka.consumer.subscription_state import SubscriptionState
 from kafka.coordinator.consumer import ConsumerCoordinator
 from kafka.coordinator.assignors.range import RangePartitionAssignor
 from kafka.coordinator.assignors.roundrobin import RoundRobinPartitionAssignor
+from kafka.metrics import DictReporter, MetricConfig, Metrics
 from kafka.protocol.offset import OffsetResetStrategy
 from kafka.structs import TopicPartition
 from kafka.version import __version__
@@ -143,6 +144,13 @@ class KafkaConsumer(six.Iterator):
             offset commits; 0.8.0 is what is left. If set to 'auto', will
             attempt to infer the broker version by probing various APIs.
             Default: auto
+        metric_reporters (list): A list of classes to use as metrics reporters.
+            Implementing the AbstractMetricsReporter interface allows plugging
+            in classes that will be notified of new metric creation. Default: []
+        metrics_num_samples (int): The number of samples maintained to compute
+            metrics. Default: 2
+        metrics_sample_window_ms (int): The number of samples maintained to
+            compute metrics. Default: 30000
 
     Note:
         Configuration parameters are described in more detail at
@@ -181,9 +189,9 @@ class KafkaConsumer(six.Iterator):
         'ssl_keyfile': None,
         'api_version': 'auto',
         'connections_max_idle_ms': 9 * 60 * 1000, # not implemented yet
-        #'metric_reporters': None,
-        #'metrics_num_samples': 2,
-        #'metrics_sample_window_ms': 30000,
+        'metric_reporters': [],
+        'metrics_num_samples': 2,
+        'metrics_sample_window_ms': 30000,
     }
 
     def __init__(self, *topics, **configs):
@@ -202,6 +210,16 @@ class KafkaConsumer(six.Iterator):
                         new_config, self.config['auto_offset_reset'])
             self.config['auto_offset_reset'] = new_config
 
+        metrics_tags = {'client-id': self.config['client_id']}
+        metric_config = MetricConfig(samples=self.config['metrics_num_samples'],
+                                     time_window_ms=self.config['metrics_sample_window_ms'],
+                                     tags=metrics_tags)
+        reporters = [reporter() for reporter in self.config['metric_reporters']]
+        reporters.append(DictReporter('kafka.consumer'))
+        self._metrics = Metrics(metric_config, reporters)
+        metric_group_prefix = 'consumer'
+        # TODO _metrics likely needs to be passed to KafkaClient, Fetcher, etc.
+
         self._client = KafkaClient(**self.config)
 
         # Check Broker Version if not set explicitly
@@ -217,14 +235,13 @@ class KafkaConsumer(six.Iterator):
         self._fetcher = Fetcher(
             self._client, self._subscription, **self.config)
         self._coordinator = ConsumerCoordinator(
-            self._client, self._subscription,
+            self._client, self._subscription, self._metrics, metric_group_prefix,
             assignors=self.config['partition_assignment_strategy'],
             **self.config)
         self._closed = False
         self._iterator = None
         self._consumer_timeout = float('inf')
 
-        #self.metrics = None
         if topics:
             self._subscription.subscribe(topics=topics)
             self._client.set_topics(topics)
@@ -277,7 +294,7 @@ class KafkaConsumer(six.Iterator):
         log.debug("Closing the KafkaConsumer.")
         self._closed = True
         self._coordinator.close()
-        #self.metrics.close()
+        self._metrics.close()
         self._client.close()
         try:
             self.config['key_deserializer'].close()
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -218,7 +218,7 @@ class KafkaConsumer(six.Iterator):
         reporters.append(DictReporter('kafka.consumer'))
         self._metrics = Metrics(metric_config, reporters)
         metric_group_prefix = 'consumer'
-        # TODO _metrics likely needs to be passed to KafkaClient, Fetcher, etc.
+        # TODO _metrics likely needs to be passed to KafkaClient, etc.
 
         self._client = KafkaClient(**self.config)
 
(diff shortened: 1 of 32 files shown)


kafka/metrics/kafka_metric.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @property
+    def measurable(self):
+        return self._measurable
+
+    def _get_config(self):
+        return self._config
+
+    def _set_config(self, config):
+        with self._lock:
```
dpkp: simple variable assignment is generally considered atomic in python. can we drop the lock here?

kafka/metrics/kafka_metric.py
```diff
@@ -0,0 +1,35 @@
...
+            self._config = config
+    config = property(_get_config, _set_config, None, None)
+
+    def value(self, time_ms=None):
+        if time_ms is None:
+            # with (self._lock): This doesn't seem necessary?
```
dpkp: agree -- dont see the point of using a lock here

kafka/metrics/measurable.py
```diff
@@ -0,0 +1,27 @@
...
+class AbstractMeasurable(object):
+    """A measurable quantity that can be registered as a metric"""
+    @abc.abstractmethod
+    def measure(self, config, now):
+        """
+        Measure this quantity and return the result as a double
+
+        Arguments:
+            config (MetricConfig): The configuration for this metric
+            now (int): The POSIX time in milliseconds the measurement
```
dpkp: what is the purpose of the now parameter?

kafka/metrics/metric_config.py
```diff
@@ -0,0 +1,23 @@
...
+        self.event_window = event_window
+        self.time_window_ms = time_window_ms
+        # tags should be OrderedDict (not supported in py26)
+        self.tags = tags if tags else {}
+
+    @property
+    def samples(self):
+        return self._samples
+
+    @samples.setter
```
dpkp: where does samples.setter decorator come from?

kafka/metrics/metric_config.py
```diff
@@ -0,0 +1,23 @@
+import sys
+
+
+class MetricConfig(object):
+    """Configuration values for metrics"""
```
dpkp: should we include argument description in this docstring?

kafka/metrics/metric_name.py
```diff
@@ -0,0 +1,104 @@
...
+    e.g. domainName:type=group,key1=val1,key2=val2
+
+    Usage looks something like this:
+
+        # set up metrics:
+        metric_tags = {'client-id': 'producer-1', 'topic': 'topic'}
+        metric_config = MetricConfig(tags=metric_tags)
+
+        # metrics is the global repository of metrics and sensors
+        metrics = new Metrics(metric_config)
```
dpkp: ~new

kafka/metrics/metrics.py
```diff
@@ -0,0 +1,258 @@
...
+    A metric is a named, numerical measurement. A sensor is a handle to
+    record numerical measurements as they occur. Each Sensor has zero or
+    more associated metrics. For example a Sensor might represent message
+    sizes and we might associate with this sensor a metric for the average,
+    maximum, or other statistics computed off the sequence of message sizes
+    that are recorded by the sensor.
+
+    Usage looks something like this:
+        # set up metrics:
+        metrics = Metrics # this is the global repository of metrics and sensors
```
dpkp: `Metrics()`

kafka/metrics/metrics.py
```diff
@@ -0,0 +1,258 @@
...
+                 enable_expiration=False):
+        """
+        Create a metrics repository with a default config, given metric
+        reporters and the ability to expire eligible sensors
+
+        Arguments:
+            default_config (MetricConfig, optional): The default config
+            reporters (list of AbstractMetricsReporter, optional):
+                The metrics reporters
+            enable_expiration (bool, optional): true if the metricsinstance
```
dpkp: sp: metricsinstance

kafka/metrics/metrics.py
```diff
@@ -0,0 +1,258 @@
...
+        self._reporters = reporters or []
+        for reporter in self._reporters:
+            reporter.init([])
+
+        if enable_expiration:
+            def expire_loop():
+                while True:
+                    # delay 30 seconds
+                    time.sleep(30)
+                    self.ExpireSensorTask.run(self)
```
dpkp: do we need to worry about circular reference here?

kafka/metrics/metrics.py
```diff
@@ -0,0 +1,258 @@
...
+                for this sensor for metrics that don't have their own config
+            inactive_sensor_expiration_time_seconds (int, optional):
+                If no value if recorded on the Sensor for this duration of
+                time, it is eligible for removal
+            parents (list of Sensor): The parent sensors
+
+        Returns:
+            Sensor: The sensor that is created
+        """
+        with self._lock:
```
dpkp: one way to avoid wrapping everything in a lock is to check if the sensor exists first; if it does, return it. if it doesn't, grab a lock and check if it exists again; if it does, release lock and return it; if it doesn't, create it, release lock and return it.

kafka/metrics/metrics.py
```diff
@@ -0,0 +1,258 @@
...
+        reporters and the ability to expire eligible sensors
+
+        Arguments:
+            default_config (MetricConfig, optional): The default config
+            reporters (list of AbstractMetricsReporter, optional):
+                The metrics reporters
+            enable_expiration (bool, optional): true if the metricsinstance
+                can garbage collect inactive sensors, false otherwise
+        """
+        self._lock = threading.RLock()
```
dpkp: do we really need RLock v. Lock?

kafka/consumer/group.py
```diff
@@ -181,6 +182,16 @@ def __init__(self, *topics, **configs):
                         new_config, self.config['auto_offset_reset'])
             self.config['auto_offset_reset'] = new_config
 
+        metrics_tags = {'client-id': self.config['client_id']}
+        metric_config = MetricConfig(samples=self.config['metrics_num_samples'],
+                                     time_window_ms=self.config['metrics_sample_window_ms'],
+                                     tags=metrics_tags)
+        reporters = [reporter() for reporter in self.config['metric_reporters']]
+        reporters.append(DictReporter('kafka.consumer'))
```
dpkp: should we just make this the default configuration for `metric_reporters` ?

kafka/metrics/metric_config.py
zackdever: introduced in py26. cool huh?

kafka/metrics/measurable.py
zackdever: i think it's partially to make testing easier, but from what i've seen in the comments it's also a slight optimization. when you record or read a stat, you pass in the time and it's reused through the calls to check for expired stats or anything else that needs a timestamp, rather than calling to get the current (and slightly advanced) timestamp each time it is used.

could possibly be removed since we can mock time, but i don't have a good feel for how it's used in instrumentation yet.

kafka/metrics/metric_config.py
zackdever: sure i can add them. i just copied over what was in the java code initially.

kafka/metrics/metrics.py
zackdever: shouldn't need to. `ExpireSensorTask.run` is a static method.

kafka/consumer/group.py
zackdever: i thought about it. i think the idea is to always guarantee that there is this 'default' reporter regardless of what is configured. probably harmless to move to a default config, but i thought i'd leave it 1:1 with the java implementation until we get more things instrumented and have a better feel for things.

kafka/metrics/kafka_metric.py
zackdever: will remove the lock entirely from KafkaMetric and leave a note that the java implementation uses one.

kafka/metrics/metrics.py
zackdever: the current implementation requires it (you can see this by changing them to regular Locks - the tests will hang). i'm looking at seeing what we can remove/change.

zackdever:
Assuming tests pass, I think this is ready to go. It provides a complete port of the java client's metric system, minus the JMX reporter. I haven't done much of the instrumentation, but have instrumented fetch requests and the consumer coordinator as a first pass example. My plan is to ship this with the next release, and then soon thereafter work on matching the java client's instrumentation.

Important: This doesn't publicly expose the metrics we're collecting just yet. Because the java client relies on JMX as the default built-in reporter, we need to make a design decision about how to handle that in kafka-python. I think a simple dictionary based approach would make sense, but would like to mess around with that some more before committing to a public interface. Towards this end I've made DictReporter that can export a snapshot of all stats at the current time and their values. That said, users can easily provide their own reporters through configuration to expose the metrics, just like in the java implementation. I've updated the documentation on `KafkaConsumer` to reflect this.

#424 and #505 both add some metric instrumentation and provide a couple ways of retrieving those metrics - one exposes a dictionary of metrics, and the other has a handler system. This PR provides the same reporter interface the java client provides, which I think is similar to the handler system in #505. My intent for the DictReporter is to make the default instance of it available through a public API similar to #424. I don't know if the java implementation has the same instrumentation as the two mentioned PRs, but I would plan to mirror the java instrumentation fairly closely, but there's no reason we couldn't add our own instrumentation if it made sense.

zackdever:
Will be interesting to watch https://issues.apache.org/jira/browse/KAFKA-3377 to see if a REST metrics interface gets added to Kafka.

PR #642, opened 2016-04-11
first pass at more thorough IPv6 support that uses getaddrinfo to resolve names

### LINKED ISSUE #641: Client never chooses IPv6 after bootstrapping
If you provide IPv6 addresses when creating the client (e.g., `bootstrap_servers=["[::1]:9092"]`) and the server responds with a hostname (e.g., `localhost`), this library always actually uses IPv4. This is because the `get_ip_port_afi` function, when passed a hostname, always returns and address family of `AF_INET`.

Ideally, I think this code should be restructured to use `getaddrinfo` and a loop to detect whether the host is available over v4 or v6. It looks like there's no actual resolution in `kafka/conn.py`; we're relying on implicit resolution done by passing a hostname to `socket.connect_ex`; hopefully that uses getaddrinfo behind the scenes and follows RFC 3484.

Roguelazer:
This is raw because, uh, I just wrote it. Haven't even run tests yet — wanted to see if this looked okay, or wether this is too much spaghetti.

Fixes #641

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -52,8 +52,8 @@ class ConnectionStates(object):
     CONNECTED = '<connected>'
 
 
-InFlightRequest = collections.namedtuple('InFlightRequest',
-    ['request', 'response_type', 'correlation_id', 'future', 'timestamp'])
+InFlightRequest = collections.namedtuple('InFlightRequest', [
+    'request', 'response_type', 'correlation_id', 'future', 'timestamp'])
 
 
 class BrokerConnection(object):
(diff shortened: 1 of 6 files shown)


Roguelazer:
I don't actually think it's a problem to use standard `getaddrinfo` and be subject to the 5s minimum timeout in the event that the user has no working resolvers (anyone who's using a hostname in configuration and _doesn't_ understand how libc does name resolution is probably going to find out in an outage soon, anyway), but I included that comment for completeness.

Roguelazer:
These integration tests are pretty flaky. @dpkp, I think this code works now, although one of the Kafka instances in the integration test failed to start, so the test is red. Let me know what you think and how you'd like me to squash these commits, presuming that the approach looks good to you.

Roguelazer:
@dpkp anything you'd like me to change?

dpkp:
Great - you got all tests to pass! I'd prefer that the style changes were separated, but I can probably do that myself w/ a little manual merge magic.

kafka/conn.py
```diff
@@ -548,37 +591,66 @@ def __repr__(self):
...
-            * host_or_ip:port     <- This is only for IPv4
-            * [host_or_ip]:port.  <- This is only for IPv6
+        :return: either socket.AF_INET or socket.AF_INET6 or socket.AF_UNSPEC if the address family
+                 could not be determined
+    """
+    if address.startswith('[') and address.endswith(']'):
+        return socket.AF_INET6
+    for af in (socket.AF_INET, socket.AF_INET6):
+        try:
+            socket.inet_pton(af, address)
```
dpkp: this is not available on windows. can you do this in a platform-compatible way?

dpkp:
Rebased / squashed, dropped unrelated style changes, and added checks for socket.inet_pton b/c it isn't supported on Windows.

Merged to master at b06e423

PR #644, opened 2016-04-12
pass request timeout into the check version call when initialising a Producer

eastlondoner:
Because, believe it or not, people's network performance varies

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -246,7 +246,8 @@ class KafkaProducer(object):
 
         # Check Broker Version if not set explicitly
         if self.config['api_version'] == 'auto':
-            self.config['api_version'] = client.check_version()
+            request_timeout_seconds = int(self.config['request_timeout_ms'] / 1000)
+            self.config['api_version'] = client.check_version(timeout=request_timeout_seconds)
         assert self.config['api_version'] in ('0.9', '0.8.2', '0.8.1', '0.8.0')
 
         # Convert api_version config to tuple for easy comparisons


dpkp:
hah - snark! unfortunately I won't merge this. I'm willing to consider other approaches, but using the default request_timeout_ms is not ok (indeed, that is why there's a timeout parameter internally). This would mean that by default the check_version in consumer and producer instantiation could block up to 40secs! Unless you can prove that there is no broker compatibility case that will cause a timeout (I am not convinced of this), then I think the timeout needs to be small here.

My suggestion remains -- if you are concerned about network latency, do not depend on dynamic broker version checks. The approach I've taken here is very controversial and is not explicitly supported by the kafka server development team. It is, in other words, a HACK.

eastlondoner:
Fair enough, I was going for the smallest change.

I am kind of intrigued about this because I use Kafka libraries in other languages and they don't require me to tell them the kafka version.  BUT I don't know if that's just because they aren't making use of latest features or because they have a supported way of determining the kafka version.

Would adding another parameter version_check_timeout and setting it to default to 2 (as it currently does) be ok? It would at least then be externally configurable.

dpkp:
I would merge a version check timeout config!

I have somewhat unique views on version checking in the Kafka community. There is a long running debate on the dev list about how to make this easier, see KIP-35. No other clients have been willing to take the approach I've used here, mai ly because the sercer behavior is not well defined, well understood or well tested in this space. The official client requires users to manually verify that their client library is older than their server. They generally do not go back and improve old clients -- you must upgrade server for improvements.

I think the community is better served if a single client can operate with older brokers relatively seamlessly (minus certain features of course). I am working w librdkafka and others to convince server team to officially support a method to get the broker version dynamically. If that gets merged, I plan to switch check_version to a more sane approach that would be able to reasonably handle network latency issues (at least for all newer brokers)

zackdever:
Updated PR at #647

PR #647, opened 2016-04-13
allow client.check_version timeout to be set in Producer and Consumer constructors

eastlondoner:
No unit tests, sorry, it didn't seem like something that would be easy to add.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -5,6 +5,7 @@ import logging
 import io
 from random import shuffle
 import socket
+import ssl
 import struct
 from threading import local
 import time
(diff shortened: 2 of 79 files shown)


kafka/consumer/group.py
```diff
@@ -128,6 +128,9 @@ class KafkaConsumer(six.Iterator):
             offset commits; 0.8.0 is what is left. If set to 'auto', will
             attempt to infer the broker version by probing various APIs.
             Default: auto
+        client_check_version_timeout_s (int): number of seconds to throw a
```
zackdever: i'd vote for milliseconds for consistency with the rest of the configs

eastlondoner: I know, the timeout argument already written in to client.check_version accepts seconds.

it seems ugly to convert so if we want ms then I'll change client.check_version to use ms

eastlondoner:
As far as I can tell, one of the CI builds failed because of a timeout, which is probably nothing to do with the code change in the PR...

zackdever:
Yep I think so. I kicked off the test again. Original log is at https://s3.amazonaws.com/archive.travis-ci.org/jobs/124422979/log.txt

eastlondoner:
proves my point about timeouts!

Pls merge or comment :-D

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dpkp: I'd rather not commit this file to the git repo

eastlondoner: bah!

eastlondoner:
sorry about the .idea file that snuck in - removed and updated PR

kafka/conn.py
```diff
@@ -487,7 +487,7 @@ def connect():
             self.connect()
             if self.connected():
                 return
-            timeout_at = time.time() + timeout
+            timeout_at = time.time() + (timeout_ms/1000)
```
zackdever: `(timeout_ms / 1000.0)` so we don't lose precision.

That said, I'd personally rather keep the calls as `timeout` instead of `timeout_ms` to reduce changes, but keep the config as `_ms` to stay consistent with the java configs. Not a huge deal though.

dpkp:
I need to think about what I think the best config name should be here. client_check_version_timeout_ms is probably not right -- version checks will be moving to per-connection once brokers add support. I'd like to stay consistent w/ librdkafka and java client as best we can. KIP-35 is about to land and I expect that to impact the way we manage version checks substantially.

eastlondoner:
Hey guys, any chance of merging this in? I've been using it in production for a while and it works fine for me.

dpkp:
Can you address the merge conflicts? I can't merge the code via the github UI.

Also, I'd prefer if the config parameter were named `api_version_auto_timeout_ms`

eastlondoner:
sure. Realised I hadn't pushed the merge fixes, sorry

dpkp:
Thanks!

PR #649, opened 2016-04-14
to squelch the __del issue

jayalane:


dpkp:
Pls use the close method, not os.close() . the wake fds are now sockets.

jayalane:
Ping

PR #650, opened 2016-04-14
Quick ugly fix Python 3 compatibility

dronnix:
conn.py: 'ret' exception variable is local in Pytthon3. I think is better to rewrite connect method with self._sock.connect() to avoid handling errno & exceptions both. Let me know if this acceptable, will think about more elegant fix.

message.py: value is str. Can't figure out why. 

Fixes tested in Python 3.4.3 & 2.7.6.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -120,11 +120,12 @@ class BrokerConnection(object):
         if self.state is ConnectionStates.CONNECTING:
             # in non-blocking mode, use repeated calls to socket.connect_ex
             # to check connection status
+            ret = None
             request_timeout = self.config['request_timeout_ms'] / 1000.0
             try:
                 ret = self._sock.connect_ex((self.host, self.port))
-            except socket.error as ret:
-                pass
+            except socket.error as e:
+                ret = e
 
             # Connection succeeded
             if not ret or ret == errno.EISCONN:
--- a/kafka/protocol/message.py
+++ b/kafka/protocol/message.py
@@ -25,6 +25,8 @@ class Message(Struct):
     HEADER_SIZE = 14 # crc(4), magic(1), attributes(1), key+value size(4*2)
 
     def __init__(self, value, key=None, magic=0, attributes=0, crc=0):
+        if isinstance(value, str):
+            value = value.encode()
         assert value is None or isinstance(value, bytes), 'value must be bytes'
         assert key is None or isinstance(key, bytes), 'key must be bytes'
         self.crc = crc


dpkp:
Use value_serializer to handle encoding messages to bytes. I don't think i agree with default str encode.

Can you write a test case to show the issue you are concerned with ret variable?

dronnix:
> Can you write a test case to show the issue you are concerned with ret variable?


Sure:

```
#!/usr/bin/env python3

if __name__ == '__main__':
    producer = KafkaProducer(bootstrap_servers='non-existent.test:9092')
    producer.send('foo', 'bar').get(timeout=100)
    producer.flush()

```
got:
```
  File "./producer.py", line 6, in <module>
    producer = KafkaProducer(bootstrap_servers='non-existent.test:9092')
  File "/home/dronnix/venv/kafka/lib/python3.4/site-packages/kafka/producer/kafka.py", line 266, in __init__
    client = KafkaClient(**self.config)
  File "/home/dronnix/venv/kafka/lib/python3.4/site-packages/kafka/client_async.py", line 130, in __init__
    self._bootstrap(collect_hosts(self.config['bootstrap_servers']))
  File "/home/dronnix/venv/kafka/lib/python3.4/site-packages/kafka/client_async.py", line 156, in _bootstrap
    bootstrap.connect()
  File "/home/dronnix/venv/kafka/lib/python3.4/site-packages/kafka/conn.py", line 130, in connect
    if not ret or ret == errno.EISCONN:
UnboundLocalError: local variable 'ret' referenced before assignment
Exception ignored in: <bound method KafkaClient.__del__ of <kafka.client_async.KafkaClient object at 0x7fd09eccb668>>
  File "/home/dronnix/venv/kafka/lib/python3.4/site-packages/kafka/client_async.py", line 136, in __del__
    self._wake_r.close()
AttributeError: 'KafkaClient' object has no attribute '_wake_r'
```
The reason is described in [PEP-3110](https://www.python.org/dev/peps/pep-3110/):

> This PEP intends to resolve this issue by adding a cleanup semantic to except clauses in Python 3 whereby the target name is deleted at the end of the except suite.

Roguelazer:
This issue is also addressed in #642, since I exclusively run on py3.5

kafka/protocol/message.py
```diff
@@ -25,6 +25,8 @@ class Message(Struct):
     HEADER_SIZE = 14 # crc(4), magic(1), attributes(1), key+value size(4*2)
 
     def __init__(self, value, key=None, magic=0, attributes=0, crc=0):
+        if isinstance(value, str):
```
dpkp: the interface so far has been that message values should be bytes

dpkp:
re: serialization, `producer.send('foo', 'bar')` should be `producer.send('foo', b'bar')` or otherwise you should specify a value_serializer when initializing the KafkaProducer

kafka/conn.py
```diff
@@ -120,11 +120,12 @@ def connect(self):
         if self.state is ConnectionStates.CONNECTING:
             # in non-blocking mode, use repeated calls to socket.connect_ex
             # to check connection status
+            ret = None
             request_timeout = self.config['request_timeout_ms'] / 1000.0
             try:
```
dpkp: I dont think we even need the try / except here - connect_ex should never raise an exception.

dpkp: strike that -- I believe your test case shows that an exception is raised on dns lookup failures:
```
gaierror: [Errno 8] nodename nor servname provided, or not known
```

dpkp:
> This issue is also addressed in #642, since I exclusively run on py3.5

Cool -- #642 is a bit more comprehensive, so I'm going to close this. @dronnix: please see my notes above re: message serialization.

dronnix:
OK, sorry for wasting Your time

dpkp:
not a waste of time! thank you for raising this issue. we obviously missed it in our initial testing, so this was a clear improvement for py3 users. thanks again

PR #664, opened 2016-04-25
Consume-produce integration test with Snappy codec

sibiryakov:
Hi guys,
Before 1.0.x series consumer was able to detect if message was compressed and use the necessary decompression codec. During refactoring from `kafka.common` to `kafka.structs` this feature was lost. That breaks backwards compatibility. It would be nice if we could merge this test along with getting back lost feature, or figure some other way how to fix it.

### DIFF
--- a/.idea/encodings.xml
+++ b/.idea/encodings.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="Encoding">
+    <file url="PROJECT" charset="UTF-8" />
+  </component>
+</project>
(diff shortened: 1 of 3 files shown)


dpkp:
Is this just for the Simple* interfaces? Have you looked at https://github.com/dpkp/kafka-python/blob/master/test/test_producer.py#L25 ? We do end-to-end testing of produce / consume using all compression types.

sibiryakov:
I didn't tested it for Kafka* interfaces, but it's broken for Simple* as you see. This test is passing on my local machine with kafka-python 0.9.5.

dpkp:
Did #684 fix your issue?

sibiryakov:
It looks like it works now!

PR #678, opened 2016-05-03
Add protocol support for ApiVersionRequest

dpkp:
Use the api to identify 0.10 brokers in check_version.

Leaving deeper support for ApiVersion response (min / max versions) for a separate task.

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -225,7 +225,7 @@ class KafkaConsumer(six.Iterator):
         # Check Broker Version if not set explicitly
         if self.config['api_version'] == 'auto':
             self.config['api_version'] = self._client.check_version()
-        assert self.config['api_version'] in ('0.9', '0.8.2', '0.8.1', '0.8.0'), 'Unrecognized api version'
+        assert self.config['api_version'] in ('0.10', '0.9', '0.8.2', '0.8.1', '0.8.0'), 'Unrecognized api version'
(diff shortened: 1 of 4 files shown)


kafka/consumer/group.py
```diff
@@ -225,7 +225,7 @@ def __init__(self, *topics, **configs):
         # Check Broker Version if not set explicitly
         if self.config['api_version'] == 'auto':
             self.config['api_version'] = self._client.check_version()
-        assert self.config['api_version'] in ('0.9', '0.8.2', '0.8.1', '0.8.0'), 'Unrecognized api version'
+        assert self.config['api_version'] in ('0.10', '0.9', '0.8.2', '0.8.1', '0.8.0'), 'Unrecognized api version'
```
dpkp: This should probably be refactored. A user should not ever need to pass in api_version='0.10' since that broker supports ApiVersionRequest.

zackdever:
Is this missing the `check_version` change?

kafka/consumer/group.py
zackdever: But the user still _could_ set api_version='0.10' (or higher in the future) right?

zackdever:
Woot! LGTM

kafka/consumer/group.py
dpkp: I'm convinced! merging

PR #682, opened 2016-05-09
kafka/conn: use original hostname for SSL checks

vincentbernat:
When the address family is not provided, `self.host` is resolved to one
of the IP addresses and replaced by it. The SSL context is then built
using `self.host` which is now an IP instead of the proper name. Most of
the time, hostname cannot be checked this way. Therefore, save the
original hostname in a dedicated property and use this property for the
SSL context.

Only lightly tested with `example.py`.


dpkp:
Thanks!

PR #683, opened 2016-05-09
Add CRL support to SSL support

vincentbernat:
A user can provide a CRL whose peer certificate will be checked
against. This only works with Python 3.4+ and Python 2.7.9+.

Lightly tested with `example.py`.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -70,6 +70,7 @@ class BrokerConnection(object):
         'ssl_cafile': None,
         'ssl_certfile': None,
         'ssl_keyfile': None,
+        'ssl_crlfile': None,
         'api_version': (0, 8, 2),  # default to most restrictive
         'state_change_callback': lambda conn: True,
     }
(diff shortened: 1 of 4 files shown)


kafka/conn.py
```diff
@@ -220,6 +221,15 @@ def _wrap_ssl(self):
                 self._ssl_context.load_cert_chain(
                     certfile=self.config['ssl_certfile'],
                     keyfile=self.config['ssl_keyfile'])
+            if self.config['ssl_crlfile']:
+                if not hasattr(ssl, 'VERIFY_CRL_CHECK_LEAF'):
+                    log.error('No CRL support with this version of Python.'
+                              ' Disconnecting.', self)
```
dpkp: Travis tests failing on this log line -- assume you meant to prefix the log line with `s%: ` and pass `str(self)` ?

dpkp:
LGTM -- please fix the log line that's failing tests and I'll merge.

kafka/conn.py
vincentbernat: Yes, you are right. However, this code path shouldn't have hit Travis. Let me investigate a bit. I keep `self` since it should be the same as `str(self)` and most of the code in the same file uses `self`.

vincentbernat: Ok, this is a pylint error. I corrected the error. Let's see if this works with Travis.

dpkp:
Odd -- pylint is throwing a no-member error on ssl.VERIFY_CRL_CHECK_LEAF for python3.5 ... any idea what's happening there?

vincentbernat:
Maybe it's dependant on the version of OpenSSL (but why would it work with Python 3.4). Let me try to reproduce this locally on a system running Precise. Also, I will add checks for the error on 3.3.

vincentbernat:
This seems an issue with pylint: https://github.com/PyCQA/pylint/issues/399. Should I make pylint ignore the error for now? I am pushing the fix for Python 3.3 since it is right for this one.

vincentbernat:
For Python 3.3, the path is dead code since it doesn't have the `VERIFY_CRL_CHECK_LEAF` either. I have added a pylint ignore directive that happens to cover both member checks. Test should be OK now.

vincentbernat:
The current errors seem unrelated to this PR. Would it be possible to retrigger the three failed tests?

dpkp:
Thanks!

PR #728, opened 2016-06-17
Use weakref when registering a producer.close atexit to fix normal gc

### LINKED ISSUE #727: KafkaProducer is not garbage collected
This only affects the KafkaProducer and is caused by the extra system reference caused by `atexit.register(self.close)`. We should use a weakref proxy to allow gc to function normally on producer objects.

dpkp:
Fix #727

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -5,12 +5,13 @@ import copy
 import logging
 import threading
 import time
+import weakref
 
+from .. import errors as Errors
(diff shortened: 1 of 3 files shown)


dpkp:
Mike reminds me that this leaks a weakref proxy. There's no way to unregister a handler, so we'd probably need to maintain a global set of active producer weakref proxies, adding on `__init__` and popping on `__del__`

Issue #560, opened 2016-02-20
Twisted interface


glyph:
Kafka-Python claims to be "async" but this seems to just mean "uses lots of threads to put your I/O into the background" and not actually "event-driven".  It would be nice to have an interface which avoids the overhead (thread stacks, GIL switching, etc) and risks (sharing too much state, deadlocks, difficult to debug differing behavior at scale) of threads, but presents a similar interface.

I filed the ticket with the summary "twisted" because of my particular biases, but this should be done in a way which is as framework-neutral as possible, since supporting the `asyncio` stdlib module at some point in the future would probably be interesting as well, not to mention Tornado and Vanilla and so on.

Given that one would be consuming a queue, perhaps this would be a good landmark project for https://github.com/twisted/tubes which, despite its github URL, has an (almost) framework-neutral architecture and lots of good abstractions around flow control and propagating backpressure.

glyph:
I have no idea what the implications for the internals would be yet (I haven't looked too closely at the code) but in terms of the external interface, it would seem that there would be two core changes necessary:

1. an interface to `KafkaConsumer` where the values are delivered to a callable rather than as the results of an iterable, like

 ```python
 consumer = KafkaConsumer(value_deserializer=msgpack.dumps)
 @consumer.consume_into
 def a_message_arrived(message):
    "process it"
 ```

2. an interface to `KafkaProducer` where `send` produces a `Deferred` (or asyncio or Tornado `Future`)

paulproteus:
This sounds great; I believe it would let people use Kafka along with the upcoming Django channels work!

dpkp:
The goal of this project is to stay closely aligned with the official apache client. I do not intend to implement a different architecture or make significant changes to the interface. The reason for that is not that I think one model is better than another (though you probably should read the code more closely before making assertions about event loops and thread use). It is because maintaining a Kafka client that provides all the features that users expect is difficult. There are not a lot of unofficial drivers that come close. Staying in step with the Apache project is, in my opinion, the best approach until Kafka itself is fully matured and you won't need to constantly keep up with protocol changes, client state issues, and requirements necessary to provide ordering guarantees, coordination, etc.

I welcome you to explore what it would take to deliver a viable kafka client for asyncio or twisted (or any other event loop framework). I am happy to comment on PRs / forks / POCs / etc. If you can provide unobtrusive wrapper classes to bridge the gaps, I would consider merging. Otherwise I think you might consider writing a new client txkafka.

[...]

PR #735, opened 2016-06-21
Allow injecting clients into Kafka Consumers and Producers.

Julian:
We use the older interfaces and are considering evaluating the newer ones, but in order to do any testing, we need to be able to inject fake clients into consumers and producers to prevent them from doing networking.

It would also be nice to prevent some of the other side effects that happen in **init** on these newer objects, but that work isn't done yet here.

Love to hear your thoughts obviously. Cheers.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/735)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -46,7 +46,10 @@ class KafkaConsumer(six.Iterator):
             It just needs to have at least one broker that will respond to a
             Metadata API Request. Default port is 9092. If no servers are
             specified, will default to localhost:9092.
-        client_id (str): A name for this client. This string is passed in
+        client (kafka.client_async.KafkaClient): a kafka client to
+            use, or if unprovided, one is constructed from the provided
+            configuration.
+        client_id (str): a name for this client. This string is passed in
             each request to servers and can be used to identify specific
             server-side log entries that correspond to this client. Also
             submitted to GroupCoordinator for logging with respect to
@@ -243,6 +246,7 @@ class KafkaConsumer(six.Iterator):
     """
     DEFAULT_CONFIG = {
         'bootstrap_servers': 'localhost',
+        'client': None,
         'client_id': 'kafka-python-' + __version__,
         'group_id': None,
         'key_deserializer': None,
@@ -338,7 +342,11 @@ class KafkaConsumer(six.Iterator):
             log.warning('use api_version=%s [tuple] -- "%s" as str is deprecated',
                         str(self.config['api_version']), str_version)
 
-        self._client = KafkaClient(metrics=self._metrics, **self.config)
+        client = self.config.pop('client', None) or KafkaClient(
+            metrics=self._metrics,
+            **self.config
+        )
+        self._client = client
 
         # Get auto-discovered version from client if necessary
         if self.config['api_version'] is None:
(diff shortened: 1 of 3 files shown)


kafka/consumer/group.py
```diff
@@ -201,6 +204,8 @@ class KafkaConsumer(six.Iterator):
     }
 
     def __init__(self, *topics, **configs):
+        client = configs.pop("client", None)
```
dpkp: The pattern we've been using so far is to put this in the class-level DEFAULT_CONFIG dict. Just add an entry for `'client': None`

kafka/producer/kafka.py
```diff
@@ -253,7 +256,7 @@ class KafkaProducer(object):
         'api_version': 'auto',
     }
 
-    def __init__(self, **configs):
+    def __init__(self, client=None, **configs):
```
dpkp: same as above -- you should just add `'client': None` to the class DEFAULT_CONFIG dict

Julian:
Moved. Thanks.

dpkp:
Thanks -- I'll wait for tests to pass and then merge.

kafka/producer/kafka.py
```diff
@@ -270,7 +274,7 @@ def __init__(self, **configs):
         if self.config['acks'] == 'all':
             self.config['acks'] = -1
 
-        client = KafkaClient(**self.config)
+        self.client = self.config['client'] or KafkaClient(**self.config)
```
dpkp: this should just be `client = ` not `self.client`

kafka/consumer/group.py
```diff
@@ -226,7 +230,8 @@ def __init__(self, *topics, **configs):
         metric_group_prefix = 'consumer'
         # TODO _metrics likely needs to be passed to KafkaClient, etc.
 
-        self._client = KafkaClient(**self.config)
+        client = self.config.pop('client', None) or KafkaClient(**self.config)
+        self._client = client
```
dpkp: I'd probably just combine these into a one-line assignment

dpkp:
tests are failing

jeffwidman:
@Julian I know this is super late, but are you still interested in updating this or should we close it?

I think it's a good idea, although if your rationale is injecting fake clients purely for testing purposes, the work of #1032 / #1230 (rationale in #560) may make your life easier for testing the protocol layer separate from the network layer.

Julian:
@jeffwidman that looks promising! It looks like I've already seen #560 before (we currently use afkak for the Twisted case) so definite +1 from me. Afkak's internal interfaces are very clearly super close to what's in here, so being able to separate the protocol parts separately definitely sounds like it'd help.

Given that we've never been able to upgrade to the newer kafka-python versions/clients because of this issue (and have stuck with the older versions) I *suspect* it's still relevant for us, but will have to spend a few minutes reviewing the patch here and reviewing our Kafka fake (which I'd also love to upstream :).

I suspect I would still like to get this merged if possible, so assuming that's the case, I'll have a look and see where this is.

dpkp:
Apologies, our test suite was broken due to upstream dependency changes in pylint and lz4. I've pinned the versions in master. Do you mind doing a rebase or merge and push for a new test run?

Julian:
@dpkp done

jeffwidman:
Afraid tests are failing...

PR #750, opened 2016-07-07
Add ssl_password to default_config dicts. Send ssl_password when loading cert chains

amckemie:
Our team wanted to be able to set an ssl_password in config so a KafkaConsumer could be created with SSL certification and not need to enter a password manually. This minor changes add the ssl_password in all the default config dicts necessary and allow for it to be set when instantiating a consumer instance.


dpkp:
Thanks!

PR #756, opened 2016-07-12
Fix KafkaConsumer autocommit for 0.8 brokers

### LINKED ISSUE #706: Support KafkaConsumer auto-commit with 0.8 brokers
kafka 0.8.2    kafka-python 1.1.1

when enable auto_commit, an AutoCommitTask instance will be created
but when Enable the AutoCommitTask instance?
in the code , only find the function _on_join_complete will enable  the AutoCommitTask instance

```
def _on_join_complete(self, generation, member_id, protocol,
                          member_assignment_bytes):
        ...
        # restart the autocommit task if needed
        if self._auto_commit_task:
            self._auto_commit_task.enable()
```

but api_version < 0.9, the function _on_join_complete never be called

dpkp:
Fixes #706

### DIFF
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -50,6 +50,7 @@ class BaseCoordinator(object):
         'session_timeout_ms': 30000,
         'heartbeat_interval_ms': 3000,
         'retry_backoff_ms': 100,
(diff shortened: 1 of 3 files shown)


sibiryakov:
Thank you @dpkp !

Issue #718, opened 2016-06-03
KafkaProducer produces corrupt "double-compressed" messages on retry when compression is enabled. KafkaConsumer gets "stuck" consuming them


zoltan-fedor:
This is an interesting one.

In all of our topics every day a handful of partitions get "stuck".
Basically the reading of the partition stops at a given message and kafka-python reports that there are no more messages in the given partition (just like it would have consumed all messages), while there are unconsumed messages.
The only way to get the consumers moving again is to manually seek the offset forward by stepping over the "stuck" messages and then works again for a few million records and then get stuck again at some later offset.

I have multiple consumers consuming from the same topic and they all get stuck at the same messages of the same topics. Random number of partitions are affected day-to-day.

We are using Kafka broker version 0.9.0.1, kafka-python 1.2.1 (had the same issue with 1.1.1).

The consumer code is very simple (the below code is trying to read only partition #1, which is currently "stuck"):

```
kafka_consumer = KafkaConsumer(
    group_id=kafka_group_id,
    bootstrap_servers=kafka_servers,
    enable_auto_commit=False,
    consumer_timeout_ms=10000,
    fetch_max_wait_ms=10*1000,
    request_timeout_ms=10*1000
)
topics = [TopicPartition(topic, 1)]
kafka_consumer.assign(topics)

for message in kafka_consumer:
    print(message)

print("Completed")
```

The above code prints "Completed", but not the messages, while there is a 5M offset lag in partition 1, so there would be plenty of messages to read.
After seeking the consumer offset forward the code works again until it doesn't get "stuck" again.

dpkp:
This can happen to a kafka system if there is a message larger than your consumer's max_partition_fetch_bytes . Have you tried tuning that?

zoltan-fedor:
Thanks. The max_partition_fetch_bytes is already set to pretty high (10*1048576), so I don't think that would be the issue, but I will try to increase it even further to verify.

zoltan-fedor:
Unfortunately as expected, the max_partition_fetch_bytes did not make any difference, even increasing it 1000-fold did make the consumer read the message from the partition.

So with having:
`max_partition_fetch_bytes=1000*1048576`

The issue is the same as above.

dpkp:
Can you get debug logs for a consumer in this "stuck" state?

zoltan-fedor:
Sure, here comes the debug log of the consumer when it is in this "stuck" state.
[...]

PR #757, opened 2016-07-13
Fix bug causing KafkaProducer to double-compress message batches

dpkp:
This addresses producer-side bug from #718 .

### DIFF
--- a/kafka/producer/buffer.py
+++ b/kafka/producer/buffer.py
@@ -89,22 +89,29 @@ class MessageSetBuffer(object):
         return self._buffer.tell() >= self._batch_size
 
     def close(self):
-        if self._compressor:
-            # TODO: avoid copies with bytearray / memoryview
-            self._buffer.seek(4)
-            msg = Message(self._compressor(self._buffer.read()),
-                          attributes=self._compression_attributes,
-                          magic=self._message_version)
-            encoded = msg.encode()
-            self._buffer.seek(4)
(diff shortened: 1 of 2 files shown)


zackdever:
Looks like lz4tools>=1.2.1 is broken for python 2.6. Could try using lz4tools==1.2.0 just for python2.6.

```
$ python --version
Python 2.6.9
$ pip freeze | grep lz4
DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
lz4tools==1.2.1
$ python -c "import lz4f"
  File "<string>", line 1, in <module>
ImportError: dlopen(/Users/zdever/.virtualenvs/tmp26/lib/python2.6/site-packages/lz4f.so, 2): Symbol not found: _PyCapsule_GetPointer
  Referenced from: /Users/zdever/.virtualenvs/tmp26/lib/python2.6/site-packages/lz4f.so
  Expected in: flat namespace
 in /Users/zdever/.virtualenvs/tmp26/lib/python2.6/site-packages/lz4f.so
$ pip freeze | grep lz4
DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
lz4tools==1.2.0
$ python -c "import lz4f"
$
```

`tox -e py26` passes with the pinned `lz4tools==1.2.0` in tox.ini

PR #777, opened 2016-07-23
Change _pickle_method to use python3 special attributes

jpaulodit:
Issue: _pickle_method() uses attributes are not python 3 compatible.
Fix: updated to use python 3 special attributes.


dpkp:
Thanks!

PR #778, opened 2016-07-25
made errors more obvious

cookingkode:
It was very difficult for me to debug various broker errors. Adding string variants instead of nos help in easier debugging

### DIFF
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -292,7 +292,7 @@ class BaseCoordinator(object):
 
     def _failed_request(self, node_id, request, future, error):
         log.error('Error sending %s to node %s [%s]',
-                  request.__class__.__name__, node_id, error)
+                  request.__class__.__name__, node_id,  type(error).__name__)
         # Marking coordinator dead
         # unless the error is caused by internal client pipelining
         if not isinstance(error, (Errors.NodeNotReadyError,


cookingkode:
The failure is a KakfaServer Unavailable error..

dpkp:
Can you give an example of the error logging you were having trouble with? I had thought the class name was always included. The standard error formatting is:
```
[Error {errorno}] {__name__}: {msg}
```

cookingkode:
Well it's not..

For example: I get NodeNorReadyError on every fresh boot. I guess connections are  created on demand. Thus when initially requests go out for things like GroupCo-ordinator etc, then the client finds no connection https://github.com/dpkp/kafka-python/blob/master/kafka/client_async.py#L272

The error I get is something like this
2016-07-25 18:38:09,028.28.156042099:kafka.coordinator:123145306509312:ERROR:30790:Error sending GroupCoordinatorRequest_v0 to node 2 [1]

With the change it would appear as 

2016-07-25 18:38:09,028.28.156042099:kafka.coordinator:123145306509312:ERROR:30790:Error sending GroupCoordinatorRequest_v0 to node 2 [NodeNotReadyError]

dpkp:
Are you using master? That error formatting was updated in #760

cookingkode:
i was using 1.2.5  . Will migrate to the 1.2.6 when it comes. thanks!

Can you also please verify about my hypothesis about NodeNotReadyError? That it's benign and it's expected that connections are built on-demand and thus initial errors are expected..

dpkp:
Yes, that is correct.

PR #779, opened 2016-07-25
Support for PLAIN sasl authentication

larsjsol:
This PR adds support sasl authentication using the PLAIN mechanism.

I've tested this on a kafka 0.10 cluster.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -311,7 +311,7 @@ class BrokerConnection(object):
             self._sasl_auth_future = future
         self._recv()
         if self._sasl_auth_future.failed():
-            raise self._sasl_auth_future.exception
+            raise self._sasl_auth_future.exception # pylint: disable-msg=raising-bad-type
         return self._sasl_auth_future.succeeded()
 
     def _handle_sasl_handshake_response(self, future, response):
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -294,7 +294,7 @@ class BrokerConnection(object):
         if self.config['security_protocol'] == 'SASL_PLAINTEXT':
             log.warning('%s: Sending username and password in the clear', str(self))
 
-        # build a SaslHandShakeRequest message
+        # Build a SaslHandShakeRequest message
         correlation_id = self._next_correlation_id()
         request = SaslHandShakeRequest[0](self.config['sasl_mechanism'])
         header = RequestHeader(request,
@@ -304,7 +304,7 @@ class BrokerConnection(object):
         message = b''.join([header.encode(), request.encode()])
         size = Int32.encode(len(message))
 
-        # attempt to send it over our socket
+        # Attempt to send it over our socket
         try:
             self._sock.setblocking(True)
             self._sock.sendall(size + message)
@@ -323,7 +323,7 @@ class BrokerConnection(object):
                               timestamp=time.time())
         self.in_flight_requests.append(ifr)
 
-        # listen for a reply and check that the server supports the PLAIN mechanism
+        # Listen for a reply and check that the server supports the PLAIN mechanism
         response = None
         while not response:
             response = self.recv()
@@ -340,7 +340,7 @@ class BrokerConnection(object):
         data = b''
         try:
             self._sock.setblocking(True)
-            # send our credentials
+            # Send our credentials
             msg = bytes('\0'.join([self.config['sasl_username'],
                                    self.config['sasl_username'],
                                    self.config['sasl_password']]).encode('utf-8'))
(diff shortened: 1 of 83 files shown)


larsjsol:
It looks like it failed one of the tests (python 3.4, kafka 0.10) with the error
 `Exception: Failed to start KafkaInstance before max_timeout`.

Travis ran all tests successfully when i pushed so I assume that this is not an issue with the code.

kafka/conn.py
```diff
@@ -70,6 +71,9 @@ class BrokerConnection(object):
         'ssl_password': None,
         'api_version': (0, 8, 2),  # default to most restrictive
         'state_change_callback': lambda conn: True,
+        'sasl_mechanism': None,
+        'sasl_username': None,
+        'sasl_password': None
```
dpkp: I dont see sasl.username / sasl.password in the java kafka docs. I see some reference to plain user/pass stored in Jaas configuration files. Is that the standard method for java?

For kafka-python, should we add plain_ : `sasl_plain_username` etc?

kafka/conn.py
```diff
@@ -263,6 +278,90 @@ def _try_handshake(self):
...
+                                   self.config['sasl_username'],
+                                   self.config['sasl_password']]).encode('utf-8'))
+            size = Int32.encode(len(msg))
+            self._sock.sendall(size + msg)
+
+            # The server will send a zero sized message (that is Int32(0)) on success.
+            # The connection is closed on failure
+            received_bytes = 0
+            while received_bytes < 4:
+                data = data + self._sock.recv(4 - received_bytes)
```
dpkp: here and above you've chosen to do inline calls to `recv()` and block until the response is available. Can we use a callback and let the client selector poll for sockets ready to read? Or is there something about the SASL mechanism that prevents that?

dpkp:
Thanks alot, this is really great! I've got one design question inline about inline recv / blocking for a response. I added a note about username/password configuration names. And I think the other bit that I can probably add myself once this is merged is checking for appropriate api_versions. I'm not sure whether this would work on an 0.9 cluster, but I know it doesn't work on 0.8. So I'll want to raise an error quickly if the broker version is too old.

dpkp:
Re: failed test, that is a bug in our fixtures that sometimes doesn't bring up brokers correctly. I restarted the test.

kafka/conn.py
larsjsol: I did not consider using a callback, but that would probably work. 

My understanding is that no other traffic must come between the server sending a SaslHandshakeResponse and the actual authentication. I think we can talk to other brokers while this goes on though. 

I'll try to look into it tomorrow.

kafka/conn.py
larsjsol: > Is that the standard method for java?

As in "standard authentication method"? No. It's either PLAIN or GSSAPI (kerboros) which as far as I know does not use the terms "username" and "password". The (client) [config file](http://kafka.apache.org/documentation.html#security_sasl_plain_clientconfig) use the terms "username" and "password" with no prefix. 

But, now that you mention it, I do it think `sasl_plain_` is clearer.

larsjsol:
Thank you for you speedy reply! I've responded to your questions inline.

I hope to find the time to make an update tomorrow.

>  I'm not sure whether this would work on an 0.9 cluster
In theory it should work if we skip any metadata requests and the SaslHandShakeRequest, but it should probably be tested :)

kafka/conn.py
larsjsol: I've tried to send a `SaslHandShakeRequest` with a callback to a modified `_try_authenticate_plain()`, but that seemed to cause `client_async._bootstrap()` to send a `MetadataRequest` before we had a chance to process the handshake request. That in turn caused the server to close the connection.

Maybe I'm doing wrong though. I am having some difficulties in groking the poll/callback-system.

larsjsol:
>> here and above you've chosen to do inline calls to recv() and block until the response is available. 

> Maybe I'm doing wrong though. I am having some difficulties in groking the poll/callback-system.
￼ Add a line note

I guess what I am saying is that I will need some pointers in order to implement this using callbacks. 

On the other hand this does work and since authentication only occurs when connecting to a broker, the blocking might not be a noticeable issue as one usually do not do that wery often.

dpkp:
I fixed merge conflicts and callback structure for SaslHandshake request / response.

larsjsol:
> I fixed merge conflicts and callback structure for SaslHandshake request / response.

Cool! I only had to make som tiny changes to get it working

(This PL got a bit messy and contains a bunch of unrelated commits. Sorry about that)

dpkp:
I squashed and pushed to master manually. Thanks for the great PR!

ottomata:
Does this mean tag time? :D

PR #780, opened 2016-07-25
Document the ssl_password config option

ssaamm:
PR #750 added the code for passing along a password, but not any documentation on it.


dpkp:
Thanks!

PR #807, opened 2016-08-18
Workaround gevent bug in python3 connect_ex monkeypatch

asgoel:
For example, the gevent socket implementation returns a `BlockingIOError` instead of an int errno. This is the only place I see it being used, so I figured this is the simplest fix for now. 

Let me know your thoughts.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -270,7 +270,7 @@ class BrokerConnection(object):
                     self.afi = self._sock.family
                     self._gai = None
             except socket.error as err:
-                ret = err
+                ret = self._get_errno(err)
 
             # Connection succeeded
             if not ret or ret == errno.EISCONN:
@@ -317,6 +317,11 @@ class BrokerConnection(object):
                 self.config['state_change_callback'](self)
 
         return self.state
+
+    def _get_errno(self, socket_error):
+        if isinstance(socket_error, BlockingIOError):
+            return socket_error.errno
+        return socket_error
 
     def _wrap_ssl(self):
         assert self.config['security_protocol'] in ('SSL', 'SASL_SSL')


dpkp:
I can't duplicate that behavior with gevent. Can you provide a snippet that reproduces? My quick check in a repl suggests that gevent does return integer error codes like the unpatched socket call:

```
In [1]: import socket
In [2]: from gevent import monkey
In [3]: monkey.patch_socket()
In [4]: s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
In [5]: s.setblocking(False)
In [6]: try:
    ret = s.connect_ex(('www.google.com', 80))
except socket.error as err:
    ret = err
   ...:
In [7]: ret
Out[7]: 36

In [8]: import errno
In [9]: errno.errorcode[36]
Out[9]: 'EINPROGRESS'
```

asgoel:
What version of python and gevent are you running. Here is a version that fails with python 3.5.0, gevent 1.1.2. Technically you should patch everything before you import, but even if I import before patching it will still fail.:

```
from gevent import monkey
monkey.patch_all()

import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.setblocking(False)
try:
    ret = s.connect_ex(("www.google.com", 80))
except socket.error as err:
    ret = err

print(ret)
print(type(ret))
``

Output:
```
[Errno 115] Operation now in progress
<class 'BlockingIOError'>
```

asgoel:
Seems like your snippet only works on python 2

dpkp:
This seems to me like a bug in gevent. Their code attempts to translate the exception to an errno here:
https://github.com/gevent/gevent/blob/b550598753d3cf4763cb47ba3bf804bfb9db26b0/src/gevent/_socket3.py#L305-L314  specifically, L312

I would prefer if gevent were fixed to correctly patch socket.connect_ex on python3!

dpkp:
I filed https://github.com/gevent/gevent/issues/841

asgoel:
Ahh I see. That makes sense. Will wait for that to be resolved and test
again

guojh:
What's the expected type of variable `ret` in `BrokerConnection.connect`?
`int | None` or `int | Exception | None`?

jeffwidman:
Can this be closed since it's been fixed upstream in gevent?

PR #808, opened 2016-08-19
Add select timeout

ecanzonieri:
When the timeout expires we close all pending connections and return FailedPayloadError for all pending requests.

This is a bit of a temporary fix. The upstream SimpleKafkaClient has been totally refactored to support this case in a much better way. We should pull the upstream client, but we need a bit more time to weight the changes.


ecanzonieri:
Sorry I meant to open this pr in my fork. Please, ignore!

Issue #810, opened 2016-08-22
Clients cache broker DNS for life of process


benauthor:
We deploy a kafka cluster on AWS, where hostnames like `broker-1.foo.com`, `broker-2.foo.com`, etc. are mapped to particular AWS instances `i-1234.aws`, `i-5678.aws`. Once in a while, not often but frequently enough, AWS instances go sour and we must replace them.

This goes seamlessly enough on the broker side (after configuring the JVM to cache DNS with a low TTL or not at all): kill instance`i-1234.aws` aka `broker-1.foo.com`, the cluster rebalances without it, spin up a new instance `i-9876.aws`,  configure DNS stuff so `broker-1.foo.com` resolves there, it joins the cluster, and we're back to normal and off to the races.

Unfortunately, on the client side, long-lived worker processes using kafka-python cannot connect to the new broker at the old hostname; they enter a connection error loop. The errors are handled in kafka-python and never bubble up to the level of my application code, so I can't act on this. Once I bounce the worker process, it can connect.

Looks like `kafka.conn.BrokerConnection` does a DNS lookup once on first connection via `getaddrinfo`, after which the IP is stashed for later use, and the BrokerConnection object is stashed in the client's `_conns` dict for the life of the process.

I realize my cluster scenario is only one possible deployment scenario among many, but I imagine other folks might be doing something similar. Would it make more sense for the client to stash just the pre-lookup hostname and do it again on every reconnection, leaving DNS caching up to the host OS? Or at least keep a configurable TTL on it?

dpkp:
This sounds like a bug! Will investigate. PRs welcome if you have thoughts about how to improve handling.

benauthor:
I hope to have time to poke at it later this week...

dpkp:
Also, can you confirm the version of kafka-python you are using?

benauthor:
Oh sure, sorry. Still on 1.2.5.

jeffwidman:
Wasn't this solved by #812 ?

benauthor:
do believe so

PR #812, opened 2016-08-23
When hostname lookup is necessary, do every connect

benauthor:
Proposed solution for https://github.com/dpkp/kafka-python/issues/810

Fixes a bug where lookup was done only once for the whole life of the process -- if a broker's IP changed, client couldn't reconnect.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -157,6 +157,9 @@ class BrokerConnection(object):
         self.hostname = host
         self.port = port
         self.afi = afi
+        self._init_host = host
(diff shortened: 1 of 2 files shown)


benauthor:
@dpkp any thoughts on this? Would love to get this patch in production without going off on a fork.

otoolep:
This is an important fix for me too, any updates? I'll fork if it looks this is not going to be merged.

dpkp:
Merged as cbe8a6a w/ conflict resolved

PR #815, opened 2016-08-24
Fix murmur2 bug handling python2 bytes that do not ascii encode

dpkp:
As reported by @srmerritt


srmerritt:
What happens if the caller passes in unicode for a key, or has a key_serializer that returns unicode?

dpkp:
serializers should always return bytes -- that is what kafka messages contain. If there are serializers relying on the previous behavior, they will need to be changed to return bytes.

PR #824, opened 2016-09-07
Raise exception if given a bad topic name

ssaamm:
I accidentally tried to `.subscribe([None])` this morning, and it took a little while to figure out what I'd done wrong.

Thought this check might be useful in the future. What do you think?

### DIFF
--- a/kafka/consumer/subscription_state.py
+++ b/kafka/consumer/subscription_state.py
@@ -133,6 +133,9 @@ class SubscriptionState(object):
         if self._user_assignment:
             raise IllegalStateError(self._SUBSCRIPTION_EXCEPTION_MESSAGE)
 
+        if isinstance(topics, str):
+            topics = [topics]
+
         if self.subscription == set(topics):
             log.warning("subscription unchanged by change_subscription(%s)",
                         topics)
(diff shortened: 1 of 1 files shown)


ssaamm:
```Exception: Failed to start KafkaInstance before max_timeout```

It looks to me like the build that failed isn't due to this change.

dpkp:
The more error prone call, I think, is `.subscribe('mytopic')` (bare string, no list)

ssaamm:
I can imagine a number of ways to handle that case. My first intuition is to make sure the argument is iterable, but `str`s *are* iterable.

Some ways that stick out to me are:

1. Check specifically for the `topics` argument to be a `str`, log a warning, and proceed with the current behavior (subscribing to a bunch of topics with names of length 1)
2. Check specifically for `str`, put the `topics` argument into a list for the user (`topics = [topics]`), and proceed as normal
3. Raise a `TypeError` if `topics` is not a `list` (disadvantage being that it's conceivable for users to send in a different iterable/collection type)

Option 2 seems like a sane way to handle things to me. Do you agree? Is there a better way to handle the `.subscribe('bare string')` case?

dpkp:
Agree -- I think #2 is the best option. want to update this PR and I'll merge?

ssaamm:
Sure! How does this look?

ssaamm:
@dpkp, any updates on this one?

dpkp:
Looks good -- we should use 'six.string_types' to cover unicode strings in python2 . I will update after merging.

PR #826, opened 2016-09-09
Partitioners update

barrotsteindev:
This branch has updated partitioners which use the __call__ method to partition.
The round robin partitioner is also able to dynamically add partitions to its cycle, so if the topic is altered or brokers are down it will round robin only the available partitions.


barrotsteindev:
Fixing minor bugs

PR #827, opened 2016-09-13
Partitioners fix

barrotsteindev:
The partitioners have been updated to use the new KafkaProducer calls and support the old SimpleProducer APIs.
The partitioners are callable and the RoundRobinPartitioner caches the last partition and supports dynamic partition altering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ docs/_build
 .cache*
 .idea/
 integration-test/
+tests-env/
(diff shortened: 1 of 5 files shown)


dpkp:
I had planned on removing many of these when SimpleProducer is removed. Do you think these partitioners are still useful w/ the new KafkaProducer ?

barrotsteindev:
The changes are made so the old partitioners work with the new KafkaProducer, while still maintaining support for the old producers.
So yes, these partitioners will still be useful, especially the round robin partitioner which has a better logic compared to the old one

dpkp:
Ok, thanks!

PR #860, opened 2016-10-26
Fix of exception raise in case of auto_offset_reset is set to None in KafkaConsumer

sibiryakov:
Without that patch exception wasn't raised, instead consumer keeps retrying fetching.

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -236,7 +236,7 @@ class Fetcher(six.Iterator):
         current_out_of_range_partitions = {}
 
         # filter only the fetchable partitions
-        for partition, offset in self._offset_out_of_range_partitions:
+        for partition, offset in six.iteritems(self._offset_out_of_range_partitions):
             if not self._subscriptions.is_fetchable(partition):
                 log.debug("Ignoring fetched records for %s since it is no"
                           " longer fetchable", partition)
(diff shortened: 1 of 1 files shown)


kafka/consumer/fetcher.py
```diff
@@ -740,12 +740,13 @@ def _handle_fetch_response(self, request, send_time, response):
...
                     fetch_offset = fetch_offsets[tp]
+                    log.info("Fetch offset %s is out of range", fetch_offset)
                     if self._subscriptions.has_default_offset_reset_policy():
                         self._subscriptions.need_offset_reset(tp)
+                        log.info("Resetting offset")
                     else:
                         self._offset_out_of_range_partitions[tp] = fetch_offset
-                    log.info("Fetch offset %s is out of range, resetting offset",
-                             fetch_offset)
+                        log.info("Raising exception")
```
dpkp: There are edge cases where we dont actually raise an exception -- if the partition gets marked unfetchable, or if a seek invalidates the fetch position.

kafka/consumer/fetcher.py
```diff
@@ -740,12 +740,13 @@ def _handle_fetch_response(self, request, send_time, response):
                     self._client.cluster.request_update()
                 elif error_type is Errors.OffsetOutOfRangeError:
                     fetch_offset = fetch_offsets[tp]
+                    log.info("Fetch offset %s is out of range", fetch_offset)
```
dpkp: This log statement should probably include the topic partition

kafka/consumer/fetcher.py
```diff
@@ -740,12 +740,13 @@ def _handle_fetch_response(self, request, send_time, response):
...
                     if self._subscriptions.has_default_offset_reset_policy():
                         self._subscriptions.need_offset_reset(tp)
+                        log.info("Resetting offset")
```
dpkp: Log statement should include the topic-partition

PR #867, opened 2016-11-01
Fix of wrong MetadataRequest in case of empty topics list

sibiryakov:
``topics`` were set to ``[]`` instead of ``None`` because of empty ``self._topics`` set for MetadataRequest_v1


sibiryakov:
@dpkp Hi Dana, please have a look at this. It shouldn't take much time, but it's blocking out https://github.com/scrapinghub/frontera release. Ty!

PR #877, opened 2016-11-12
Typo "chance" --> "change"

jeffwidman:
Thankfully this was only a typo in the docs, not the actual API


jeffwidman:
Squashed into #874

PR #897, opened 2016-12-02
Add docstring for max_records

### LINKED ISSUE #893: RTD docs for max_records don't mention defaults to inheriting from max_poll_records
Source shows that max_records defaults to inheriting from max_poll_records: https://github.com/dpkp/kafka-python/blob/07237d98945f8e1f1161ab5082230d9112016620/kafka/consumer/group.py#L513

Needs to be added to the docstring.

I'll try to submit a PR shortly

jeffwidman:
Fix #893


jeffwidman:
Anything else needed here? Our developers could use this in the docs as I've been asked twice already about whether they need to specify this.

PR #898, opened 2016-12-02
Add sphinx formatting to hyperlink methods

jeffwidman:
Got tired of scrolling, so I attempted to hyperlink the methods in Sphinx. 

I *think* this is the correct way to format things, but not 100% confident.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -91,11 +91,11 @@ class SimpleClient(object):
         Returns the leader for a partition or None if the partition exists
         but has no leader.
 
-        UnknownTopicOrPartitionError will be raised if the topic or partition
-        is not part of the metadata.
-
-        LeaderNotAvailableError is raised if server has metadata, but there is
-        no current leader
+        Raises:
+            UnknownTopicOrPartitionError: If the topic or partition is not part
+                of the metadata.
+            LeaderNotAvailableError: If the server has metadata, but there is no
+        current leader.
         """
 
         key = TopicPartition(topic, partition)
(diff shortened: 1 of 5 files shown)


dpkp:
I mostly use `help(method)` in ipython - I dont think this works with that, does it?

jeffwidman:
Color me ignorant on that one.

I don't think it breaks it, so probably no harm in merging but not positive. I should try using `help(method)` myself more often.

dpkp:
it doesn't break in the sense of a stacktrace or error, but it does show the markup which I think is confusing to non-sphinx users: ":meth:`reinit`" doesn't make a lot of sense literally. I wonder if there is a way to have cake and eat it too here?

dpkp:
When I test locally I need a `.` prefix to get the links to work correctly. For example:
```
:meth:`.subscribe`
```

jeffwidman:
That makes sense. I never built it locally, just copied the style I'd seen on other projects. Before I update the PR though, I want to check if this is something you actually want to merge? Wasn't sure that was the case given your comment on `help(method)`

jeffwidman:
Rebased and updated with local refs using the syntax:
    :meth:`.method`

I wasn't sure how to format refs that were local to the project but not to the file... see the references in `kafka/producer/base.py`. That was the only file that had cross-file references.

jeffwidman:
Rebased.

dpkp:
Sorry I left this one lingering for so long!

jeffwidman:
No worries, thanks for merging. Very convenient to have these hyperlinks in RTD.

PR #900, opened 2016-12-02
Changes:

chillaxd:
Current Scenario:
If cluster is unreachable, instance creation of KafkaProducer does not throw any error or exception. So during message send to Kafka, it will get KafkaTimeout Error (as per documentation).

Problem:
Due to this, user could not know the status of connection before actual send. It will unnecessary waste time and make hard to handle error in some cases. Like if someone tries to implement in AWS Lambda as KafkaProducer.

This changes addresses this problem with keeping the existing flow as it is.
How:
It will introduce a variable in KafkaProducer which will be set False if and only if Kafka cluster failed to reach/connect. So as per the state of this boolean variable user can handle failed secnarios better.

Comment for this commit:
* Boolean variables are introduced in kafka.py and client_async.py
  So that if kafka cluster is not reachable during creation of KafkaProducer instance
  that variable would be False neither True

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -191,6 +191,7 @@ class KafkaClient(object):
         self._selector.register(self._wake_r, selectors.EVENT_READ)
         self._closed = False
         self._sensors = None
+        self.reachable = True
         if self.config['metrics']:
             self._sensors = KafkaClientMetrics(self.config['metrics'],
                                                self.config['metric_group_prefix'],
@@ -198,8 +199,8 @@ class KafkaClient(object):
 
         self._bootstrap(collect_hosts(self.config['bootstrap_servers']))
 
-        # Check Broker Version if not set explicitly
-        if self.config['api_version'] is None:
+        # Check Broker Version if not set explicitly and only when kafka cluster is reachable
+        if self.reachable and self.config['api_version'] is None:
             check_timeout = self.config['api_version_auto_timeout_ms'] / 1000
             self.config['api_version'] = self.check_version(timeout=check_timeout)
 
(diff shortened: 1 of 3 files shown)


williamsjj:
I think this is a dangerous pattern to introduce. It can encourage all sorts of manual failover pathologies...for example the "failed" endpoint A has come back and now the "healthy" endpoint B is actually down, so you're preemptively sending your request to a failed endpoint.  

The best approach is to try the publish and retry it after timeout. If your app has some sort of SLA that would benefit from a centralized view of endpoints being up/down, that really should be in located in your application logic where you can imbue the specific intelligence you need for your use case. I don't think this generalizes well.

chillaxd:
Thanks Jason for those words. But still I have doubt. Like in the example you have depicted, bootstrap servers consists A and B. A is down and B is up. Now according to my understanding when I try to send to B, if B got down then, I will still got Timeout error right? So user can put logic accordingly for that error. What I try to tell this will actually tell you whether any node in the cluster is actually reachable or not during creation of Producer.

Now say, my cluster have 3 nodes. During creation of producer I have found only 1 is up, but I put ack='all'. So after send I will not get the ack back to my end. Still I will get timeout error. 
Now if all nodes are down. Then as per current implementation it will print only error, not even it raises any error.

williamsjj:
From my reading of the ack=all docs, you're going to get a timeout error if all of the nodes (or the minimum corresponding to your ack setting) don't respond to the leader. A timeout error should be treated that it wasn't committed in my understanding.

Since you're going to get a timeout when the server doesn't respond, I don't know what your code adds except pre-emptive failover, which I oppose for the reasons above.

PR #909, opened 2016-12-12
Fix SimpleClient Broker API version default

### LINKED ISSUE #902: BrokerConnection API_VERSION has wrong default value in docstring
I'm confused about the API_VERSION default value.

[`conn.py`](https://github.com/dpkp/kafka-python/blob/master/kafka/conn.py#L77) says:
> 'api_version': (0, 8, 2), # default to most restrictive

But the [docstring](https://github.com/dpkp/kafka-python/blob/master/kafka/conn.py#L139) for that class says:

>  api_version (tuple): specify which kafka API version to use. Accepted
                values are: (0, 8, 0), (0, 8, 1), (0, 8, 2), (0, 9), (0, 10)
                If None, KafkaClient will attempt to infer the broker
                version by probing various APIs. *Default: None*

So the default isn't `None`, it's `(0, 8, 2)`. Which should be fixed, the docstring or the default value?

I'm *think* this class is only used by SimpleClient not by the newer KafkaClient async, since I see another set of values here:
https://github.com/dpkp/kafka-python/blob/master/kafka/client_async.py#L78 where both the code and the docstring state the default value is `None`.

Help?

jeffwidman:
Fix #902

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -114,10 +114,9 @@ class BrokerConnection(object):
             providing a file, only the leaf certificate will be checked against
             this CRL. The CRL can only be checked with Python 3.4+ or 2.7.9+.
             default: None.
-        api_version (tuple): specify which Kafka API version to use. Accepted
-            values are: (0, 8, 0), (0, 8, 1), (0, 8, 2), (0, 9), (0, 10)
-            If None, KafkaClient will attempt to infer the broker
-            version by probing various APIs. Default: None
+        api_version (tuple): Specify which Kafka API version to use.
+            Accepted values are: (0, 8, 0), (0, 8, 1), (0, 8, 2), (0, 9),
+            (0, 10). Default: (0, 8, 2)
         api_version_auto_timeout_ms (int): number of milliseconds to throw a
             timeout exception from the constructor when checking the broker
             api version. Only applies if api_version is None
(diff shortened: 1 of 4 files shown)


kafka/conn.py
```diff
@@ -135,8 +135,8 @@ def __init__(self, host, port, afi, **configs):
                 default: None.
             api_version (tuple): specify which kafka API version to use. Accepted
                 values are: (0, 8, 0), (0, 8, 1), (0, 8, 2), (0, 9), (0, 10)
-                If None, KafkaClient will attempt to infer the broker
-                version by probing various APIs. Default: None
+                If None, SimpleClient will attempt to infer the broker
```
dpkp: Both the new KafkaClient and the old SimpleClient use BrokerConnection, so we should probably describe just the BrokerConnection behavior here

jeffwidman:
Updated, also cleaned up docstrings while I was in there to be more consistent across the various clients.

jeffwidman:
I just rebased to fix merge conflicts...

jeffwidman:
Rebased to fix merge conflicts.

jeffwidman:
Rebased

jeffwidman:
Anything need changing on this one?

dpkp:
Thanks!

jeffwidman:
Thanks for merging this one

PR #914, opened 2016-12-18
Fix #913: If key is null, skip deserializing

### LINKED ISSUE #913: KafkaConsumer always passes message key to deserializer, even if key is None
I specified `json.loads` as the `KafkaConsumer`'s `key_deserializer` callable, then produced a message without a key.

`json.loads` throws an exception about expecting a string. Looks like root cause is here: https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/fetcher.py#L546

Fetcher is always using passing the key to the deserializer, even if the key is `None`.

Will submit a PR shortly.

jeffwidman:
Fix #913 

I didn't test the JavaConsumer, but I suspect it skips null keys as it seems the most logical behavior. 

I thought about adding this logic the message value processing as well, but I can't think of a scenario where you'd send a null message, versus it is realistic that you'd send a message with a null key.

Didn't have time to add tests yet.

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -542,12 +542,12 @@ class Fetcher(six.Iterator):
             raise
 
     def _deserialize(self, msg):
-        if self.config['key_deserializer']:
-            key = self.config['key_deserializer'](msg.key) # pylint: disable-msg=not-callable
+        if self.config['key_deserializer'] and msg.key is not None:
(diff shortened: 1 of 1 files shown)


dpkp:
I'd rather keep the interface simple: deserializers are functions that receive bytes or None. I do not want to prevent users from specifying deserializers that transform None -> object, which would be the effect if we force skipping deserializer on None.

jeffwidman:
Makes sense.

PR #919, opened 2016-12-21
Update pytest fixtures to new yield syntax

jeffwidman:
http://doc.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -18,10 +18,8 @@ def version():
 def zookeeper(version, request):
     assert version
     zk = ZookeeperFixture.instance()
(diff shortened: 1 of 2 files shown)


dpkp:
Looks great. There's also one more of these in `test/test_metrics.py`

dpkp:
Looks good. Should we add a version pin to pytest dep in tox.ini ?

jeffwidman:
Up to you, I don't feel strongly either way. IIRC, this syntax requires pytest >= 3.x.

PR #924, opened 2016-12-30
Make SSL warning list the correct Python versions

jeffwidman:
Was a bit surprised to see this warning being thrown even though we're already on 2.7

According to the Python docs, the new SSL error classes were added in 2.7.9 and 3.3, so I updated the warning:
https://docs.python.org/3/library/ssl.html?highlight=ssl#ssl.SSLZeroReturnError
https://docs.python.org/2/library/ssl.html#ssl.SSLZeroReturnError

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -36,9 +36,9 @@ try:
     ssl.SSLWantWriteError
     ssl.SSLZeroReturnError
 except:
-    log.warning('old ssl module detected.'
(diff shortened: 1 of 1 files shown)


jeffwidman:
Also, I'd prefer this warning only be happen if we're actually using SSL. I don't see much point in throwing it for `PLAINTEXT` connections. But I think that should be a separate PR.

PR #949, opened 2017-01-18
use select.select without instance bounding

adamwen829:
when we use invoking self._select, python will pass self as select.select's first argument and we will get `TypeError: select() takes at most 4 arguments (5 given)`

### DIFF
--- a/kafka/vendor/selectors34.py
+++ b/kafka/vendor/selectors34.py
@@ -331,7 +331,7 @@ class SelectSelector(_BaseSelectorImpl):
             r, w, x = select.select(r, w, w, timeout)
             return r, w + x, []
     else:
-        _select = select.select
+        _select = staticmethod(select.select)
 
     def select(self, timeout=None):
         timeout = None if timeout is None else max(timeout, 0)


dpkp:
Not sure what to do with this one -- we vendor `selectors34` and this patch has not been accepted upstream. See https://github.com/berkerpeksag/selectors34/pull/4

berkerpeksag:
Note that the patch was rejected because of Gevent's use case. I'm happy to merge berkerpeksag/selectors34#4 and reopen the upstream issue if you could share a reproducer. The patch looks good to me, but I don't get any `TypeError`s with the following snippet:

```py
>>> import selectors34 as s
>>> ss = s.SelectSelector()
>>> ss.select(0)
[]
>>> ss._select([], [], [], 0)
([], [], [])
```

berkerpeksag:
Just an FYI: This is now fixed in selectors34 master by https://github.com/berkerpeksag/selectors34/commit/6128e067bcd0cc7e280d673c6eb14c3c754863d8.

PR #951, opened 2017-01-18
Add support for offsetRequestV1 messages

jlafaye:
Those can be used to request the offset of a stream at a given timepoint

### DIFF
--- a/kafka/protocol/offset.py
+++ b/kafka/protocol/offset.py
@@ -46,7 +46,7 @@ class OffsetRequest_v0(Struct):
             ('topic', String('utf-8')),
             ('partitions', Array(
                 ('partition', Int32),
-                ('time', Int64),
+                ('timestamp', Int64),
                 ('max_offsets', Int32)))))
     )
     DEFAULTS = {
--- a/kafka/protocol/offset.py
+++ b/kafka/protocol/offset.py
@@ -22,6 +22,19 @@ class OffsetResponse_v0(Struct):
                 ('offsets', Array(Int64))))))
     )
 
+class OffsetResponse_v1(Struct):
+    API_KEY = 2
+    API_VERSION = 1
+    SCHEMA = Schema(
+        ('topics', Array(
+            ('topic', String('utf-8')),
+            ('partitions', Array(
+                ('partition', Int32),
+                ('error_code', Int16),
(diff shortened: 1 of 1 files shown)


kafka/protocol/offset.py
```diff
@@ -40,6 +53,24 @@ class OffsetRequest_v0(Struct):
...
+    )
+    DEFAULTS = {
+        'replica_id': -1
+    }
+
 
 OffsetRequest = [OffsetRequest_v0]
 OffsetResponse = [OffsetResponse_v0]
+OffsetRequest = [OffsetRequest_v0, OffsetRequest_v1]
+OffsetResponse = [OffsetResponse_v0, OffsetResponse_v1]
```
dpkp: Should only assign the variables once here -- you can delete the earlier lists.

jlafaye:
Hello,

thank you for taking the time to review the pull request. This is my first time so do not hesitate to show me what's wrong so I do not keep doing the wrong things.
I fixed the obvious error in first commit  (a026051).
I noticed that Travis was complaining for one specific version. I could not find anything wrong in the logs. I guess it is a random runtime issue. Should I investigate further ?

kafka/protocol/offset.py
```diff
@@ -40,6 +53,22 @@ class OffsetRequest_v0(Struct):
...
+    API_KEY = 2
+    API_VERSION = 1
+    RESPONSE_TYPE = OffsetResponse_v1
+    SCHEMA = Schema(
+        ('replica_id', Int32),
+        ('topics', Array(
+            ('topic', String('utf-8')),
+            ('partitions', Array(
+                ('partition', Int32),
+                ('time', Int64)))))
```
dpkp: can we call this 'timestamp' to be consistent with the protocol docs ?

dpkp:
Travis test suite is not 100% reliable. In most cases it seems like our fixtures (kafka broker or zookeeper) crash or become unresponsive. You do not need to investigate these errors; I'll typically just restart the job.

dpkp:
thanks!

kafka/protocol/offset.py
jeffwidman: See also https://github.com/dpkp/kafka-python/pull/969

Issue #939, opened 2017-01-11
disable independency with SSL


minagawa-sho:
Kafka-python seems to need the SSL implement on Python interepreter where it be running.
But I think that it should be comfortable to be able to disable its dependency because ssl module is optional (be able to disable on building) for Python.

Please tell me a known workaround for this topic, if exists.
Or is there this feature on the future plan of the product ?

dpkp:
Can you provide an example of the issue you're seeing specifically?

minagawa-sho:
This error occurs on my enviroment.
With the Python interpreter built on the machine where openssl is not installed, ssl module importing is unsuccessful.

```
$ ./python 
Python 2.7.12 (default, Jan 13 2017, 15:26:28) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
\>\>\> import kafka
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/var/service/mx/krb/sqr/pkg.v0.1.0/python/lib/python2.7/site-packages/kafka_python-1.3.2-py2.7.egg/kafka/__init__.py", line 21, in <module>
    from kafka.consumer import KafkaConsumer
  File "/var/service/mx/krb/sqr/pkg.v0.1.0/python/lib/python2.7/site-packages/kafka_python-1.3.2-py2.7.egg/kafka/consumer/__init__.py", line 5, in <module>
    from .group import KafkaConsumer
  File "/var/service/mx/krb/sqr/pkg.v0.1.0/python/lib/python2.7/site-packages/kafka_python-1.3.2-py2.7.egg/kafka/consumer/group.py", line 11, in <module>
    from kafka.client_async import KafkaClient, selectors
  File "/var/service/mx/krb/sqr/pkg.v0.1.0/python/lib/python2.7/site-packages/kafka_python-1.3.2-py2.7.egg/kafka/client_async.py", line 24, in <module>
    from .conn import BrokerConnection, ConnectionStates, collect_hosts, get_ip_port_afi
  File "/var/service/mx/krb/sqr/pkg.v0.1.0/python/lib/python2.7/site-packages/kafka_python-1.3.2-py2.7.egg/kafka/conn.py", line 10, in <module>
    import ssl
  File "/var/tmp/minagawa/kurobe-sqr/build/work/Python-2.7.12/Lib/ssl.py", line 97, in <module>
[...]

PR #954, opened 2017-01-22
Add support for Python built without ssl (#939)

minagawa-sho:
I wrapped ssl module importing and add the recovery path from ImportError.
* Unlike my initial assumption, I understood that ssl library is depended by many standard library modules (e.g hashlib), so this workaround may not be effective on practice. :( 
* Such workaround is implemented many third-party modules (e.g redis-py), so regardless of the effectiveness I think this workaround is necessarily to retain the library's versatility.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -7,7 +7,6 @@ import logging
 import io
 from random import shuffle
 import socket
-import ssl
 import time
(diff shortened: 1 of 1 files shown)


dpkp:
Thanks!!

Issue #957, opened 2017-01-24
CPU usage high after upgrading to 1.3.1


jianbin-wei:
We upgraded from 0.9.5 to 1.3.1.  Our producer code still uses SimpleClient interface before migrating to new KafkaClient.  We noticed that the cpu usage doubled after the upgrade.

After profiling, it turns out that `SimpleClient._send_broker_aware_request` calls BrokerConnection.recv` continuously without blocking.  We have the call ratio as 1:1500 on average.  It is similar for `SimpleClient._send_broker_unaware_request`.  The old KafkaConnection object uses blocking socket while the new one uses non-blocking one along with selector.  Thus the issue between the old SimpleClient and new non-blocking BrokerConnection.

dpkp:
PR merged to master!

PR #958, opened 2017-01-24
Use select to poll sockets for read to reduce CPU usage

jianbin-wei:
@dpkp, this pull request is for people who may have the same issue as us #957.  

As SimpleClient is to be deprecated, I used BrokerConnection._sock directly instead of updating BrokerConnection to expose _sock.  Also `select` module is used directly.  You can update it if you think it is needed.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -284,7 +284,7 @@ class SimpleClient(object):
             # block until a socket is ready to be read
             sockets = [
                 conn._sock
-                for future, (conn, _) in connections_by_future.iteritems()
+                for future, (conn, _) in six.iteritems(connections_by_future)
                 if not future.is_done and conn._sock is not None]
             if sockets:
                 read_socks, _, _ = select.select(sockets, [], [])
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -6,6 +6,7 @@ import functools
 import logging
 import random
 import time
+import select
 
 from kafka.vendor import six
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
@dpkp Jianbin and I are working together on our internal `kafka-python` upgrade. I know `SimpleClient` is deprecated, but it'd still be nice to have this patched in some way in `kafka-python`. There's a lot of code out in the wild that still uses `SimpleClient` and a doubling of CPU usage is a non-trivial performance regression.

dpkp:
Thanks - I'll take a look. Seems like a good improvement.

jianbin-wei:
I just close and re-open the pull request to re-try the build.  The failure is due to not availability of https://archive.apache.org/dist/kafka/0.10.1.1/kafka_2.10-0.10.1.1.tgz

jeffwidman:
Typically to trigger rebuilds you have to push a new commit. Just force-push to the existing branch to overwrite existing commits, for example if you squashed your two current commits into one and then force-pushed, it would re-run the tests. Or a project collaborator/member/owner can manually re-trigger the test.

dpkp:
We use the `selectors` module in the main async client. Would it be better to use that over raw select?

jianbin-wei:
> Also select module is used directly. You can update it if you think it is needed.

Yes I agree that `selectors` would be better.  I would update it when I get some time.

dpkp:
I'm ok to merge as-is if you want to follow-up with a selectors patch later. I don't work much on the old SimpleClient so I'm unlikely to get to it myself.

jeffwidman:
When switching to `selectors`, we should verify it plays nice with `gevent` code. 

We're hitting a bug internally when we migrated a service that uses `gevent` to the new `KafkaConsumer`. The traceback shows select.epoll was used, but then disappeared (probably related to gevent monkey-patching)... If I hardcode the selector used by `KafkaConsumer` to `PollSelector` the issue disappears. While that's a separate issue that I'm still debugging, in the meantime we need to verify that switching this patch to use `selectors` doesn't break `SimpleConsumer`.

PR #960, opened 2017-01-27
Fix raise exception from SubscriptionState.assign_from_subscribed

Artimi:
Raising `ValueError` from `SubscriptionState.assign_from_subscribed()`
always fails because there is used percent formatting of `TopicPartition`.
But `TopicPartition` is [namedtuple](https://github.com/dpkp/kafka-python/blob/master/kafka/structs.py#L59) with two elements and percent
formatting tries to unpack tuples, fails and raise exception:
`TypeError: not all arguments converted during string formatting`.
Converting to string first solves the issue.


dpkp:
Nice catch!

Issue #870, opened 2016-11-06
Use Kafka Fetch Request Version 3


Drizzt1991:
Hi, as I can see there's still no support for Version 3 of Fetch requests (with an max_bytes for all partitions, see https://kafka.apache.org/protocol).

Drizzt1991:
It was introduced in Kafka 0.10.1.0, is anyone currently working on adding this? If not I think, I can add it myself. It shouldn't be too hard, with the exception, that topics will now be order-sensitive.

dpkp:
I have not started working on this. PR welcome!

jeffwidman:
Added the struct itself in #943 because I had a little extra time... not sure I'll be able to finish this, anyone else feel free to continue...

Still to do before closing this issue:
- [ ]  add `(0, 10, 1)` to the list of supported `api_version`s
- [ ]  verify the logic of `api_version` checking still works... not sure how it handles a three part tuple vs two part.
- [ ]  do we need to shuffle the topics on request? @Drizzt1991 mentioned this, but I'm not sure where/how to implement.
- [ ]  actually pass in a value for the new `max_bytes` field

Drizzt1991:
Hmm, as I opened the ticket it will take it. I do have time this evening. Thanks for the start.

Drizzt1991:
Emm, it was closed and released in 1.3.3 :) closing

PR #962, opened 2017-01-30
Added `max_bytes` option and FetchRequest_v3 usage.

Drizzt1991:
Also added a check for api_version=(0, 10, 1) and MetadataRequest_v2
[KIP-74](https://cwiki.apache.org/confluence/display/KAFKA/KIP-74%3A+Add+Fetch+Response+Size+Limit+in+Bytes)
NOTE: Depends on PR #974

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -790,7 +790,8 @@ class BrokerConnection(object):
         log.addFilter(log_filter)
 
         test_cases = [
-            ((0, 10), ApiVersionRequest[0]()),
+            ((0, 10, 1), MetadataRequest[2]([])),
+            ((0, 10, 0), ApiVersionRequest[0]()),
             ((0, 9), ListGroupsRequest[0]()),
             ((0, 8, 2), GroupCoordinatorRequest[0]('kafka-python-default-group')),
             ((0, 8, 1), OffsetFetchRequest[0]('kafka-python-default-group', [])),
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -761,10 +761,10 @@ class BrokerConnection(object):
 
         error_type = Errors.for_code(response.error_code)
         assert error_type is Errors.NoError, "API version check failed"
-        max_versions = {
-            api_key: max_version
+        max_versions = dict([
+            (api_key, max_version)
             for api_key, _, max_version in response.api_versions
-        }
+        ])
         # Get the best match of test cases
         for broker_version, struct in test_cases:
             if max_versions.get(struct.API_KEY, -1) >= struct.API_VERSION:
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -18,6 +18,7 @@ from kafka.metrics.stats import Avg, Count, Max, Rate
 from kafka.protocol.api import RequestHeader
 from kafka.protocol.admin import SaslHandShakeRequest
 from kafka.protocol.commit import GroupCoordinatorResponse
+from kafka.protocol.metadata import MetadataRequest
 from kafka.protocol.types import Int32
 from kafka.version import __version__
 
@@ -752,6 +753,24 @@ class BrokerConnection(object):
         self._correlation_id = (self._correlation_id + 1) % 2**31
         return self._correlation_id
 
+    def _check_version_above_0_10(self, response):
+        test_cases = [
+            # format (<broker verion>, <needed struct>)
+            ((0, 10, 1), MetadataRequest[2])
+        ]
+
+        error_type = Errors.for_code(response.error_code)
+        assert error_type is Errors.NoError, "API version check failed"
+        max_versions = {
+            api_key: max_version
+            for api_key, _, max_version in response.api_versions
+        }
+        # Get the best match of test cases
+        for broker_version, struct in test_cases:
+            if max_versions.get(struct.API_KEY, -1) >= struct.API_VERSION:
+                return broker_version
+        return (0, 10, 0)
+
     def check_version(self, timeout=2, strict=False):
         """Attempt to guess the broker version.
 
@@ -776,7 +795,6 @@ class BrokerConnection(object):
         # socket.error (32, 54, or 104)
         from .protocol.admin import ApiVersionRequest, ListGroupsRequest
         from .protocol.commit import OffsetFetchRequest, GroupCoordinatorRequest
-        from .protocol.metadata import MetadataRequest
 
         # Socket errors are logged as exceptions and can alarm users. Mute them
         from logging import Filter
@@ -790,8 +808,8 @@ class BrokerConnection(object):
         log.addFilter(log_filter)
 
         test_cases = [
-            ((0, 10, 1), MetadataRequest[2]([])),
-            ((0, 10, 0), ApiVersionRequest[0]()),
+            # All cases starting from 0.10 will be based on ApiVersionResponse
+            ((0, 10), ApiVersionRequest[0]()),
             ((0, 9), ListGroupsRequest[0]()),
             ((0, 8, 2), GroupCoordinatorRequest[0]('kafka-python-default-group')),
             ((0, 8, 1), OffsetFetchRequest[0]('kafka-python-default-group', [])),
@@ -831,6 +849,10 @@ class BrokerConnection(object):
                 self._sock.setblocking(False)
 
             if f.succeeded():
+                if version == (0, 10):
+                    # Starting from 0.10 kafka broker we determine version
+                    # by looking at ApiVersionResponse
+                    version = self._check_version_above_0_10(f.value)
                 log.info('Broker version identifed as %s', '.'.join(map(str, version)))
                 log.info('Set configuration api_version=%s to skip auto'
                          ' check_version requests on startup', version)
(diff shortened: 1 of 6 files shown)


Drizzt1991:
Took me long enough =) Fixes #870

kafka/consumer/fetcher.py
```diff
@@ -664,19 +665,34 @@ def _create_fetch_requests(self):
...
+                    -1,  # replica_id
+                    self.config['fetch_max_wait_ms'],
+                    self.config['fetch_min_bytes'],
+                    partition_data.items())
+            else:
+                # As of version == 3 partitions will be returned in order as
+                # they are requested, so to avoid starvation with
+                # `fetch_max_bytes` option we need this shuffle
+                partition_data = list(partition_data.items())
+                random.shuffle(partition_data)
```
Drizzt1991: A note to this shuffle. As python dicts and sets are not ordered, currently it will work properly without this shuffle (it will basically be shuffled by _create_fetch_requests, as it uses sets and dicts), but I still prefere it here, as:

* Python3.6 has ordered dicts, which can make the distribution way less equal
* While dicts are not retain insert order, they preserve key order by hashes, which can (probably) result in unequal consumption.

Please correct me if I'm wrong here. My tests run OK even without this shuffle, so I'm a bit concerned.

kafka/consumer/group.py
```diff
@@ -63,6 +63,13 @@ class KafkaConsumer(six.Iterator):
             the server will block before answering the fetch request if
             there isn't sufficient data to immediately satisfy the
             requirement given by fetch_min_bytes. Default: 500.
+        fetch_max_bytes (int): The maximum amount of data the server should
+            return for a fetch request. This is not an absolute maximum, if the
+            first message in the first non-empty partition of the fetch is
+            larger than this value, the message will still be returned to
+            ensure that the consumer can make progress. NOTE: consumer performs
+            multiple fetches in parallel so memory usage will be higher.
```
jeffwidman: This note is a great idea, but tweak the wording slightly? It's unclear 1) what it will be higher than , and 2) how much higher... are we talking 20%, 2x, 5x, etc?

kafka/protocol/metadata.py
```diff
@@ -47,6 +47,30 @@ class MetadataResponse_v1(Struct):
     )
 
 
+class MetadataResponse_v2(Struct):
```
jeffwidman: It's great that you added this, but maybe it should be in a dedicated commit? It just feels logically separate, and I'm unclear why it's needed for the `Fetch_Request/Response` stuff. Personally prefer a commit that says "Here's this new struct, here's what it changes compared to the old one" just because often spelunking in code I'll lookup the commit history to understand changes like this.

kafka/consumer/fetcher.py
jeffwidman: Good call on 3.6 ordered dicts--makes sense. 

However, if v3 supports returning partitions in the order requested, would it be a good idea to expose this as an optional "feature"? Obviously default to random, but give the option to fetch partitions in the same order they are requested. 

For example if a user eventually wanted to request partitions in a particular order, knowingly risking starvation, but wanting to use it as a poor-man's prioritization queue? 

Just tossing it out there... it might be a bad idea, and also might require a lot of internal changes in `kafka-python`, I haven't spent enough time thinking about it.

kafka/consumer/fetcher.py
```diff
@@ -664,19 +665,34 @@ def _create_fetch_requests(self):
...
+            if version < 3:
+                requests[node_id] = FetchRequest[version](
+                    -1,  # replica_id
+                    self.config['fetch_max_wait_ms'],
+                    self.config['fetch_min_bytes'],
+                    partition_data.items())
+            else:
+                # As of version == 3 partitions will be returned in order as
+                # they are requested, so to avoid starvation with
+                # `fetch_max_bytes` option we need this shuffle
```
jeffwidman: Maybe also add to the comment something about Python 3.6 consistent ordering of dicts...

> ...this shuffle. Otherwise, in Python >=  3.6 dicts have consistent order, so can't guarantee hashing to dict/set will randomize order of partitions.

kafka/consumer/fetcher.py
jeffwidman: Weird, Github swallowed my earlier comment. 

+1 for explicitly shuffling to avoid py3.6 issues. 

However, since the protocol explicitly responds in the order requested, maybe this should be exposed to the user? Default to shuffling the order to avoid starvation, but allow the user to override that to always request a particular order... I can see someone using this as a sort of poor-man's priority-queue.  Not sure if this is a good idea, or if it'd require reworking a lot of of `kafka-python` internals... just a thought.

Drizzt1991: I did have the same idea, but could not find any applications to this. And it's probably not this PR's responsibility, as it's rather a feature.

kafka/protocol/metadata.py
jeffwidman: Broken out in #974.

kafka/consumer/group.py
```diff
@@ -63,6 +63,14 @@ class KafkaConsumer(six.Iterator):
             the server will block before answering the fetch request if
             there isn't sufficient data to immediately satisfy the
             requirement given by fetch_min_bytes. Default: 500.
+        fetch_max_bytes (int): The maximum amount of data the server should
+            return for a fetch request. This is not an absolute maximum, if the
+            first message in the first non-empty partition of the fetch is
+            larger than this value, the message will still be returned to
+            ensure that the consumer can make progress. NOTE: consumer performs
+            fetches to multiple nodes in parallel so memory usage will depend
+            on the number of nodes containing partitions for the topic.
```
jeffwidman: Can we call them 'brokers' rather than 'nodes'?

jeffwidman:
LGTM with the caveat that I didn't have time to try the code, just read through it.

kafka/conn.py
```diff
@@ -778,7 +778,8 @@ def filter(self, record):
         log.addFilter(log_filter)
 
         test_cases = [
-            ((0, 10), ApiVersionRequest[0]()),
+            ((0, 10, 1), MetadataRequest[2]([])),
```
dpkp: note to self: for any broker after 0.10 we should just use the results of ApiVersions api request to set api version support.

kafka/protocol/metadata.py
```diff
@@ -47,6 +47,30 @@ class MetadataResponse_v1(Struct):
...
+    API_KEY = 3
+    API_VERSION = 1
```
dpkp: should be API_VERSION = 2 ?

kafka/protocol/metadata.py
dpkp: agree -- would prefer to separate this into separate PR.

Drizzt1991: It's separated, see #974

kafka/protocol/metadata.py
Drizzt1991: Ou, yea =)

Drizzt1991:
As for the failing tests, I do suspect that something is strange, as they fail quite strangely on py27 and py26 on KAFKA<0.10.1, but the fail message has no sense:
```
ERROR: InvocationError: '/home/travis/build/dpkp/kafka-python/.tox/py27/bin/py.test --pylint --pylint-rcfile=pylint.rc --pylint-error-types=EF --cov=kafka --cov-config=.covrc'
```

And the tests run OK on my machine, I have no clue.

kafka/conn.py
Drizzt1991: Should I implement it?

dpkp: ah yes. forgot about this little wart of kafka brokers. so it turns out that kafka < 0.10 will silently ignore incorrect metadata api version requests (or at least not close the socket as expected). The socket behavior is different between py2 and py3, causing KafkaConnection._recv() to block indefinitely on py2.

So we can't use MetadataRequest[2] to check version and will need to implement the smarter ApiVersion approach to avoid breaking auto version checks on older brokers.

dpkp: test via `KAFKA_VERSION=0.8.0 tox -e py27 -- test/test_consumer_group.py::test_paused` -- this hangs indefinitely on my laptop

Drizzt1991: ok, will try it.

Drizzt1991:
Changed check_version to use ApiVersionResponse result. Works quite good for me.

Drizzt1991:
@dpkp Is there anything else for this PR, that needs to be addressed?

kafka/conn.py
```diff
@@ -830,6 +849,10 @@ def connect():
                 self._sock.setblocking(False)
 
             if f.succeeded():
+                if version == (0, 10):
```
dpkp: better would be `if isinstance(request, ApiVersionRequest[0]):`

kafka/conn.py
```diff
@@ -752,6 +753,24 @@ def _next_correlation_id(self):
         self._correlation_id = (self._correlation_id + 1) % 2**31
         return self._correlation_id
 
+    def _check_version_above_0_10(self, response):
+        test_cases = [
+            # format (<broker verion>, <needed struct>)
```
dpkp: might note that the order here matters, and/or make sure we reverse sort it before checking for best match below

dpkp:
Looks great. I added two very minor points that we can fix later (or you can address now if you have time). I'll merge this before next release.

Drizzt1991:
Thanks for the cleanup, looks good =) I missed it on the weekend

PR #966, opened 2017-01-30
DRY-up the MetadataRequest_v1 struct

jeffwidman:


kafka/protocol/metadata.py
```diff
@@ -52,17 +52,15 @@ class MetadataRequest_v0(Struct):
     API_VERSION = 0
     RESPONSE_TYPE = MetadataResponse_v0
     SCHEMA = Schema(
-        ('topics', Array(String('utf-8'))) # Empty Array (len 0) for all topics
+        ('topics', Array(String('utf-8')))  # Empty Array (len 0) for all topics
```
jeffwidman: Should this have additional class vars?

    ALL_TOPICS = None  # Empty Array (len 0) for all topics

kafka/protocol/metadata.py
```diff
@@ -52,17 +52,15 @@ class MetadataRequest_v0(Struct):
...
 
 
 class MetadataRequest_v1(Struct):
     API_KEY = 3
     API_VERSION = 1
     RESPONSE_TYPE = MetadataResponse_v1
-    SCHEMA = Schema(
-        ('topics', Array(String('utf-8'))) # Null Array (len -1) for all topics
-    )
+    SCHEMA = MetadataRequest_v0.SCHEMA  # Null Array (len -1) for all topics
```
jeffwidman: Should this have additional class vars?

    ALL_TOPICS = -1  # Null Array (len -1) for all topics
    NO_TOPICS = None

kafka/protocol/metadata.py
```diff
@@ -52,17 +52,15 @@ class MetadataRequest_v0(Struct):
     API_VERSION = 0
     RESPONSE_TYPE = MetadataResponse_v0
     SCHEMA = Schema(
-        ('topics', Array(String('utf-8'))) # Empty Array (len 0) for all topics
+        ('topics', Array(String('utf-8')))  # Empty Array (len 0) returns all topics
```
jeffwidman: Should this have additional class vars?

    ALL_TOPICS = None  # Empty Array (len 0) returns all topics

Not sure what correct values are for empty arrays in this structs class

kafka/protocol/metadata.py
```diff
@@ -52,17 +52,15 @@ class MetadataRequest_v0(Struct):
...
 
 
 class MetadataRequest_v1(Struct):
     API_KEY = 3
     API_VERSION = 1
     RESPONSE_TYPE = MetadataResponse_v1
-    SCHEMA = Schema(
-        ('topics', Array(String('utf-8'))) # Null Array (len -1) for all topics
-    )
+    SCHEMA = MetadataRequest_v0.SCHEMA  # Null Array (len -1) returns all topics
```
jeffwidman: Should this have additional class vars?

    ALL_TOPICS = -1  # Null Array (len -1) returns all topics
    NO_TOPICS = None  # Empty array (len 0) returns no topics

Not sure what correct values are for empty arrays in this structs class

dpkp: Sure, we could add something like that to make it easier to remember which is which

jeffwidman: Pushed an updated version that includes these class vars. I'm still not sure what correct values are for empty arrays in this structs class, so I set them to `None`, but perhaps they should be `0`.

PR #969, opened 2017-01-31
time --> timestamp to match Java API

jeffwidman:
Kafka server protocol calls this timestamp, not time. 

See also  https://github.com/dpkp/kafka-python/pull/951#discussion_r96803189

Thankfully this struct isn't yet used internally in the code, so easy to correct it now.


dpkp:
Thanks!

Issue #950, opened 2017-01-18
KIP-88 / KAFKA-3853 Report offsets stored in Kafka for groups without an active consumer


jeffwidman:
Adding support for this would make writing external monitoring scripts much easier.

https://cwiki.apache.org/confluence/display/KAFKA/KIP-88%3A+OffsetFetch+Protocol+Update

https://issues.apache.org/jira/browse/KAFKA-3853

Structs for this were added in #971.

Related: #633

PR #971, opened 2017-01-31
KIP-88 / KAFKA-3853: OffsetFetch v2 structs

jeffwidman:
Adds the new structs that are used for #950


dpkp:
we might consider renaming "consumer_group" to "group_id" to be consistent with the published protocol spec.

PR #979, opened 2017-02-07
Fixed couple of "leaks" when gc is disabled

Mephius:
Since I'm using kafka-python in an environment with disabled gc these
"leaks" hit me hard.

### DIFF
--- a/kafka/vendor/selectors34.py
+++ b/kafka/vendor/selectors34.py
@@ -93,7 +93,8 @@ class _SelectorMapping(Mapping):
         return iter(self._selector._fd_to_key)
 
 
-class BaseSelector(six.with_metaclass(ABCMeta)):
+@six.add_metaclass(ABCMeta)
+class BaseSelector(object):
     """Selector abstract base class.
 
     A selector supports registering file objects to be monitored for specific
--- a/kafka/vendor/six.py
+++ b/kafka/vendor/six.py
@@ -70,7 +70,9 @@ else:
         else:
             # 64-bit
             MAXSIZE = int((1 << 63) - 1)
-        del X
+
+        # Don't del it here, cause with gc disabled this "leaks" to garbage
+        # del X
 
 
 def _add_doc(func, doc):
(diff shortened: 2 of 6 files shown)


kafka/vendor/six.py
```diff
@@ -70,7 +70,9 @@ def __len__(self):
         else:
             # 64-bit
             MAXSIZE = int((1 << 63) - 1)
-        del X
+
+        # Don't del it here, cause with gc disabled this "leaks" to garbage
+        # del X
```
dpkp: this one is a vendored library, and I really would prefer fixing upstream. otherwise I'm likely to forget and overwrite when I import a newer version. Have you checked if this is fixed already in six?

kafka/vendor/selectors34.py
```diff
@@ -93,7 +93,8 @@ def __iter__(self):
         return iter(self._selector._fd_to_key)
 
 
-class BaseSelector(six.with_metaclass(ABCMeta)):
+@six.add_metaclass(ABCMeta)
```
dpkp: same re: vendored library. Can you check with upstream version?

kafka/vendor/six.py
Mephius: Upstream doesn't have this fix, but I've filed a pull request: https://github.com/benjaminp/six/pull/176

kafka/vendor/selectors34.py
Mephius: Pull request to vendored library: https://github.com/berkerpeksag/selectors34/pull/7

Mephius: Upstream merged pull request.

Mephius:
Sorry for spamming PR with commits.
The trick with this implementation of WeakMethod is that one can't call it via call chaining:`Message().encode()` cause in this case it looses ref to target object (because of ref_count==0 I believe). So I had to unwind generators and lists comprehensions in order to have ref_count=1 for Message object until it goes out of scope.

Python is magic sometimes.

dpkp:
I merged the selectors34 code from upstream separately. I am ok merging the rest, though I would prefer to wait for upstream to merge before changing six. That said, given that the change is mostly trivial I would be ok merging.

Can you rebase?

dpkp:
Merged manually as `5a0e9715f45b62cfe43e6873b8828f49ab73f710`

Mephius:
Thank you!
P.S. wasn't able to react because of vacation.

Issue #739, opened 2016-06-24
Help: Error sending GroupCoordinatorRequest_v0 to node


williamsjj:
Upgraded to kafka-python 1.2.2 running against a 3-node Kafka 0.10 cluster. We're receiving this error once on every subscribe action:

```
ERROR:kafka.coordinator:Error sending GroupCoordinatorRequest_v0 to node XXX [XXX]
```

The node specified appears to be random.

dpkp:
What is under those XXXs ? In particular, the string in brackets should be the error type. What does it say?

dpkp:
I can't reproduce based on the information here. Please re-open if you have more!

hicqu:
recently I also has same problem. I'm working with kafka 0.9.0.1 and kafka-python 1.2.1.
Here is full error message:
Error sending GroupCoordinatorRequest_v0 to node 1 [NodeNotReadyError: 1]
So could you help me?

dpkp:
This error is logged when there is a networking failure. There isn't enough information here for me to provide any further advice. You might try enabling debug logs (`logging.basicConfig(level=logging.DEBUG)`) and/or checking your broker logs and configuration.

hicqu:
@dpkp , I have resolved this problem by setting request_timeout_ms to 300000 insttead of 40000 by default. So I think the reason is I fetch too many messages from kafka broker one time, and after I processed all of them, call consumer.poll() again, it's beyond request_timeout_ms. What do you think? If you don't agree me or you can not certain, I will attach more information here to help you analyze.

hicqu:
@dpkp , same error was found again. So I print kafka log at DEBUG level. I attach the log file below. Please check that and help me.
[kafka.zip](https://github.com/dpkp/kafka-python/files/436726/kafka.zip)

thrbowl:
+1, i got the same error.

fillest:
I don't see how I can re-open this issue, I don't see anything that could be related to "re-open" in the github UI.

I'm getting this error on localhost with a single broker and zookeeper on a fresh topic populated with some messages, it's clearly not a networking issue.
kafka-python==1.3.1, kafka 0.10.1.0
This is how i configure the consumer:
```python
topic_partition = kafka.TopicPartition('test', 0)
consumer = kafka.KafkaConsumer(
    bootstrap_servers = ['localhost'],
    client_id = 'kafka_cat',
    value_deserializer = cPickle.loads,
    auto_offset_reset = 'earliest',
[...]

PR #986, opened 2017-02-13
Fail-fast on timeout constraint violations during KafkaConsumer creation

harelba:
Related to #739

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -6,6 +6,8 @@ import socket
 import sys
 import time
 
+from kafka.errors import KafkaConfigurationError
+
 from kafka.vendor import six
 
 from kafka.client_async import KafkaClient, selectors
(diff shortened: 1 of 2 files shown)


harelba:
saw that there are failures in travis. I'm taking a look at it.

If you could provide instructions on how to run it locally with tox, that would be great.

harelba:
fixed - Another build has started. Monitoring.

jeffwidman:
Can we drop support for python 2.6? I'd rather be able to use `.format()`

Drizzt1991:
I like the idea of such errors.

kafka/consumer/group.py
```diff
@@ -267,6 +269,18 @@ def __init__(self, *topics, **configs):
                         new_config, self.config['auto_offset_reset'])
             self.config['auto_offset_reset'] = new_config
 
+        request_timeout_ms = self.config['request_timeout_ms']
+        session_timeout_ms = self.config['session_timeout_ms']
+        fetch_max_wait_ms = self.config['fetch_max_wait_ms']
+        if request_timeout_ms <= session_timeout_ms:
+            raise KafkaConfigurationError(
+                "Request timeout (" + str(request_timeout_ms) + ") must be larger than session timeout (" +
```
dpkp: `"Request timeout (%s) must be larger than session timeout (%s)" % (request_timeout_ms, session_timeout_ms)` is another alternative here

harelba: true, and better of course. fixing.

harelba:
fixed string interpolation according to comment. Travis running.

kafka/consumer/group.py
harelba: done, pushed fix.

dpkp:
Thanks!

PR #991, opened 2017-02-20
change close() timeout to threading.TIMEOUT_MAX

mmyjona:


jeffwidman:
Why the need for this? And why wrap in a try-catch block rather than just changing the existing value?

mmyjona:
#990 failed in python2.7 and 2.6

kafka/producer/kafka.py
```diff
@@ -400,7 +400,10 @@ def close(self, timeout=None):
             log.info('Kafka producer closed')
             return
         if timeout is None:
-            timeout = 999999999
+            try:
+                timeout = threading.TIMEOUT_MAX
```
dpkp: perhaps a comment that this is available in python 3.3+

dpkp: and do we also need to assert that a user-supplied timeout is <= threading.TIMEOUT_MAX if on python 3.3+ ?

dpkp:
Lint failing on python2.7. May need some pylint magic, or else use getattr.

dpkp:
Merged as `ebb6c3bd85638d4f09c0b392a7de65c63a8a20da`

PR #992, opened 2017-02-21
Run tests in python3.6 too

ms7s:


dpkp:
My main concern with this is the overall time to run the test suite on travis. I think we should consider dropping automated tests for 3.3 and possibly 3.4

ms7s:
+1 for dropping 3.3, but I'd probably wait with 3.4 a while longer. Python 3.4 is the default python3 in Ubuntu 14.04 LTS, so maybe wait until 18.04 is released next year? Python 3.7 beta4 should be released around that time too, so we can swap 3.4 with 3.7.

dpkp:
Sounds good to me. Test failures look like flappers and I restarted them (if you want to help fix flapping tests, that would be awesome too!)

PR #995, opened 2017-02-24
Short-circuit group coordinator requests when NodeNotReady

dpkp:
This PR attempts to reduce error logging caused by _failed_request, primarily for JoinGroupRequest. Also included are snippets for SyncGroupRequest and HeartbeatRequest, though logically these should be less likely to trigger as the coordinator connection should already be primed by a successful JoinGroupRequest the time those requests are sent.

### DIFF
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -422,9 +422,11 @@ class BaseCoordinator(object):
             e = Errors.GroupCoordinatorNotAvailableError(self.coordinator_id)
             return Future().failure(e)
 
-        elif not self._client.ready(self.coordinator_id):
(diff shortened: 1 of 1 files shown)


jeffwidman:
LGTM

dpkp:
It's troubling that 4 tests failed all in the exact same way -- timeout during consumer group test.

dpkp:
The problem here is that KafkaClient.ready() also returns false if there is a metadata request in progress. But we added a metadata update request during the group assignment process as part of KAFKA-3117 (`dd50847a`). I think the solution here is to add an optional flag to `ready()` that allows a user to skip the metadata request check.

dpkp:
Or just skip the NodeNotReady check when sending SyncGroup

PR #1001, opened 2017-03-01
Add send_list_offset_request for searching offset by timestamp

charsyam:
Currently, Kafka Python has OffsetRequest_v1(It's name changed as ListOffsetRequest v1)
But there is no way to using this request. this is very helpful for searching offset by timestamp.

so, I add send_list_offset_request to request OffsetRequest_v1 and OffsetResponse_v1
and this feature only can use kafka 0.10.1.x or higher.

and, I missed test code, when I looked OffsetRequest test code, they are all skip annotation.
How do I add test for this feature?

Thanks.

```
from kafka import KafkaClient
from kafka.structs import ListOffsetRequestPayload

client = KafkaClient(hosts=['broker1:9092', 'broker2:9092', 'broker3:9092'])
reqs = []
reqs.append(ListOffsetRequestPayload("test2", 0, 1488161648929))
v = client.send_list_offset_request(reqs)
print(v)
```

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -686,6 +686,16 @@ class SimpleClient(object):
         return [resp if not callback else callback(resp) for resp in resps
                 if not fail_on_error or not self._raise_on_response_error(resp)]
 
+    def send_list_offset_request(self, payloads=[], fail_on_error=True,
+                            callback=None):
+        resps = self._send_broker_aware_request(
+            payloads,
+            KafkaProtocol.encode_list_offset_request,
+            KafkaProtocol.decode_list_offset_response)
+
+        return [resp if not callback else callback(resp) for resp in resps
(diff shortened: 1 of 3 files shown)


jeffwidman:
You added it to the old `SimpleClient` implementation which is deprecated and has minimal support at a bugfix level, not much of feature-request level. 

You should probably switch to the new `KafkaClient` implementation, which lives in `client_async.py` and uses `Structs` rather than payloads. Looks like there's already a [OffsetRequest_v1 struct](https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/offset.py#L56). 

Have you tried using that?

dpkp:
Agree with jeff -- this PR changes unsupported legacy code. But the code looks fine and tests pass, so I don't have a problem merging if it's helpful in your context.

I would definitely recommend switching to the new KafkaConsumer.

Issue #948, opened 2017-01-18
KIP-62 / KAFKA-3888: Allow consumer to send heartbeats from a background thread


jeffwidman:
Allows consumers to take their time processing messages without being timed out from their consumer group. 

`max_poll_records` is a decent workaround for most of the pain here, but it'd still be nice to add this for consumers that have inconsistent message processing times.

Drizzt1991:
Hey @jeffwidman do you plan to do any work on this? If not I probably can take it on the weekend.

jeffwidman:
No, I'm afraid I won't have time to work on this anytime soon... if you have time, go for it!

Only comment--and you probably already know this--is as far as implementation, the closer it mirrors the Java consumer (without the bugs 😉 ), the better IMHO. Among other things, it makes it so config tuning advice from the Kafka mailing list etc is easier to translate to `kafka-python`.

jeffwidman:
We were just discussing this internally... since Python has the GIL, how would this background thread ever issue a heartbeat if the main consumer thread is CPU-bound in it's message processing?

Drizzt1991:
Hey @jeffwidman 
Sorry, but GIL will not block you from doing IO in a separate thread. At least if you are not doing something crazy like passing all messages in a pure C code, that processes them. Python will yield GIL after some time (python3) or instruction count (python2), so as long as you are not stuck in C code it should be OK.

jianbin-wei:
I agree that in most cases the GIL would not prevent the heart-beat background thread from doing its job.

jeffwidman:
@Drizzt1991 are you still planning on working on this? 

It's not something I can tackle in the near term, just wanted to check in to see where you're at on it as it would be a very useful addition...

Drizzt1991:
@jeffwidman Hey, sorry, but not seems possible in near term. I did look at how Java client does it, quite a hassle with locking, not sure I can do it properly in my limited free time.

oscarbatori:
@jeffwidman @Drizzt1991 - is there a pull request for this? We would quite like to leverage KIP-62 functionality in our organization.

Drizzt1991:
@oscarbatori Sorry, nothing too good to show...

pvanderlinden:
[...]

PR #1002, opened 2017-03-02
Bump KafkaConsumer's request_timeout to 305000ms

jeffwidman:
The Java Consumer [changed the value to 305000 in `0.10.1.0`](https://kafka.apache.org/documentation/#upgrade_1010_notable):

>  The new Java Consumer now supports heartbeating from a background thread. There is a new configuration max.poll.interval.ms which controls the maximum time between poll invocations before the consumer will proactively leave the group (5 minutes by default). The value of the configuration request.timeout.ms must always be larger than max.poll.interval.ms because this is the maximum time that a JoinGroup request can block on the server while the consumer is rebalancing, so we have changed its default value to just above 5 minutes. Finally, the default value of session.timeout.ms has been adjusted down to 10 seconds, and the default value of max.poll.records has been changed to 500.

See the current upstream value here: https://kafka.apache.org/documentation/#configuration

This value was only changed in `KafkaConsumer`. So we don't also need to update `KafkaProducer` / `KafkaClient` as the upstream Java implementations still default to `30000`. The broker default is also `30000`.

----

Note: 
Merging this may be slightly premature since #948 hasn't been done to add `max.poll.interval.ms` yet, but I don't see any harm in adding this now. 

Sessions will still timeout, this just gives the brokers/cluster a bit longer to handle things like consumer group coordination etc.

In one of our dev environments, we were seeing a lot of these timeouts at the current default of `40000` using the new `KafkaConsumer`. Weirdly, the timeouts were being hit even though the consumer had no messages to consume and they were starting with default offset of `latest`. 

This doesn't quite feel right. It's a dev cluster, so everything on VMs and the network can be flaky/slow. But I still wouldn't have expected a consumer that was simply sitting there would be hitting this request timeout at 40 seconds.

Still, when I bumped their request timeouts up to this new `305000` value, the timeout errors went away, and the consumers remained stable as best I could tell.

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -74,7 +74,7 @@ class KafkaConsumer(six.Iterator):
             happens, the consumer can get stuck trying to fetch a large
             message on a certain partition. Default: 1048576.
         request_timeout_ms (int): Client request timeout in milliseconds.
-            Default: 40000.
+            Default: 305000.
         retry_backoff_ms (int): Milliseconds to backoff when retrying on
             errors. Default: 100.
         reconnect_backoff_ms (int): The amount of time in milliseconds to
@@ -213,7 +213,7 @@ class KafkaConsumer(six.Iterator):
         'fetch_max_wait_ms': 500,
         'fetch_min_bytes': 1,
         'max_partition_fetch_bytes': 1 * 1024 * 1024,
-        'request_timeout_ms': 40 * 1000,
+        'request_timeout_ms': 305000,
         'retry_backoff_ms': 100,
         'reconnect_backoff_ms': 50,
         'max_in_flight_requests_per_connection': 5,


jeffwidman:
This may be slightly premature since #948 hasn't added `max.poll.interval.ms` yet, but I don't see any harm in adding this now. 

Sessions will still timeout, this just gives the broker/cluster a bit longer to handle things like consumer group coordination etc. 

We were seeing timeouts in dev clusters using the current default of 4 seconds and I suspect it's because the consumer group coordination wasn't finished due to slow/flaky network connections within our dev environment.

jeffwidman:
Actually, don't merge yet, I need to also check if this should be changed in `KafkaProducer` and `KafkaClient`

jeffwidman:
In one of our dev environments, we were seeing a lot of these timeouts at the current default of `40000` using the new `KafkaConsumer`. Weirdly, the timeouts were being hit even though the consumer had no messages to consume and they were starting with default offset of `latest`. 

This doesn't quite feel right. It's a dev cluster, so everything on VMs and the network can be flaky/slow, but I still wouldn't have expected a consumer that was doing nothing other than sitting there should be hitting this request timeout at 40 seconds.

Still, when I bumped their request timeouts up to this new `305000` value, the timeout errors went away, and the consumers remained stable as best I could tell.

dpkp:
It does seem strange to hit 40sec timeouts in an environment like that.

dpkp:
Would like to wait on changing this default until the background heartbeat thread is implemented

yaman-jain:
Hi @dpkp,
Possible to merge this PR?
Thanks in advance

jeffwidman:
Closing in favor of #1266

PR #1017, opened 2017-03-06
Add optional kwarg to disable metadata-priority logic

dpkp:
This should not change existing behavior. Intent here is simply to support using ready without the metadata-priority check. This may be useful with group coordinator logic.


dpkp:
These failures are repeatable and very strange...

dpkp:
Failures are from master. Fix pushed there and I'll rebase to get fresh tests here.

PR #1032, opened 2017-03-13
BrokerConnection receive bytes pipe

dpkp:
This is a first step towards #560 . Refactors the internals of BrokerConnection to separate handling network IO from decoding responses. The initial refactor only addresses recv . Refactoring send will be tackled in a separate PR. The eventual goal is to separate the base protocol parser into an isolated class that is completely independent of network IO.


dpkp:
There's a test failure that needs further investigation

PR #1037, opened 2017-03-19
fix AttributeError NoneType thrown by calling consumer subscription

pseudocodes:
when calling consumer.subscription() without subscribe any topic, we got 

```python
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "./kafka_consumer.py", line 110, in run
    print consumer.subscription()
  File "/workplace/kafka-python/kafka/consumer/group.py", line 828, in subscription
    return self._subscription.subscription.copy()
AttributeError: 'NoneType' object has no attribute 'copy'
```
this pr just fix the subscription default value


dpkp:
I think the fix will need to be different (see test failures). I'm going to file an issue and close this PR.

PR #1062, opened 2017-03-28
Don't fall into infinite cycle if future is lost during connection loss

sibiryakov:
This is simply it.
Future could be attached to some request and it should fire when response is received. If connection is lost, then request is also lost and cycle in ``KafkaClient.poll()`` will be infinite, causing the whole application to stuck without any ability to recover.

```python
            request = OffsetRequest[0](
                -1, [(topic, [(partition.partition, timestamp, 1) for partition in partitions])]
            )
            future_request = Future()
            _f = self._client.send(node_id, request)
            _f.add_callback(self._handle_offset_response, partitions, future_request)
            _f.add_errback(lambda e: future_request.failure(e))
              # blocks forever, if connection is lost after request is sent
              self._client.poll(future=future_request)
```

I propose to use ``request_timeout_ms`` as a maximum time that could be spend waiting for the future to complete.

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -489,6 +489,7 @@ class KafkaClient(object):
             timeout_ms = self.config['request_timeout_ms']
 
         responses = []
+        start_loop = time.time()
 
         # Loop for futures, break after first loop if None
         while True:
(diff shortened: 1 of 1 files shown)


dpkp:
Why is this necessary? KafkaClient._poll checks for request timeouts already and will close the connection and cause all pending futures to complete w/ RequestTimedOutError. But maybe I don't fully understand the error case you are seeing?

jeffwidman:
bump @sibiryakov

jeffwidman:
Closing due to inactivity, feel free to re-open @sibiryakov if you can address comments above.

Issue #1050, opened 2017-03-23
Implement `connections_max_idle_ms` in KafkaProducer


jeffwidman:
https://github.com/dpkp/kafka-python/blob/82d50f443e04356b2f051f7476bb4b4f5bd700d2/kafka/producer/kafka.py#L269

Would be useful per my notes in https://github.com/dpkp/kafka-python/issues/849#issuecomment-261415353

Issue #1049, opened 2017-03-23
Implement `connections_max_idle_ms` in KafkaConsumer


jeffwidman:
https://github.com/dpkp/kafka-python/blob/92a66e3009147a9909f32df2adedce831b7fc7fb/kafka/consumer/group.py#L257

Would be useful per my notes in https://github.com/dpkp/kafka-python/issues/849#issuecomment-261415353

PR #1068, opened 2017-04-05
Timeout idle connections via connections_max_idle_ms

### LINKED ISSUE #548: Support connections_max_idle_ms
KafkaClient should disconnect sockets that have not been active for the configured max idle duration.

dpkp:
Fixes #548 #1049 #1050

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -135,6 +135,7 @@ class KafkaClient(object):
         'bootstrap_servers': 'localhost',
         'client_id': 'kafka-python-' + __version__,
         'request_timeout_ms': 40000,
+        'connections_max_idle_ms': 9 * 60 * 1000,
         'reconnect_backoff_ms': 50,
         'max_in_flight_requests_per_connection': 5,
         'receive_buffer_bytes': None,
@@ -194,6 +195,7 @@ class KafkaClient(object):
         self._wake_r.setblocking(False)
         self._wake_lock = threading.Lock()
         self._selector.register(self._wake_r, selectors.EVENT_READ)
+        self._idle_expiry_manager = IdleConnectionManager(self.config['connections_max_idle_ms'])
         self._closed = False
         self._sensors = None
         if self.config['metrics']:
@@ -291,6 +293,8 @@ class KafkaClient(object):
             if self._sensors:
                 self._sensors.connection_created.record()
 
+            self._idle_expiry_manager.update(node_id)
+
             if 'bootstrap' in self._conns and node_id != 'bootstrap':
                 bootstrap = self._conns.pop('bootstrap')
                 # XXX: make conn.close() require error to cause refresh
@@ -308,7 +312,13 @@ class KafkaClient(object):
                 pass
             if self._sensors:
                 self._sensors.connection_closed.record()
-            if self._refresh_on_disconnects and not self._closed:
+
+            idle_disconnect = False
+            if self._idle_expiry_manager.is_expired(node_id):
+                idle_disconnect = True
+            self._idle_expiry_manager.remove(node_id)
+
+            if self._refresh_on_disconnects and not self._closed and not idle_disconnect:
                 log.warning("Node %s connection failed -- refreshing metadata", node_id)
                 self.cluster.request_update()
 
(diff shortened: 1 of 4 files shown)


jeffwidman:
You have a bunch of `return None` calls... why not just `return`? It's more idiomatic...

dpkp:
I chose `return None` so as to contrast more explicitly with the tuple return value when there is something to expire.

Re consumer and producer, connections are created as needed in normal course. Disconnecting idle connections should not have any impact.

kafka/client_async.py
```diff
@@ -920,6 +943,76 @@ def pop_ready(self):
...
+            del self.lru_connections[conn_id]
+        self.lru_connections[conn_id] = time.time()
+
+    def remove(self, conn_id):
+        if conn_id in self.lru_connections:
+            del self.lru_connections[conn_id]
+
+    def is_expired(self, conn_id):
+        if conn_id not in self.lru_connections:
+            return None
```
jianbin-wei: Should this be False instead of None to be consistent for the return type `bool`?

kafka/client_async.py
```diff
@@ -920,6 +943,76 @@ def pop_ready(self):
...
+            return int((self.next_idle_close_check_time - now) * 1000)
+
+    def update_next_idle_close_check_time(self, ts):
+        self.next_idle_close_check_time = ts + self.connections_max_idle
+
+    def poll_expired_connection(self):
+        if time.time() < self.next_idle_close_check_time:
+            return None
+
+        if not len(self.lru_connections):
```
jianbin-wei: redundant `len()` check:  `if not self.lru_connections:`

kafka/client_async.py
```diff
@@ -920,6 +943,76 @@ def pop_ready(self):
...
+            return float('inf')
+        elif self.next_idle_close_check_time <= now:
+            return 0
+        else:
+            return int((self.next_idle_close_check_time - now) * 1000)
+
+    def update_next_idle_close_check_time(self, ts):
+        self.next_idle_close_check_time = ts + self.connections_max_idle
+
+    def poll_expired_connection(self):
```
jianbin-wei: Personally, I would return an iterator for all expired connections instead of `None`.  In the call to this function, it would iterate through and close them.

kafka/producer/kafka.py
```diff
@@ -266,7 +266,7 @@ class KafkaProducer(object):
         'linger_ms': 0,
         'partitioner': DefaultPartitioner(),
         'buffer_memory': 33554432,
-        'connections_max_idle_ms': 600000,  # not implemented yet
+        'connections_max_idle_ms': 9 * 60 * 1000,  # not implemented yet
```
jianbin-wei: Should remove the "not implemented yet" comment? 😉

jeffwidman: I think there's a similar comment in the consumer that also needs updating

kafka/client_async.py
dpkp: I chose None here to be "falsey" but distinguishable from False since the connection in this case is not "unexpired" but rather "unknown".

kafka/client_async.py
dpkp: true

kafka/client_async.py
dpkp: we could do that, but I chose to stay fairly close to the java client logic just to be safe. I believe this implementation works, and I'm also skeptical that there's much performance optimization to be gained given the relative infrequency with which connection timeouts are likely to occur.

kafka/producer/kafka.py
dpkp: Yes -- this is fixed up in master

kafka/client_async.py
jianbin-wei: I didn't check Java client implementation.  If that is the case, I would also try to be close to Java' client behavior.  It is not to gain performance but convenient.

kafka/client_async.py
jianbin-wei: That can be confusing as the next return is bool type.

In which case that one connection would not be managed by this and None is returned?

PR #1074, opened 2017-04-10
Update RTD links with https / .io

jeffwidman:


dpkp:
Thanks -- looks like there's also a couple links under docs/ that could be updated

PR #1085, opened 2017-04-24
Fix #1083: Deal with brokers that disappear, reappear with different IP address

### LINKED ISSUE #1083: Producer doesn't reconnect if broker goes down, reappears with new IP
I'm using Docker swarm to manage Kafka brokers, so if a broker goes down, swarm will automatically create a new instance in new container. The catch is that the new container will likely have a different IP address from the old one. The issue is this: once that new broker comes up, kafka-python will never connect to it.

Here's an example running on my local machine, where I'm running a single node that can be discovered with 'kafka.local'. I run one Kafka broker which magically sets its advertised.listeners to the container's IP, which in this case was 172.17.0.4 for the old container and 172.17.0.5 for the new one. (both Kakfa and Zookeeper are in the same container, but I'm using a volume to persist their data so clients think it's the same broker when it finally comes back up)

Logs follow.

First, the node goes down and you see a lot of this:
```
2017-04-24T02:55:27.652Z conn.py:211 DEBUG:kafka.conn: <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092>: creating new socket
2017-04-24T02:55:27.652Z sender.py:109 DEBUG:kafka.producer.sender: Node 0 not ready; delaying produce of accumulated batch
2017-04-24T02:55:27.653Z conn.py:295 ERROR:kafka.conn: Connect attempt to <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092> returned error 111. Disconnecting.
2017-04-24T02:55:27.653Z client_async.py:304 WARNING:kafka.client: Node 0 connection failed -- refreshing metadata
2017-04-24T02:55:27.654Z conn.py:211 DEBUG:kafka.conn: <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092>: creating new socket
2017-04-24T02:55:27.655Z sender.py:109 DEBUG:kafka.producer.sender: Node 0 not ready; delaying produce of accumulated batch
2017-04-24T02:55:27.655Z conn.py:295 ERROR:kafka.conn: Connect attempt to <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092> returned error 111. Disconnecting.
2017-04-24T02:55:27.656Z client_async.py:304 WARNING:kafka.client: Node 0 connection failed -- refreshing metadata
2017-04-24T02:55:27.656Z client_async.py:656 ERROR:kafka.client: No nodes found in metadata -- retrying bootstrap
2017-04-24T02:55:27.656Z client_async.py:213 DEBUG:kafka.client: Sleeping 12.6888 before bootstrapping again
```
Meanwhile, we send more messages:
```
2017-04-24T02:55:30.202Z kafka.py:508 DEBUG:kafka.producer.kafka: Sending (key=1 value=1) to TopicPartition(topic='_test-count', partition=0)
```
kafka-python realizes something's terribly wrong and tries bootstrapping again using the original bootstrap_server value, which is good:
```
2017-04-24T02:55:40.355Z client_async.py:223 DEBUG:kafka.client: Attempting to bootstrap via node at kafka.local:9092
2017-04-24T02:55:40.357Z conn.py:211 DEBUG:kafka.conn: <BrokerConnection host=kafka.local/kafka.local port=9092>: creating new socket
2017-04-24T02:55:40.358Z conn.py:281 DEBUG:kafka.conn: <BrokerConnection host=kafka.local/127.0.0.1 port=9092>: established TCP connection
2017-04-24T02:55:40.359Z client_async.py:274 DEBUG:kafka.client: Node bootstrap connected
2017-04-24T02:55:40.362Z conn.py:545 DEBUG:kafka.conn: <BrokerConnection host=kafka.local/127.0.0.1 port=9092> Request 1: MetadataRequest_v1(topics=NULL)
2017-04-24T02:55:40.387Z conn.py:721 DEBUG:kafka.conn: <BrokerConnection host=kafka.local/127.0.0.1 port=9092> Response 1: MetadataResponse_v1(brokers=[(node_id=0, host='172.17.0.5', port=9092, rack=None)], controller_id=0, topics=[(error_code=0, topic='_test-count', is_internal=False, partitions=[(error_code=0, partition=0, leader=0, replicas=[0], isr=[0])]), (error_code=0, topic='aqi-ground-station-raw-data', is_internal=False, partitions=[(error_code=0, partition=0, leader=0, replicas=[0], isr=[0])])])
2017-04-24T02:55:40.388Z cluster.py:272 DEBUG:kafka.cluster: Updated cluster metadata to ClusterMetadata(brokers: 1, topics: 2, groups: 0)
```
It's successful, but the client object doesn't seem to notice:
```
2017-04-24T02:55:40.389Z client_async.py:707 DEBUG:kafka.client: Give up sending metadata request since no node is available
```
And continues doing this forever:
```
2017-04-24T02:55:40.391Z conn.py:211 DEBUG:kafka.conn: <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092>: creating new socket
2017-04-24T02:55:40.392Z sender.py:109 DEBUG:kafka.producer.sender: Node 0 not ready; delaying produce of accumulated batch
2017-04-24T02:55:40.392Z conn.py:295 ERROR:kafka.conn: Connect attempt to <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092> returned error 111. Disconnecting.
2017-04-24T02:55:40.393Z client_async.py:304 WARNING:kafka.client: Node 0 connection failed -- refreshing metadata
2017-04-24T02:55:40.394Z conn.py:211 DEBUG:kafka.conn: <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092>: creating new socket
2017-04-24T02:55:40.395Z sender.py:109 DEBUG:kafka.producer.sender: Node 0 not ready; delaying produce of accumulated batch
2017-04-24T02:55:40.396Z conn.py:295 ERROR:kafka.conn: Connect attempt to <BrokerConnection host=172.17.0.4/172.17.0.4 port=9092> returned error 111. Disconnecting.
```
Another variation of this bug happens when, during bootstrap, the name in bootstrap_servers fails to resolve (socket.gaierror) and comes back up later. The reconnect logic seems complicated though, so I doubt the same fix would work for that.

This is a ship stopper for me (and should be for anyone else who's running Kakfa in docker swarm) so I'm going to start digging in and come up with a fix. Any help would be appreciated though!

originsmike:
When KafkaClient connects to a broker in _maybe_connect, it inserts into self._conns a BrokerConnection configured with the current host/port for that node.  The BrokerConnection remains there forever, though, so if the broker's IP or host ever changes, KafkaClient has no way to deal with this.

The fix is to compare the latest metadata with the current node's connection, and if the host/IP has changed, decommission the old connection and allow a new one to be created.

There's also a common race condition on broker startup where the initial metadata request sometimes returns an empty list of brokers, but subsequent requests behave normally.  So, we must deal with broker being None here.  This change is conservative in that it doesn't remove the connection from self._conns unless the new broker metadata contains an entry for that same node with a new IP/port.

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -324,8 +324,19 @@ class KafkaClient(object):
 
     def _maybe_connect(self, node_id):
         """Idempotent non-blocking connection attempt to the given node id."""
+        broker = self.cluster.broker_metadata(node_id)
+
+        # If broker metadata indicates that a node's host/port has changed, remove it
+        if node_id in self._conns and broker is not None:
+            conn = self._conns[node_id]
+            host, _, __ = get_ip_port_afi(broker.host)
+            if conn.host != host or conn.port != broker.port:
+                log.debug("Closing connection to decommissioned node %s at %s:%s",
+                          node_id, conn.host, conn.port)
+                conn.close()
+                self._conns.pop(node_id)
+
         if node_id not in self._conns:
-            broker = self.cluster.broker_metadata(node_id)
             assert broker, 'Broker id %s not in current metadata' % node_id
 
             log.debug("Initiating connection to node %s at %s:%s",


originsmike:
Tested with a running producer and consumer connected to a single Kafka broker, which I repeatedly killed and restarted, sometimes with a new IP. Let me know if there's any docs that describe how to run the other automated tests.

In testing, producer and consumer always reconnected, but under some circumstances a few messages were lost: depending on when the reconnect happens during the broker's boot procedure, the error message returned could be different, so sometimes Sender._can_retry will return true and sometimes not. Different issue.

jeffwidman:
> under some circumstances a few messages were lost: depending on when the reconnect happens during the broker's boot procedure, the error message returned could be different, so sometimes Sender._can_retry will return true and sometimes not. Different issue.

Can you file a issue for this? Don't want to lose track of it.

dpkp:
Looks good! Test failures are unrelated, caused by a pylint bug.

dpkp:
I've fixed the pylint errors in master. Can you rebase to get a clean test run?

jeffwidman:
bump @originsmike

dpkp:
`_maybe_connect` is called before every network operation, so I think we should be very conservative about when to run this code. Can we skip this check if the existing socket is connected and let a normal connection error propagate if a broker disappears? Once the socket is disconnected, the next _maybe_connect should handle removing the stale conn object.

dpkp:
Pushed to master w/ fixup to only check disconnected nodes. Thanks!

PR #1091, opened 2017-04-29
Warn on DNS lookup failures

dpkp:
BrokerConnection currently raises a socket exception if dns resolution of broker metadata fails. This error is typically unhandled by application code and causes a hard crash. In many cases, this error is recoverable after short delay (temporary dns infrastructure problems) and/or refresh metadata (configuration errors). This PR changes the response to dns lookup errors to no longer raise an exception and instead just log a warning.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -237,11 +237,13 @@ class BrokerConnection(object):
                                                        socket.AF_UNSPEC,
                                                        socket.SOCK_STREAM)
                     except socket.gaierror as ex:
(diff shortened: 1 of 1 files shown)


jianbin-wei:
I experienced that error while doing tests using docker.  Nice to have this change.

One question: will the retry handle by the client or need to be handled by application?

dpkp:
Retries should happen as with any other connection failure. Which means that for metadata requests, the retries will happen transparently. Consumer fetches should also retry transparently. For produce requests, the batch will use the `retries` configuration and may fail if the error is not resolved before the maximum retry attempts are exhausted.

Issue #1094, opened 2017-05-02
Flush Raising AssertionError


infecto:
https://github.com/dpkp/kafka-python/blob/07e09c1c2ec6787fc7e4f3c2578d31b4a15d20bc/kafka/producer/record_accumulator.py#L525

We are still unsure why but we are encountering the above assertion failing in a small percentage of cases. Maybe network issues, rebalancing, the reason has yet to be determined. 

This AssertionError bleeds through. Instead I think kafka-python should be catching it and raising a defined exception within this framework. Any thoughts?

dpkp:
Agree. `producer.flush()` should either raise a KafkaTimeout exception or just return False if the flush did not complete within the specified timeout. Happy to merge a PR for this.

infecto:
@dpkp Will create a PR. Thanks

infecto:
@dpkp https://github.com/dpkp/kafka-python/pull/1097

PR #1097, opened 2017-05-04
Flush timeout raises timeout

infecto:
https://github.com/dpkp/kafka-python/issues/1094

- Renamed `TimeoutException` to `KafkaTimeoutError`. Outside of the scope of original PR, let me know if I should pull this out.
- Catching both timeout and done assertion checks. My thinking was that if the user did not specify a timeout we would never hit the done check. If the user did specify a timeout and the done check fails it seems reasonable to consider it a timeout. I might be making too large of an assumption here.

### DIFF
--- a/kafka/producer/record_accumulator.py
+++ b/kafka/producer/record_accumulator.py
@@ -526,8 +526,11 @@ class RecordAccumulator(object):
             for batch in self._incomplete.all():
                 log.debug('Waiting on produce to %s',
                           batch.produce_future.topic_partition)
-                assert batch.produce_future.wait(timeout=timeout), 'Timeout waiting for future'
-                assert batch.produce_future.is_done, 'Future not done?'
+                if not batch.produce_future.wait(timeout=timeout):
+                    raise Errors.KafkaTimeoutError('Timeout waiting for future')
+                if not batch.produce_future.is_done:
+                    raise Errors.UnknownError('Future not done')
+
                 if batch.produce_future.failed():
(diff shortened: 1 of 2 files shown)


kafka/producer/record_accumulator.py
```diff
@@ -526,8 +526,12 @@ def await_flush_completion(self, timeout=None):
             for batch in self._incomplete.all():
                 log.debug('Waiting on produce to %s',
                           batch.produce_future.topic_partition)
-                assert batch.produce_future.wait(timeout=timeout), 'Timeout waiting for future'
-                assert batch.produce_future.is_done, 'Future not done?'
+                try:
+                    assert batch.produce_future.wait(timeout=timeout), 'Timeout waiting for future'
+                    assert batch.produce_future.is_done, 'Future not done?'
+                except AssertionError as e:
```
dpkp: might as well simply drop the asserts and use `if not`

though technically if wait() returns True, but is_done is False then it is not a Timeout but an unexpected error state, for which we could raise a timeout error but we also might pick some other more obviously bad exception.

dpkp:
Looks good -- thanks!!

infecto:
@dpkp Sorry for the delay was sick. Updated PR to reflect your comments.

infecto:
@dpkp I know time is precious but could I get a one last review over this. Don't want it to get too stale. Thanks.

jeffwidman:
This looks very straightforward, merging.

PR #1125, opened 2017-06-19
Backoff on unavailable group coordinator retry

dpkp:
When testing 0.11.0.0 I noticed that KafkaConsumer would enter a tight CPU loop if the broker's group coordinator api returned retriable errors.


jeffwidman:
Can you describe what you mean by `tight CPU loop`? Was it merely a resource hog, or did it actually inhibit functionality of the consumer in some way?

PR #1126, opened 2017-06-19
producer: Set exit timeout to 0 for atexit handler to match __del__

Ormod:
Hit a problem with pytest hitting the atexit handler and waiting
for close() timeout forever at teardown.

This commit makes atexit close() equivalent to __del__ behavior,
namely using timeout of 0 for close() completion. If you need a
longer timeout you should be setting it explicitly.


dpkp:
+1

PR #1132, opened 2017-06-26
change_subscription warning log to info

Artimi:
When we are using subscription by pattern change subscription is
called every metadata update even when nothing changes. It logs
warning every time topics are not changed. It creates unnecessary
noise in logs. I changed it to level info so it can be easily ignored
(I want to see only warnings from kafka client). Do you think this is
a right approach or I should check in `ConsumerCoordinator._handle_metadata_update`
that no topics has changed and not call `change_subscription` in 
this case? I would then need to change it in aiokafka too because I'm
using that.

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -140,8 +140,9 @@ class ConsumerCoordinator(BaseCoordinator):
                 if self._subscription.subscribed_pattern.match(topic):
                     topics.append(topic)
 
-            self._subscription.change_subscription(topics)
-            self._client.set_topics(self._subscription.group_subscription())
+            if set(topics) != self._subscription.subscription:
+                self._subscription.change_subscription(topics)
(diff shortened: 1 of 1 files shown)


jeffwidman:
@dpkp how much extra work would it save to skip calling `change_subscription` if nothing actually changed? Feels like that's an optimization probably worth making unless you're trying to guard against edge-cases.

And I agree, this log should be `info` (I'd even go as far as making it `debug`) if nothing actually changed.

Artimi:
@jeffwidman I changed it so we don't call `change_subscription` when it is not needed. I don't think that this can break something but lets see how tests will turn out.

dpkp:
I would prefer to keep the log level at warning. Otherwise, this is a good improvement. +1

Artimi:
@dpkp level changed back to warning

dpkp:
Thanks!

PR #1133, opened 2017-06-26
Prevent Fetcher from refetching messages when consuming compacted topics with poll() and max_records

flavray:
When a topic is compacted, consecutive messages might not have
consecutive offsets. `Fetcher._append` seems to  discard
`PartitionRecords` whenever the offset of the first message of a part
was not equal to the offset of the last message of the previous part + 1.
This is almost always the case for compacted topics (at least when
fetching from the 'earliest' offset).
Using `part.fetch_offset` instead ensures the whole
`PartitionRecords` is not discarded the first time offsets are not
consecutive, avoiding sending "useless" new FetchRequests.

In my case, using `consumer.poll(max_records=50)`, the first
`FetchResponse` returned ~13,000 records, ~12,950 were discarded
because the offset of the 51st record was not equal to the offset of
the 50th record + 1, a new `FetchRequest` was sent, and so on....
With this change, the whole ~13,000 messages were correctly used and
only one `FetchRequest` had to be sent.
(The topic was `__consumer_offsets` which is compacted).

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -360,7 +360,7 @@ class Fetcher(six.Iterator):
                 part_records = part.take(max_records)
                 if not part_records:
                     return 0
-                next_offset = part_records[-1].offset + 1
+                next_offset = part.fetch_offset
 
                 log.log(0, "Returning fetched records at offset %d for assigned"
                            " partition %s and update position to %s", position,


dpkp:
I haven't had a chance to dig into this one, but this issue is definitely serious. I need to convince myself that the solution you've proposed doesn't cause problems in other use cases, but we definitely need to fix this if indeed it is a bug.

dpkp:
Looked a bit deeper. This isn't as serious as I thought (the title made me think there was message loss -- there isn't). Basically the issue here is that when using consumer.poll() with max_records, there will be "unnecessary" additional FetchRequests if/when the "max_record" cutoff lands in a compacted (i.e., deleted) message range.

I looked at the official java client and it appears that this logic is the same through 0.10.2.1:
https://github.com/apache/kafka/blob/0.10.2.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L524

The v2 message format refactor in 0.11.0.0 updates a large part of the Fetcher, including a change to how fetch responses are buffered and iterated internally. I think it would be better to keep aligned with the java client than to attempt to tweak the logic for optimizations like this.

So I'm going to wontfix on this PR. If you want to avoid these refetches in the current kafka-python implementation, you might consider switching to the iterator interface. It does not have the same problem since it maintains internal state within a generator / yield loop:

```
max_records = 50
msgs = [next(consumer) for _ in range(max_records)]
```

PR #1146, opened 2017-07-12
Use for join-time-max and sync-time-max metrics Max() 

billyevans:
Seems like copy-paste typo.
Use Max() instead of Avg() for join-time-max and sync-time-max metrics.


dpkp:
Thanks !

PR #1149, opened 2017-07-14
Additional note about order to docs

tvoinarovskyi:
Add note, that `max_in_flight_requests_per_connection>1` may change order or messages to `max_in_flight_requests_per_connection` description. I know, that `retries` mentions this, but often you only change max_in_flight_requests_per_connection without reading all option descriptions.


dpkp:
Thanks!

PR #1152, opened 2017-07-17
added gssapi support (Kerberos) for SASL

Harald-Berghoff:
this is the first version, please have a look on it

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -55,17 +55,19 @@ except ImportError:
         pass
 
 # needed for SASL_GSSAPI authentication:
-disable_gssapi = False
 try:
     import gssapi
-    from gssapi import raw
+    #from gssapi import raw
 except ImportError:
     #no gssapi available, will disable gssapi mechanism
-    disable_gssapi = True
-    class GSSError(Exception):
-        pass
+    gssapi = None
+    #class GSSError(Exception):
+    #    pass
 
-
+try:
+    from gssapi.raw.misc import GSSError
+except ImportError:
+    GSSError = None
 
 
 
@@ -183,9 +185,9 @@ class BrokerConnection(object):
         'sasl_mechanism': 'PLAIN',
         'sasl_plain_username': None,
         'sasl_plain_password': None,
-        'sasl_servicename_kafka':'kafka'
+        'sasl_kerberos_service_name':'kafka'
     }
-    if disable_gssapi:
+    if gssapi is None:
         SASL_MECHANISMS = ('PLAIN')
     else:
         SASL_MECHANISMS = ('PLAIN', 'GSSAPI')
@@ -226,8 +228,8 @@ class BrokerConnection(object):
                 assert self.config['sasl_plain_username'] is not None, 'sasl_plain_username required for PLAIN sasl'
                 assert self.config['sasl_plain_password'] is not None, 'sasl_plain_password required for PLAIN sasl'
             if self.config['sasl_mechanism'] == 'GSSAPI':
-                assert disable_gssapi is not True, 'GSSAPI lib not available'
-                assert self.config['sasl_servicename_kafka'] is not None, 'sasl_servicename_kafka required for GSSAPI sasl'
+                assert gssapi is not None, 'GSSAPI lib not available'
+                assert self.config['sasl_kerberos_service_name'] is not None, 'sasl_servicename_kafka required for GSSAPI sasl'
 
         self.state = ConnectionStates.DISCONNECTED
         self._reset_reconnect_backoff()
@@ -508,7 +510,7 @@ class BrokerConnection(object):
     def _try_authenticate_gssapi(self, future):
 
         data = b''
-        gssname = self.config['sasl_servicename_kafka'] + '@' + self.hostname
+        gssname = self.config['sasl_kerberos_service_name'] + '@' + self.hostname
         ctx_Name      = gssapi.Name(gssname, name_type=gssapi.NameType.hostbased_service)
         ctx_CanonName = ctx_Name.canonicalize(gssapi.MechType.kerberos)
         log.debug('%s: canonical Servicename: %s', self, ctx_CanonName)
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -21,7 +21,6 @@ from kafka.protocol.commit import GroupCoordinatorResponse
 from kafka.protocol.metadata import MetadataRequest
 from kafka.protocol.types import Int32
 from kafka.version import __version__
-from mechanize import _response
 
 
 if six.PY2:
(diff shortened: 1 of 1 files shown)


kafka/conn.py
```diff
@@ -21,6 +21,7 @@
 from kafka.protocol.metadata import MetadataRequest
 from kafka.protocol.types import Int32
 from kafka.version import __version__
+from mechanize import _response
```
tvoinarovskyi: Is this used?

Harald-Berghoff: No, i have not willingly added this line

Harald-Berghoff:
the tests fails for the unknown GSSError type, which comes with the gssapi module and is added with the line `from gssapi import raw`. any hint how to fix this now?

kafka/conn.py
```diff
@@ -481,6 +505,61 @@ def _try_authenticate_plain(self, future):
...
+
+                if not output_token:
+                    if ctx_Context.complete:
+                        log.debug("%s: Security Context complete ", self)
+                    log.debug("%s: Successful GSSAPI handshake for %s", self, ctx_Context.initiator_name)
+                    break
+                try:
+                    self._sock.setblocking(True)
+                    # Send output token
+                    msg = output_token
```
dpkp: will this be bytes on python3 ? if not, should encode('utf-8')

kafka/conn.py
```diff
@@ -481,6 +505,61 @@ def _try_authenticate_plain(self, future):
...
+        log.debug('%s: canonical Servicename: %s', self, ctx_CanonName)
+        ctx_Context   = gssapi.SecurityContext(name=ctx_CanonName, usage='initiate')
+        #Exchange tokens until authentication either suceeded or failed:
+        received_token = None
+        try:
+            while not ctx_Context.complete:
+                #calculate the output token
+                try:
+                    output_token = ctx_Context.step(received_token)
+                except GSSError as e:
```
dpkp: won't this raise a NameError unless GSSError has been imported directly?

kafka/conn.py
```diff
@@ -167,9 +182,13 @@ class BrokerConnection(object):
         'metric_group_prefix': '',
         'sasl_mechanism': 'PLAIN',
         'sasl_plain_username': None,
-        'sasl_plain_password': None
+        'sasl_plain_password': None,
+        'sasl_servicename_kafka':'kafka'
```
dpkp: is this the same as `sasl.kerberos.service.name` in the java client? If so, let's reuse that config name (but with underscores)

kafka/conn.py
```diff
@@ -54,6 +54,21 @@ class SSLWantReadError(Exception):
     class SSLWantWriteError(Exception):
         pass
 
+# needed for SASL_GSSAPI authentication:
+disable_gssapi = False
+try:
+    import gssapi
+    from gssapi import raw
```
dpkp: is `raw` used? I dont see it in the changes below

kafka/conn.py
```diff
@@ -54,6 +54,21 @@ class SSLWantReadError(Exception):
     class SSLWantWriteError(Exception):
         pass
 
+# needed for SASL_GSSAPI authentication:
+disable_gssapi = False
```
dpkp: mild preference for just setting `gssapi = None` on ImportError. See kafka/codec.py

kafka/conn.py
```diff
@@ -54,6 +54,21 @@ class SSLWantReadError(Exception):
...
+try:
+    import gssapi
+    from gssapi import raw
+except ImportError:
+    #no gssapi available, will disable gssapi mechanism
+    disable_gssapi = True
+    class GSSError(Exception):
```
dpkp: looking at this again, I might do something like
```
try:
    from gssapi.raw.misc import GSSError
except ImportError:
    GSSError = None
```

kafka/conn.py
dpkp: Alternately I'd be ok with `except Exception as e` here

kafka/conn.py
Harald-Berghoff: OK for me, i'll change it

kafka/conn.py
Harald-Berghoff: not raw, but it solved the name resolution error of GSSError for me.

kafka/conn.py
Harald-Berghoff: OK

kafka/conn.py
Harald-Berghoff: Yes, this is exactly the same as sasl.kerberos.service.name in the java client, i'll change the name accordingly

kafka/conn.py
Harald-Berghoff: To avoid the NameError i added the lines below in case of ImportError
```
class GSSError(Exception):
        pass
```

kafka/conn.py
Harald-Berghoff: it is a binary token that is exchanged. I'll check the type for python3, but i think encoding will harm the authentication token.

Harald-Berghoff: return type is bytes

kafka/conn.py
```diff
@@ -167,9 +184,13 @@ class BrokerConnection(object):
         'metric_group_prefix': '',
         'sasl_mechanism': 'PLAIN',
         'sasl_plain_username': None,
-        'sasl_plain_password': None
+        'sasl_plain_password': None,
+        'sasl_kerberos_service_name':'kafka'
     }
-    SASL_MECHANISMS = ('PLAIN',)
+    if gssapi is None:
+        SASL_MECHANISMS = ('PLAIN')
```
dpkp: this needs a trailing comma for tuple-ization

kafka/conn.py
```diff
@@ -54,6 +54,23 @@ class SSLWantReadError(Exception):
...
+try:
+    import gssapi
+    #from gssapi import raw
+except ImportError:
+    #no gssapi available, will disable gssapi mechanism
+    gssapi = None
+    #class GSSError(Exception):
+    #    pass
+
+try:
```
dpkp: cool. you could also combine w/ the above import try. I think we can safely assume that either both imports work, or both will fail.

Harald-Berghoff:
I am a little lost on the issues with the travis ci test?

dpkp:
Test failure looks unrelated -- we have sporadic fixture issues on travis that I haven't been able to figure out. For now I typically just restart the failed test if it appears related to a kafka broker not starting up properly. I'm sure these could be improved, but it's out of scope for this PR.

dpkp:
This is great! Thanks for the PR. I'm sure a lot of folks will appreciate the new feature.

stys35:
Thank you

PR #1158, opened 2017-07-26
Fix batch expiry messages to state seconds

dnwe:
time.time() is seconds since epoch, and the deltas are also calculated
in seconds


jeffwidman:
Since the `kwargs` end in `_ms`, there is a chance @dpkp might prefer that the error message stay in `ms` and we multiply the reported value by 10000. 

That said, I went ahead and merged this because I'd rather we have the error message be at least correct. And I think it's easier for humans to parse values that are in seconds.

dpkp:
+1 thanks!

Issue #1155, opened 2017-07-21
UnknownTopicOrPartitionError should be retriable


tvoinarovskyi:
This ticket is a result of https://github.com/aio-libs/aiokafka/issues/176. The issue is that producer will raise UnknownTopicOrPartitionError to the user if a manual partition reassignment is in process.
How to reproduce:
* Create a multi-broker cluster, for ex. as here https://kafka.apache.org/quickstart#quickstart_multibroker. Say we have 3 nodes: `broker0`, `broker1`, `broker2`
* Set a producer that sends 1 message per second to the replicated topic
* Kill one of the nodes (say `broker2`) and launch another oner with a new config (new node `broker3`).
* Reassign partitions to the new node using instructions here https://kafka.apache.org/documentation/#basic_ops_automigrate
* You will get UnknownTopicOrPartitionError immediately on the producer.

As I understand `NotLeaderForPartition` is only raised if node actually contains replication data on it. On reassignment a previous Leader partition will migrate data to a new node and remove it's data, so the error raised will be UnknownTopicOrPartitionError instead.

Java client treats `UnknownTopicOrPartitionError` as retriable too:
https://kafka.apache.org/protocol#protocol_error_codes
https://github.com/apache/kafka/blob/0.11.0.0/clients/src/main/java/org/apache/kafka/common/errors/UnknownTopicOrPartitionException.java

dpkp:
Good catch. It looks like that exception was made retriable in 0.8.2 here
https://github.com/apache/kafka/commit/3f0b67b6ac864befccfdd4bb5dee08c0b33c3b43#diff-f556dadb08192e14b5ab3c1cb5839083  So I think ideally we would only retry on 0.8.2+ brokers.

jeffwidman:
Fixed by #1160

PR #1160, opened 2017-07-30
Make UnknownTopicOrPartitionError retriable error

tvoinarovskyi:
Checked the old client, and the error was retriable there (https://github.com/dpkp/kafka-python/blob/master/kafka/errors.py#L505), so decided there's no need to skip the retry in v0.8 brokers.

### DIFF
--- a/kafka/errors.py
+++ b/kafka/errors.py
@@ -109,6 +109,7 @@ class UnknownTopicOrPartitionError(BrokerResponseError):
     message = 'UNKNOWN_TOPIC_OR_PARTITION'
     description = ('This request is for a topic or partition that does not'
                    ' exist on this broker.')
+    retriable = True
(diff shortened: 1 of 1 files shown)


jeffwidman:
Fix #1155 

+1 on merging this, not sure if there's a reason you held off @tvoinarovskyi ?

tvoinarovskyi:
@jeffwidman https://travis-ci.org/dpkp/kafka-python/jobs/259081044 kinda want this to finish...

dpkp:
We are still seeing periodic test failures caused by kafka fixture not starting correctly. I'm not sure what the root cause is (definitely not this change). I tend to go in to travis and restart the test when I see a fixture startup failure. Tests are all passing here, so +1 for merge.

Issue #1036, opened 2017-03-17
Add support for KAFKA-4148 / KIP-79 / search offsets by timestamp in consumer


anthonyjchriste:
I've looked all around, but I can't seem to find if this library supports Kafka 10.1 or greater yet. I'm specifically trying to access offsetsForTimes as described in https://kafka.apache.org/0102/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#offsetsForTimes(java.util.Map)

This method appears to have been added in Kafka 10.1. 

I did not see the method mentioned under the KafkaConsumer for the API docs at http://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html.

I also looked through the KafkaConsumer class in the master branch, but did not see it there either. 

Is it safe to assume that 10.1 is not supported yet or am I just looking in the wrong place?

Thank you for your time and I look forward to your response.

dpkp:
Looks like you're looking for KIP-79. The new api is available and supported at the protocol level, but no one has added an equivalent to the KafkaConsumer.offsetsForTimes method. I renamed the issue to flag that this is needed.

anthonyjchriste:
That's exactly it. 

Thank you for renaming the topic. I really appreciate the work you put into this project.

ZmeiGorynych:
I was looking for the very same thing :) 

Looking at https://github.com/dpkp/kafka-python/blob/master/CHANGES.md, under March 14 I see a line saying 'Add send_list_offset_request for searching offset by timestamp (charsyam 1001)' - is that not the same task, and if it is, does that mean it's done? And if so, how do I use it? :)

Thanks a lot for any update you can give.

Drizzt1991:
@ZmeiGorynych No, that PR is about legacy code, no relation to timestamp search. As far as I know timestamp search is yet to be implemented.

anthonyjchriste:
I'm interested in attempting to implement this feature. Is there any developer documentation and are there any examples in the code that show the best (or recommended) way to implement a public API method by utilizing the protocol? Thank you for your time.

dpkp:
This python code still follows the java code fairly closely, so a reasonable first step is to just take a look at the java implementation. The java code is KAFKA-4148 / https://github.com/apache/kafka/commit/eaaa433f

anthonyjchriste:
Mahalo @dpkp! I'll fork and give it a go over the next few weeks.

Xaelias:
[...]

Issue #865, opened 2016-10-31
Use KIP-35/Version API to fetch protocol versions


tvoinarovskyi:
Currently I see some places like this in code:
```python
        if self.config['api_version'] >= (0, 10):
            version = 2
        elif self.config['api_version'] == (0, 9):
            version = 1
        else:
            version = 0
```
As of 0.10 this is not quite right, as we have a proper API for Version checking and should use that. I do understand, that Java driver still did not deliver it ([KAFKA-3600](https://issues.apache.org/jira/browse/KAFKA-3600)), but setting up a checker is not that hard to do, and it's still better, than api_version checking.

For example if instead of the check above we could write:
```python
fetch_version = self._client.best_supported_version(FetchRequest)
```
Any thoughts?

The main issue is probably that we have to do the Version check's asynchronously,  but I think I will manage to get over it. I can do a PR if we agree on the need and a bit on API.

dpkp:
Thanks -- kafka-python checks api version on startup and sets `api_version` accordingly. It is a single, cluster-wide api version. I dont think we need more complexity (as I discussed during the KIP-35 process). Why do you think a per-method map is needed?

Drizzt1991:
I do see a point there, I agree, that proper implementation of the idea of KIP-35 (at least I understand that it's more for different broker version support of sort) is complicated, but it's not even ready for that. The point is different.

Kafka will keep adding new features, requests and versions. It's evolving every month. With every new update you will add new API version checks on startup. For example in #857 you will probably add check for version `0.10.1` to check if those methods are supported or not. It will mess up the code further.

What I propose is not something that complex, we just store the results of Protocol Versions on startup, cluster-wide, same as we did with `api_version`. We even already issue the same request, just never look at the results for it. And just move the code, that selects versions for request into 1 place.

Drizzt1991:
I just kinda want the code for version selection in 1 place. Now that I think about it, it's probably OK to have it based on ``api_version`` for now, just move it in 1 place.

dpkp:
#1169 added a private map of supported versions, so I'm going to close this issue.

PR #1161, opened 2017-07-30
Added support for offsets_for_times, beginning_offsets and end_offsets APIs.

tvoinarovskyi:
Still needs to group by nodes and send in parallel.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -19,6 +19,7 @@ from kafka.protocol.api import RequestHeader
 from kafka.protocol.admin import SaslHandShakeRequest
 from kafka.protocol.commit import GroupCoordinatorResponse, OffsetFetchRequest
 from kafka.protocol.metadata import MetadataRequest
+from kafka.protocol.fetch import FetchRequest
 from kafka.protocol.types import Int32
 from kafka.version import __version__
 
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -887,7 +887,7 @@ class BrokerConnection(object):
 
     def _infer_broker_version_from_api_versions(self, api_versions):
         # The logic here is to check the list of supported request versions
-        # in descending order. As soon as we find one that works, return it
+        # in reverse order. As soon as we find one that works, return it
         test_cases = [
             # format (<broker verion>, <needed struct>)
             ((0, 11, 0), MetadataRequest[4]),
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -184,12 +184,14 @@ class Fetcher(six.Iterator):
                 self._subscriptions.seek(tp, committed)
 
     def get_offsets_by_times(self, timestamps, timeout_ms):
-        response = {}
-        for tp, timestamp in timestamps.items():
-            timestamp = int(timestamp)
-            offset, tmst = self._offset(tp, timestamp, timeout_ms=timeout_ms)
-            response[tp] = OffsetAndTimestamp(offset, tmst)
-        return response
+        offsets = self._retrieve_offsets(timestamps, timeout_ms)
+        for tp in timestamps:
+            if tp not in offsets:
+                offsets[tp] = None
+            else:
+                offset, timestamp = offsets[tp]
+                offsets[tp] = OffsetAndTimestamp(offset, timestamp)
+        return offsets
 
     def _reset_offset(self, partition):
         """Reset offsets for the given partition using the offset reset strategy.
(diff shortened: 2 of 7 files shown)


test/test_consumer_integration.py
```diff
@@ -624,3 +632,39 @@ def test_kafka_consumer_max_bytes_one_msg(self):
...
+        offsets = consumer.offsets_for_times({tp: late_time})
+        self.assertEqual(offsets[tp].offset, late_msg.offset)
+        self.assertEqual(offsets[tp].timestamp, late_time)
+
+    @kafka_versions('<0.10.1')
+    def test_kafka_consumer_offsets_for_time_old(self):
+        consumer = self.kafka_consumer()
+        tp = TopicPartition(self.topic, 0)
+
+        with self.assertRaises():
```
dpkp: Need to pass in the exception type expected here

tvoinarovskyi:
Fixes #1036

tvoinarovskyi:
@dpkp @jeffwidman  Can I have another 2 eyes here, just in case.

kafka/consumer/group.py
```diff
@@ -861,6 +861,48 @@ def metrics(self, raw=False):
             metrics[k.group][k.name] = v.value()
         return metrics
 
+    def offsets_for_times(self, timestamps):
+        """
+        Look up the offsets for the given partitions by timestamp. The returned
```
jeffwidman: >Look up the offsets for the given partitions...

This confuses me because "given" implies a list of partitions is passed in, but that's not present in the method args. If the partitions are attached to the class, then maybe use wording other than "given" to describe the partitions?

kafka/consumer/group.py
```diff
@@ -861,6 +861,48 @@ def metrics(self, raw=False):
...
+
+        This is a blocking call. The consumer does not have to be assigned the
+        partitions.
+
+        If the message format version in a partition is before 0.10.0, i.e.
+        the messages do not have timestamps, ``None`` will be returned for that
+        partition.
+
+        Note:
+            Notice that this method may block indefinitely if the partition
```
jeffwidman: "Notice that" is superfluous

test/test_consumer_integration.py
```diff
@@ -88,6 +94,12 @@ def kafka_consumer(self, **configs):
                                  **configs)
         return consumer
 
+    def kafka_producer(self, **configs):
+        brokers = '%s:%d' % (self.server.host, self.server.port)
+        producer = KafkaProducer(
+            bootstrap_servers=brokers, **configs)
+        return producer
```
jeffwidman: I wonder if this producer should just get moved to a `pytest` fixture. Not something that needs to be handled in this PR, but should probably create an issue saying we should clean it up.

kafka/conn.py
```diff
@@ -877,7 +878,9 @@ def _check_api_version_response(self, response):
         # in descending order. As soon as we find one that works, return it
         test_cases = [
             # format (<broker verion>, <needed struct>)
-            ((0, 10, 1), MetadataRequest[2])
+            ((0, 10, 1), MetadataRequest[2]),
+            ((0, 10, 2), OffsetFetchRequest[2]),
+            ((0, 11, 0), FetchRequest[5]),
```
tvoinarovskyi: Well it's not really something that was introduced in this PR, I just extended it with new checks. I did propose using ApiVersion response for min-max per-broker checks on protocol versions in  https://github.com/dpkp/kafka-python/issues/865. Here we just determine `api_version` like we used to in versions before, only for v0.10 and above we don't do additional requests, only check that max version for a protocol is supported.
I'll change the comment to be clear on ordering.

kafka/consumer/fetcher.py
```diff
@@ -198,40 +211,63 @@ def _reset_offset(self, partition):
...
+                self._client.poll(
+                    future=refresh_future, sleep=True, timeout_ms=remaining_ms)
+            else:
+                time.sleep(self.config['retry_backoff_ms'] / 1000.0)
+
+            elapsed_ms = (time.time() - start_time) * 1000
+            remaining_ms = timeout_ms - elapsed_ms
+
+        raise Errors.KafkaTimeoutError(
+            "Failed to get offsets by times in %s ms" % timeout_ms)
```
tvoinarovskyi: Here I don't mind. But the API `offsets_for_times` should probably be left as is to match `offsetsForTimes` in Java client

kafka/consumer/group.py
tvoinarovskyi: timestamps is a map `{TopicPartition: int}`. The wording was taken from https://kafka.apache.org/0110/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#offsetsForTimes(java.util.Map), I think it's OK...

test/test_consumer_integration.py
tvoinarovskyi: Sadly we can't use pytest fixtures here, as it's a `TestCase` class. We will need to refactor those classes to simple functions to support `fixtures`.

kafka/consumer/fetcher.py
jeffwidman: Completely agree to copy Java client.

test/test_consumer_integration.py
jeffwidman: Understood. You can apply pytest fixtures at the class level in pytest, but only if it's a generic class, aka designed for pytest and not unittest.

tvoinarovskyi:
> How closely does this follow the Java implementation?

Pretty much 1 to 1. I did not move all changes (``update_fetch_positions`` is done in bulk in Java Client), want to split that out into a separate PR to lower diff

tvoinarovskyi:
Added `beggining_offsets` and `end_offsets` APIs

dpkp:
Sorry... I created a merge conflict :(

jeffwidman:
🍰

buptljy:
Thanks ~  
I was trying to add the same function, but I found that it has been finished when I pulled codes from master. 
LOL

PR #1169, opened 2017-08-05
Add private map of api key -> min/max versions to BrokerConnection

dpkp:
This is a step towards #865 . For now there is no change to behavior. We might consider faking the min/max data for old broker versions. I haven't looked at the java client to see how they handle this map, but this seems like a fairly easy step to at least get some logic into the connection object.


kafka/protocol/__init__.py
```diff
@@ -6,3 +6,40 @@
     CODEC_NONE, CODEC_GZIP, CODEC_SNAPPY, ALL_CODECS,
     ATTRIBUTE_CODEC_MASK, KafkaProtocol,
 )
+
+API_KEYS = {
```
tvoinarovskyi: What is the use case for this mapping? I could only think of logging...

tvoinarovskyi:
LGTM
As for how Java handles it, a brief overview:
* It introduces a new ConnectionState - CHECKING_API_VERSIONS. It will send if after SSL handshake, or after connect of unencrypted connections. See https://github.com/apache/kafka/blob/342f34a1994caafa3cea94a2f85ffd219b8cebc5/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L715
* It checks if request is supported exactly before sending request: https://github.com/apache/kafka/blob/342f34a1994caafa3cea94a2f85ffd219b8cebc5/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L353
* Instead of preparing an exact protocol version it uses a Builder class that can choose the best one to send.

kafka/protocol/__init__.py
dpkp: TBH, I created it primarily just to document the values. But there is some potential to switch the values from strings to the versioned list of class structs we have and use it for decoding api messages. That's obviously not in this PR though.

PR #1173, opened 2017-08-10
Poll was returning duplicate messages with auto-commit disabled due to multiple rebalance trigger

88manpreet:
While using the poll feature of KafkaConsumer, we observed that it was returning duplicate messages (consistently on second poll request) if auto-commit is disabled. On debugging we observed that after sending the cluster update-request, it could take up to 2 minutes for metadata_snapshot to get updated. In this time the consumer returned all the produced messages and on next poll, it triggered rebalance again due to metadata_snapshot mismatch (empty earlier). As a result, it fetched the messages again from earliest offset, since the auto-commit was disabled.
Updating the metadata_snapshot after sending cluster update request fixed the issue.

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -245,6 +245,8 @@ class ConsumerCoordinator(BaseCoordinator):
         # keep track of the metadata used for assignment so that we can check
         # after rebalance completion whether anything has changed
         self._cluster.request_update()
+        for topic in all_subscribed_topics:
+            self._metadata_snapshot[topic] = self._cluster.partitions_for_topic(topic)
         self._assignment_snapshot = self._metadata_snapshot
 
         log.debug("Performing assignment for group %s using strategy %s"


tvoinarovskyi:
@88manpreet Thanks for the PR. Looks legit!
Let me confirm, this PR is to fix an unneeded additional group re-join, that can happen cause _metadata_snapshot was not reset after assignment. It's not a fix for duplicate messages you mentioned, but rather the problems are just related. Is that correct?

88manpreet:
@tvoinarovskyi That's right. Duplicate messages are just the effect of unneeded additional re-join.

dpkp:
I think the problem is that the metadata_snapshot is initialized to an empty dict. It should be initialized using the subscription and cluster we get passed in the constructor.

dpkp:
This is definitely an issue that needs fixing, but I'm going to close in favor of #1174 and treat this as the issue / bug report instead. Thanks so much for spotting this and reporting!

Issue #1168, opened 2017-08-05
Raise something better than AssertionError when compression libraries unavailable


dpkp:
See #1159

tvoinarovskyi:
Should be fixed by #1252. Set an explicit check in Producer's constructor.

Issue #1119, opened 2017-06-17
Support FetchRequest v4 / Message Format v2


dpkp:
> In Kafka 0.11, the structure of the 'MessageSet' and 'Message' were significantly
> changed. Not only were new fields added to support new features like exactly once
> semantics and record headers, but the recursive nature of the previous versions of
> the message format was eliminated in favor of a flat structure. A 'MessageSet' is
> now called a 'RecordBatch', which contains one or more 'Records' (and not 'Messages').
> When compression is enabled, the RecordBatch header remains uncompressed,
> but the Records are compressed together. Further, multiple fields in the 'Record'
> are varint encoded, which leads to significant space savings for larger batches. 

See https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets

tvoinarovskyi:
fml, started poking with the new format. If we start using it the introduced CRC32C will be a problem probably.
See https://issues.apache.org/jira/browse/KAFKA-1449 for more info.
tl;dr The v2 Message format uses a different CRC32 implementation.

tvoinarovskyi:
A bit of status report here. I finished a pure python prototype on top of aiokafka, currently checking compatibility and tests. On next week plan to port it to kafka-python. Thou I think this will require a C extension to be relatively fast (pure python Crc32C and varint coder can be made much faster), did not measure yet thou.

tvoinarovskyi:
I think this can be closed already

PR #1185, opened 2017-08-25
Add v2 record batch format for FetchResponse V4 support.

tvoinarovskyi:
Changed layout of record parsing to allow V2 integration with batch concept. For now I only moved the abstract implementation and wrapped LegacyBatch using existing MessageSet and Message constructs. What I want to do next is:

  * [x] Add implementation for v2 records. A bit dirty implementation is already done [here](https://github.com/aio-libs/aiokafka/blob/kafka_11_record_batch/aiokafka/record/default_records.py)
  * [x] Add builder class implementation and fix Producer to use that. 
  * [x] Refactor LegacyRecordBatch to not use Message and MessageSet constructs (not needed overhead)

@dpkp What do you think, it's quite a hard change, but I did not find a good way to abstract Batches using existing constructs. Moreover this is based on current Java client version, thou with not as many classes.


tvoinarovskyi:
Wow, that was a bit rough... Ended up with 2-3 hacks to support old versions of clients, but this should do for now. I plan to refactor LegacyRecord part to not use Message/MessageSet constructs, but for now, I need a break. In the mean time, feedback would be great.
Will finish up next week.

jeffwidman:
> Ended up with 2-3 hacks to support old versions of clients

Do you mean the old SimpleClient stacks? Or just so new KafkaClient can support old brokers?

If the former, than I started working on #1193 last night, and if I got that finished up, that would make your life simpler now and future code maintenance more straightforward...

tvoinarovskyi:
@jeffwidman Yea, I meant the SimpleClient stacks. If we remove those, I could also remove some hacks from this PR.

jeffwidman:
Then let's remove those first so that we don't have to worry about maintaining/cleaning up those hacks later. I started working on removing them last night, will try to push up a PR later tonight...

tvoinarovskyi:
@dpkp I'm satisfied with the changes, but I want your feedback on this. Also, we need to decide where will this change be merged into, is it going for 1.3.X or 2.X.X. I'm more for the second, as we can clean up the hacks for old clients.

dpkp:
Very impressive! I may have a few style suggestions that I'll write up tomorrow, but generally it looks very good. Re: merge location, I prefer just landing to master and defer choices about what to label the next release (1.4 / 2.0 / etc).

kafka/record/util.py
```diff
@@ -0,0 +1,112 @@
...
+    def _read_byte(memview, pos):
+        """ Read a byte from memoryview as an integer
+
+            Raises:
+                IndexError: if position is out of bounds
+        """
+        return ord(memview[pos])
+
+
+def encode_varint(num):
```
dpkp: Are VARINTs used elsewhere in the new Kafka protocol APIs? Perhaps these functions belong under kafka.protocol.types ?

kafka/protocol/fetch.py
```diff
@@ -1,8 +1,25 @@
...
 from .message import MessageSet
-from .types import Array, Int8, Int16, Int32, Int64, Schema, String
+from .types import Array, Int8, Int16, Int32, Int64, Schema, String, Bytes
+
+
+class Records(Bytes):
+
+    @classmethod
+    def encode(cls, buffer):
+        # For old producer, backward compat
```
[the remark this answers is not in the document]
tvoinarovskyi: Now that I think of it, let's just leave the field as Bytes in the protocol and wrap it into respectable classes in Fetcher and Old client. No hacks that way.

kafka/record/util.py
tvoinarovskyi: No, currently only the Record structure uses it. If other API's will use it we can restructure later. I want this module to be very isolated with a good Facade. I want to cythonize this part later, I already did some work on that in `aiokafka`, gives very good results so far.

tvoinarovskyi:
@dpkp Say, now that I look at it, the PR is massive. Instead of merging the whole thing let me split in in 2 parts:
 * LegacyRecords refactoring. I will just replace the current `message.py` with the `legacy_records.py` of this PR.
 * DefaultRecords implementation. Will add the V2 structures.

 That way we can first make sure we don't have a regression from the LegacyRecords classes and only after merge the v2 records. And the changeset will be more suited for review.

dpkp:
+1

On Oct 3, 2017 6:59 AM, "Taras Voinarovskyi" <notifications@github.com>
wrote:

@dpkp <https://github.com/dpkp> Say, now that I look at it, the PR is
massive. Instead of merging the whole thing let me split in in 2 parts:

   - LegacyRecords refactoring. I will just replace the current message.py
   with the legacy_records.py of this PR.
   - DefaultRecords implementation. Will add the V2 structures.

That way we can first make sure we don't have a regression from the
LegacyRecords classes and only after merge the v2 records. And the
changeset will be more suited for review.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dpkp/kafka-python/pull/1185#issuecomment-333850726>,
or mute
the thread

kafka/record/legacy_records.py
```diff
@@ -0,0 +1,413 @@
...
+        struct.pack_into(">I", msg_buffer, self.LOG_OVERHEAD, crc)
+        return crc
+
+    def _maybe_compress(self):
+        if self._compression_type:
+            data = self._buffer.getvalue()
+            if self._compression_type == self.CODEC_GZIP:
+                compressed = gzip_encode(data)
+            elif self._compression_type == self.CODEC_SNAPPY:
+                compressed = snappy_encode(data)
```
dpkp: When I test locally and snappy libraries are not available, this will raise a `NotImplementedError` and crash the producer IO thread:
```
ERROR:kafka.producer.sender:Uncaught error in kafka producer I/O thread
```
But this is only logged and producer.send() still returns a future. If you try to resolve the future with `.get()` it hangs forever. We should probably improve this so that the exception is raised to the user or at least does not crash the IO thread and is used to resolve the future. My preference is the former.

kafka/producer/kafka.py
```diff
@@ -353,7 +354,7 @@ def __init__(self, **configs):
         if self.config['compression_type'] == 'lz4':
```
dpkp: Since compression_type is a producer configuration, we could put the check for compression libraries import in `__init__` here

dpkp:
Should also note that this PR changes the error raised when compression libraries are not available -- was AssertionError, now NotImplementedError. Related to #1168

kafka/consumer/fetcher.py
```diff
@@ -873,29 +824,29 @@ def _handle_fetch_response(self, request, send_time, response):
                                   position)
                         continue
 
+                    if not records.has_next() and records.size_in_bytes() > 0:
+                        # we did not read a single message from a non-empty
+                        # buffer because that message's size is larger than
+                        # fetch size, in this case record this exception
+                        self._record_too_large_partitions[tp] = fetch_offset
```
jeffwidman: According to the Java consumer docs, this is actually now a soft-limit so that the consumer can make progress. That change is likely outside the scope of this PR, so I filed https://github.com/dpkp/kafka-python/issues/1232 to track that.

dpkp:
Running quick benchmarks suggests that KafkaProducer performance is reduced significantly when using the new record format. My laptop shows a drop from 9K records/sec to 5.5 records/sec. On the other hand, KafkaConsumer performance improved from 25K/sec to 35K/sec. This was against a local, 3-broker cluster running 0.11.0.0:
```
KAFKA_VERSION=0.11.0.0 PYTHONPATH=. ./benchmarks/producer_performance.py --brokers 3
```

When testing against 0.10.2.1 (the old message format), this PR performed basically the same as master (perhaps slightly better).

My gut says that this is ok for now; folks that care about consumer performance will benefit, folks that care about producer performance will be sad. But since we allow users to select the api version they want, they can always pin to 0.10 in their producers.

jeffwidman:
> 9K records/sec to 5.5 records/sec

Is that truly only _5.5 records/sec_ or did you mean _5.5**K**/sec_? A 50% reduction is very different than a 99% reduction in throughput...

I'm curious if the Java producer saw such a drastic throughput drop?

dpkp:
Ha, yea. 5.5K records / sec. From about 1.2MB/sec to 600KB per sec

tvoinarovskyi:
@dpkp I do expect a drop in V2 format, I did describe the problems in https://github.com/dpkp/kafka-python/issues/1119#issuecomment-324286771. For now, don't worry about those, I will do heavy benchmarking on that part before the merge. The main problem is probably the varint pure python implementation, it's a bad fit for the language.

tvoinarovskyi:
@jeffwidman JVM's Jit probably made it barely visible ) Java's fast for those type of problems, that's all. I'm positive, that PyPy will also yield quite a good number even on the current implementation.

dpkp:
I'm going to cut a release this weekend but would like to not include this PR yet. Mostly this is because memoryview does not work on python 2.6. And although we say that we do not support 2.6, it still does work -- it is just not actively tested. I think if we're going to break python2.6 affirmatively we should do it loudly / with a large version bump.

tvoinarovskyi:
@dpkp Seems like there's a bug here and the reader goes v1 on read, cause the read speed should have dropped also on v2. Probably I know where's the bug )

tvoinarovskyi:
Latest microbenchmarks for the V2 format:
```
(.kafka-python-osx)MacBook-Pro-Taras:kafka-python taras$ python benchmarks/record_batch_read.py 
.....................
batch_read_v0: Mean +- std dev: 1.33 ms +- 0.03 ms
.....................
batch_read_v1: Mean +- std dev: 1.35 ms +- 0.03 ms
.....................
batch_read_v2: Mean +- std dev: 3.52 ms +- 0.06 ms
(.kafka-python-osx)MacBook-Pro-Taras:kafka-python taras$ python benchmarks/record_batch_compose.py 
.....................
batch_append_v0: Mean +- std dev: 1.16 ms +- 0.03 ms
.....................
batch_append_v1: Mean +- std dev: 1.18 ms +- 0.03 ms
.....................
batch_append_v2: Mean +- std dev: 4.35 ms +- 0.08 ms
(.kafka-python-osx)MacBook-Pro-Taras:kafka-python taras$ 
```
Will need to lower it to at least 2 times slower...

tvoinarovskyi:
And hereby I declare myself as a magician and thus will perform magic on the latest commit.
In other words spent some hours looking at optcodes for varint functions and did some optimisations:
```
(.kafka-python-osx)MacBook-Pro-Taras:kafka-python taras$ python benchmarks/record_batch_read.py 
.....................
batch_read_v0: Mean +- std dev: 1.32 ms +- 0.03 ms
.....................
batch_read_v1: Mean +- std dev: 1.35 ms +- 0.03 ms
.....................
batch_read_v2: Mean +- std dev: 2.56 ms +- 0.04 ms
(.kafka-python-osx)MacBook-Pro-Taras:kafka-python taras$ python benchmarks/record_batch_compose.py 
.....................
batch_append_v0: Mean +- std dev: 1.20 ms +- 0.04 ms
.....................
batch_append_v1: Mean +- std dev: 1.21 ms +- 0.03 ms
.....................
batch_append_v2: Mean +- std dev: 3.15 ms +- 0.04 ms
```

jeffwidman:
Very impressive.

FYI, since you're a collaborator on the project, you can manually re-trigger the travis builds for any PR yourself within the Travis GUI, no need to push commits to force them.

tvoinarovskyi:
@jeffwidman Yea, but it bugged out on me. It did not want to restart from the interface. Had to create another job)

jeffwidman:
Is this ready to be merged? I see merge conflicts...

tvoinarovskyi:
@dpkp @jeffwidman Will go straight to the point. I want to merge this. This PR is quite big even now and dancing around it will not bring better performance. To get better it needs a C extension, which should not be part of this PR. Any objections?

kafka/consumer/fetcher.py
```diff
@@ -696,12 +703,21 @@ def _create_fetch_requests(self):
                 #       dicts retain insert order.
                 partition_data = list(partition_data.items())
                 random.shuffle(partition_data)
-                requests[node_id] = FetchRequest[version](
-                    -1,  # replica_id
-                    self.config['fetch_max_wait_ms'],
-                    self.config['fetch_min_bytes'],
-                    self.config['fetch_max_bytes'],
-                    partition_data)
+                if version == 3:
```
dpkp: I think this should be `<= 3` ?

dpkp:
I'm +1 for merge after fixing the FetchRequest api version check. Performance hit is ok with me in order to support newer features, especially headers. I'm not super excited about C-extensions. pypy JIT is quite good.

jeffwidman:
+1 from me. Agree with @dpkp, at the end of the day, core features like this trump throughput concerns. Plus in the majority of consumers that I support, most of the time is due to processing the contents of the message, not `kafka-python`.

While the speed of C-extensions are nice, I'd rather direct those who want that to `confluent-kafka-python`. A big reason I chose `kafka-python` was due to the simplicity of debugging production issues without having to worry about C-extensions. When I get called in to firefight an emergency, being able to just drop a `pdb` breakpoint and poke around is very valuable.

kafka/consumer/fetcher.py
tvoinarovskyi: no, the 0,1 and 2 versions are handled above

dpkp: yep -- i missed that!

dpkp:
Great work!

PR #1188, opened 2017-08-28
Remove a few unused imports

jameslamb:
Thank you for creating and maintaining this excellent package! I was reading through the code today just trying to get a better sense of the internals, and found a few unused imports. Please consider this PR to remove them.

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -28,6 +28,10 @@ from .metrics import AnonMeasurable
 from .metrics.stats import Avg, Count, Rate
 from .metrics.stats.rate import TimeUnit
 from .protocol.metadata import MetadataRequest
+
+# Although this looks unused, it actually monkey-patches socket.socketpair()
+# and should be left in as long as we're using socket.socketpair() in this file
+from .vendor import socketpair
 from .version import __version__
 
 if six.PY2:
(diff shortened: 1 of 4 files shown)


kafka/client_async.py
```diff
@@ -28,8 +28,6 @@
 from .metrics.stats import Avg, Count, Rate
 from .metrics.stats.rate import TimeUnit
 from .protocol.metadata import MetadataRequest
-from .protocol.produce import ProduceRequest
-from .vendor import socketpair
```
jeffwidman: This one should be left in, as it monkey-patches `socket.socketpair()`: https://github.com/dpkp/kafka-python/blob/ba7afd9bc9362055ec0bedcf53eb6f8909dc22d2/kafka/vendor/socketpair.py#L58

Rather than removing, can you add a comment right above it saying something like "Although this looks unused, it actually monkey-patches `socket.socketpair()` and should be left in as long as we're using `socket.socketpair()` in this file."

jameslamb: Got it! Yep, I can make that change

kafka/client_async.py
```diff
@@ -28,7 +28,9 @@
 from .metrics.stats import Avg, Count, Rate
 from .metrics.stats.rate import TimeUnit
 from .protocol.metadata import MetadataRequest
-from .protocol.produce import ProduceRequest
+
```
jeffwidman: Can you remove this empty line?

jameslamb: ha hey if more people nitpicked about PEP-8 stuff, the world would be a happier and safer place!

I am on it. One moment

jeffwidman:
Thanks!

PR #1196, opened 2017-08-30
Remove old SimpleClient / SimpleProducer / SimpleConsumer

### LINKED ISSUE #1193: Remove old Simple clients
I'd like to propose removing all the old SimpleClient/Consumer/Producer code. 

My rationale is that whenever I open this codebase to debug or work on something, it requires excessive mental overhead to keep in my head which files are relevant and which are for the old stack. Similarly, when I'm checking whether a test exists for a certain scenario, I have to further doublecheck whether the test is running using the old SimpleClient or the new KafkaClient. Several times I popped this library open to look at what it'd take to achieve #948 and gave up in frustration at the mess of trying to track what needs to be changed and what I can't touch for fear of breaking the old clients.

Since the old SimpleClients are deprecated and frankly not really maintained, we lose little by removing them.

The major effect would be felt by anyone running mixed-mode but can't run two version of this library, either because they've got a monolith or they've migrated consumer but not yet producer, or because they ship a single virtualenv to all microservices (my day job currently does this, and while it saved a lot of time initially, it's now causing quite a bit of pain.) At my day job we are actually not yet fully migrated to the new KafkaClients, so we'd be stuck on the old version for another couple of releases, but I still think it's worth starting to do this.

This is obviously a major breaking change, so would be part of a 2.0 release. 

@dpkp / @tvoinarovskyi what do you think?

jeffwidman:
A bunch of old code cleanup as part of bumping to version 2.

- [x] Remove old `SimpleClient` / `SimpleProducer` / `SimpleConsumer` and all associated code.
- [ ] Clearly document the removal of these so that anyone who comes along knows they need to use the 1.x branch for these old clients. --> Moved to https://github.com/dpkp/kafka-python/issues/1926
- [x] Update tests from the old Simple* stuff to use the new `KafkaClient` / `KafkaProducer` / `KafkaConsumer`

Notably this will drop our test runs from ~6-7 minutes to ~2-3 minutes per run... really convenient.

Fix #1193

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1196)


jeffwidman:
Here's initial work on this. 

Note that I did it as three commits so far:
1. Remove code (this is a WIP)
2. Remove deprecation warnings (wasn't sure whether to leave these in or not, I'm tempted to leave but thought this might be the time to make a clean slate).
3. Some TODO's that need input from @dpkp

jeffwidman:
I spent a while updating this last night, there are a few tests of the new `KafkaConsumer` / `KafkaProducer` that depend on fixtures using the old `SimpleClient` that still need updating, once that's complete this will be close to done.

dpkp:
This is definitely a big one. Re sequencing, I would love to get a final 1.4.X patch release out with any remaining bugfixes for stability. Then we can release this as either 1.5.0 or 2.0.

jeffwidman:
IMHO this should be a 2.0 release. Massive breakage everywhere. It would pair well with the new Admin client for the 2.0 release.

And 👍 on a bugfix release for the 1.4.X series... I am actually just about to start profiling some code that hit a massive performance regression after switching from `SimpleProducer` to `KafkaProducer`, so may have one more more items to land in that 1.4.x series... I think it'd be fine to sneak in the initial admin client as well into this 1.4.x release but the big emphasis of it would be the 2.0 release.

jeffwidman:
>  I am actually just about to start profiling some code that hit a massive performance regression after switching from SimpleProducer to KafkaProducer,

As an update, the root cause of the perf regression turned out to be completely unrelated to `SimpleProducer` vs `KafkaProducer`. There was only a 1-2% increase in CPU utilization when I isolated it down to just the producer change, likely due to the addition of the background thread in the `KafkaProducer`.

tvoinarovskyi:
@dpkp @jeffwidman Do I understand correctly that the biggest blocker for this to happen is our test suite being all over the place with using partly new, partly old interfaces?

dpkp:
I think that is right, yes.

jeffwidman:
Yes. There are only a handful of tests that still need fixing. And the docs / changelog also need warnings added about these breaking changes.

But the heavy-lifting of deleting code is all done. It took quite a while, so will be nice to land this, just need to push it across the finish line at some point here.

jeffwidman:
I _think_ tests should now fully pass for this, assuming Travis passes for https://github.com/dpkp/kafka-python/pull/1886.

jeffwidman:
Tests are green.

jeffwidman:
Just rebased to pickup the new updates from the `1.4.7` release.

Hoping we can get this merged in the next few weeks so that it doesn't keep hanging out there needing to be rebased... and so we don't have folks accidentally creating `unittest`-based tests.

jeffwidman:
The tests are now failing because more `unittest`-style tests crept in here: https://github.com/dpkp/kafka-python/pull/1833

So those need to be migrated for this to work. Perhaps @ulrikjohansson can handle it when he fixes up his tests.

jeffwidman:
Many thanks to @ulrikjohansson for migrating the `KafkaAdminClient` tests to `pytest`.

I just rebased this, hopefully Travis works fine.

jeffwidman:
It bothers me that we've had two folks nicely submit PR's and then I had to ask them to update tests to drop the Unittest support so that they didn't block this... so I'm planning to merge this.

I think at this point we should do a `2.x` version, as it's a big change... I realize we don't religiously follow Semvar, but I don't think `1.5` is a big enough bump.

I filed https://github.com/dpkp/kafka-python/issues/1926 to track the documentation of the change for old users.

Mostafa-Elmenbawy:
Why were the following 3 files removed by this PR?
- `kafka/partitioner/base.py`
- `kafka/partitioner/hashed.py`
- `kafka/partitioner/roundrobin.py`

ekeric13:
@jeffwidman @dpkp yeah i am trying to use a round robin partitioner and it is gone now? how come it was removed? was it buggy?

PR #1200, opened 2017-09-04
remove beginning/end offsets request version limit

buptljy:
- Remove beginning_offset and end_offset version limit because the kafka service of lower versions can get accurate offsets in these two functions.
Our discussion is in [#1199](https://github.com/dpkp/kafka-python/pull/1199)


tvoinarovskyi:
Yea, this is good!

PR #1213, opened 2017-09-12
Stop using mutable types for default arg values

### LINKED ISSUE #1212: Mutable types are used as default args
I noticed there are a bunch of places where a mutable type is used for the default value. For example, there are a bunch of `payloads=[]` in `kafka/client.py`:
https://github.com/dpkp/kafka-python/blob/370169029a9104f1963227005349f1dc6420a924/kafka/client.py#L591

Using mutable types for default args is typically a no-no unless their surprising behavior is being explicitly abused, for an explanation see: http://effbot.org/zone/default-values.htm

I haven't spelunked in the code much, but I suspect:
1. They are an oversight rather than intentional.
2. A simple fix of switching these to tuples would solve the issue.

jeffwidman:
Using mutable types for default args is typically a no-no unless their
surprising behavior is being explicitly abused, for an explanation see:
http://effbot.org/zone/default-values.htm

Note to reviewers: This is just a find-and-replace of `=[]` with `=()` to see if the tests pass. 

Possibly some of these arguments aren't actually used and could simply be removed, for example 
the test files. If you notice any of these, please comment and I will update the PR.

Fix #1212

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -398,7 +398,7 @@ class TestSimpleClient(unittest.TestCase):
     def test_correlation_rollover(self):
         with patch.object(SimpleClient, 'load_metadata_for_topics'):
             big_num = 2**31 - 3
-            client = SimpleClient(hosts=[], correlation_id=big_num)
+            client = SimpleClient(hosts=(), correlation_id=big_num)
             self.assertEqual(big_num + 1, client._next_id())
             self.assertEqual(big_num + 2, client._next_id())
             self.assertEqual(0, client._next_id())
--- a/test/test_fetcher.py
+++ b/test/test_fetcher.py
@@ -22,7 +22,7 @@ from kafka.errors import (
 
 @pytest.fixture
 def client(mocker):
-    return mocker.Mock(spec=KafkaClient(bootstrap_servers=[], api_version=(0, 9)))
+    return mocker.Mock(spec=KafkaClient(bootstrap_servers=(), api_version=(0, 9)))
 
 
 @pytest.fixture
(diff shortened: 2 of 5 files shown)


test/test_fetcher.py
```diff
@@ -22,7 +22,7 @@
 
 @pytest.fixture
 def client(mocker):
-    return mocker.Mock(spec=KafkaClient(bootstrap_servers=[], api_version=(0, 9)))
+    return mocker.Mock(spec=KafkaClient(bootstrap_servers=(), api_version=(0, 9)))
```
jeffwidman: Should this argument instead be omitted altogether?

test/test_sender.py
```diff
@@ -19,7 +19,7 @@
 
 @pytest.fixture
 def client(mocker):
-    _cli = mocker.Mock(spec=KafkaClient(bootstrap_servers=[], api_version=(0, 9)))
+    _cli = mocker.Mock(spec=KafkaClient(bootstrap_servers=(), api_version=(0, 9)))
```
jeffwidman: Should this argument instead be omitted altogether?

test/test_client.py
```diff
@@ -398,7 +398,7 @@ def test_send_produce_request_raises_when_topic_unknown(self, protocol, conn):
     def test_correlation_rollover(self):
         with patch.object(SimpleClient, 'load_metadata_for_topics'):
             big_num = 2**31 - 3
-            client = SimpleClient(hosts=[], correlation_id=big_num)
+            client = SimpleClient(hosts=(), correlation_id=big_num)
```
jeffwidman: Should this argument instead be omitted altogether?

tvoinarovskyi:
LGTM. I'd leave the arguments, as it's more explicit.

PR #1215, opened 2017-09-13
Typo fix

metalgear386:
Changed iff to if


dpkp:
iff is intentional. see https://en.wikipedia.org/wiki/iff

PR #1221, opened 2017-09-18
Remove raw traceback

### LINKED ISSUE #1220: Traceback not sent to logger, but directly to stderr
https://github.com/dpkp/kafka-python/blob/b1ae45c10d46e881492b0fd37f0919cac6d79224/kafka/conn.py#L618

At the line above, BrokerConnection prints a traceback directly to stderr, ignoring the logger. This makes it hard to filter out the traceback. Right now, for me the traceback occasionally starts generating GBs of output while at the same time a Producer stops sending information to Kafka, although why this happens I'm not sure yet, and I don't yet have a minimal example (I think it has something to do with losing a network connection though).

I think a better implementation would probably be to remove the traceback, and add stack_info=True to the previous logging call, as I think it does roughly the same thing. Or just get rid of the traceback, if it's not providing value to the project.

I should be able to provide a pull request, but I don't know if it would be preferable to get rid of the traceback entirely or not.

dbgasaway:
Fixes #1220

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -7,7 +7,6 @@ import logging
 from random import shuffle, uniform
 import socket
 import time
-import traceback
 
 from kafka.vendor import six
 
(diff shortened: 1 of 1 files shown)


dbgasaway:
It should be noted that stack_info was added in Python 3.2, but logging accepts an arbitrary **kwargs input, so I think in older versions it will just not print the stack information, but still print the error message.

dpkp:
Thanks!

Issue #1214, opened 2017-09-13
Seek method returning incorrect messages on compressed topic when using max_poll_records


88manpreet:
While using seek method of `kafka.consumer.group.seek' for a given partition, offset, we are seeing the inconsistent behavior for the messages returned with the subsequent poll method.
The issue is easily reproducible for the given topic (compacted).

Part of Workflow: 
```
from kafka.consumer.group import KafkaConsumer

topic_partition = TopicPartition(topic, 0)
consumer = KafkaConsumer(*consumer_config)
consumer.assign([topic_partition]) 
start_offset = 100 # Example value: highwatermark - 10
consumer.seek(partition=topic_partition, offset=start_offset)
messages = consumer.poll(timeout_ms=1000, max_records=1)[topic_partition]
message = messages[0]
print('Offset found:', message.offset, 'Expected offset:', start_offset)

Sample Output:
$ Offset found:80 Expected offset:100
```

Observation:
* If iterator interface is used instead of poll interface, the issue no longer exists. My guess is somewhere while polling for messages, the fetched offsets are not updated or fetched messages are not skipped. It looks like iterator method is not using fetched_records api that's why it works fine.

* At times it does give correct messages (especially when given offset is closer to highwatermark)

Please let me know if any other details are required.

88manpreet:
On some debugging I found that, the fetch_offset over [here](https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/fetcher.py#L423) is the given start_offset (100) as expected in the first iteration but in second iteration it is same as the current record.offset and henceforth it is not skipped.

dpkp:
Can you post debug logs ? Are you sure you aren't accidentally joining a consumer group and fetching the last-commit offsets?

88manpreet:
@dpkp yes we are sure. I will try to get the relevant debug logs for the above scenario. However, the fix as part of #1222 seems to work good so far for us. Please refer my comment on that issue.

dpkp:
[...]

PR #1222, opened 2017-09-18
Fix fetched_records to skip messages with offset lesser than given fetch_offset

88manpreet:
The fix is for the issue https://github.com/dpkp/kafka-python/issues/1214

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -410,25 +410,28 @@ class Fetcher(six.Iterator):
                           " %s since it is no longer fetchable", tp)
 
             elif fetch_offset == position:
-                part_records = part.take(max_records)
-                if not part_records:
-                    return 0
-                next_offset = part_records[-1].offset + 1
+                records_found = 0
+                while records_found < max_records:
+                    part_records = part.take(max_records - records_found)
+                    if not part_records:
+                        break
+                    next_offset = part_records[-1].offset + 1
 
-                log.log(0, "Returning fetched records at offset %d for assigned"
-                           " partition %s and update position to %s", position,
-                           tp, next_offset)
+                    log.log(0, "Returning fetched records at offset %d for assigned"
+                               " partition %s and update position to %s", position,
+                               tp, next_offset)
 
-                for record in part_records:
-                    # Fetched compressed messages may include additional records
-                    if record.offset < fetch_offset:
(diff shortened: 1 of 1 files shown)


88manpreet:
@dpkp I don't think I have permission to rebuild the above-failed travis job (https://travis-ci.org/dpkp/kafka-python/jobs/277078233). Can you please trigger the same for me or let me know the way I can?

dpkp:
I don't understand how this would fix the issue you've outlined.

88manpreet:
@dpkp Earlier we were just comparing current-offset (record.offset) with the desired offset (fetch_offset) once and return immediately after setting desired offset to current-offset+1 inside 'take' api (https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/fetcher.py#L943)

Note: As part of my testing I observed that this issue was only seen with topics with 'COMPRESSED' messages. On testing with UNCOMPRESSED messages, I observed that the reason it wasn't seen was that the current-offset (message.offset) was either equal to fetch_offset or 1 less than the fetch_offset.
 
Let's pick the scenario where max_records is 1 ==> part_records will always have 1 record.
Let's say current-offset (message: m): 80, Desired-offset: 90
Before fix:
----------
The message 'm' is skipped since (80 < 90) and we return, len(part_records) == 1, which is incorrect since we don't havn't reached the record with offset 90. 
Also, inside the 'take' api, we set the Desired-offset (fetch_offset) to 81, which is also incorrect.
As a result in next call to _append returns the message with offset 81.

With Fix:
--------
I keep in while loop unless I get max_records (1) message(s). As a result, it iterates from message with current-offset 80 upto the message with current-offset (90: desired-offset). Also, I return the length of relevant messages found.

Hope this helps.

dpkp:
Thanks for the detail. I see how this approach would work now. I would prefer to address the issue slightly differently, so I'm going to close this. But I'll put up a separate PR to fix the same issue.

88manpreet:
Thanks @dpkp :) We will use this fix for the time being as it blocks some of our services. Kindly link your PR to this issue and whenever it's ready, we would pick that up.
On a side-note, do you see any obvious issue or side-effect with the above fix? Let me know if any debug logs are needed.

dpkp:
The way you've done it here should work, but I haven't tested it. I landed a change to master w/ a test to verify. It will be included in the next release.

PR #1229, opened 2017-09-29
Minor doc change. Default value for sleep is True

88manpreet:
Also, I wasn't able to locate, where exactly are we sleeping when sleep is set to True.


88manpreet:
Closing this as it seems like it has been deprecated in master branch.

PR #1230, opened 2017-10-01
Use KafkaProtocol to handle serialization/deserialization of api events

dpkp:
Further work toward Issue #560 . Following from PR #1032, this moves the translation of api request/response events to and from network IO bytes into an independent class, KafkaProtocol. 
This class is intended to be reusable between different approaches to socket handling / networking. This approach should also enable simplified unit testing of protocol-level code (though actual test updates are left for future PRs).

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -14,11 +14,10 @@ from kafka.vendor import six
 import kafka.errors as Errors
 from kafka.future import Future
 from kafka.metrics.stats import Avg, Count, Max, Rate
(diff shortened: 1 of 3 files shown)


tvoinarovskyi:
Need more concentration for the review, will get it done on weekend. I like the idea, more async like interface.

tvoinarovskyi:
We can probably make the `receive_bytes` method much simpler, but it worked before, let's go with it for now.

Issue #1039, opened 2017-03-19
KAFKA-3627: KafkaConsumer.poll() fails to execute delayed tasks in poll when records are available


braedon:
My understanding of the consumer failure detection (based on the [official client docs ](http://kafka.apache.org/0102/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html) - I couldn't spot any discussion on this issue in the kafka-python docs?) was that for a consumer to stay in a group it just needed to make sure the time between each poll wasn't more than `session_timeout_ms`, and the standard way to control this was `max_poll_records`.

However, it appears that with the kafka-python implementation, a call to KafkaConsumer.poll() doesn't always cause a KafkaClient.poll(), and it is the latter that is required to stay in the group (i.e. heartbeats). Specifically, if enough data is already available to satisfy the KafkaConsumer.poll() it is returned without calling KafkaClient.poll(): https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/group.py#L600

As far as I can tell, the official client doesn't have this issue; it always calls ConsumerCoordinator.poll(): https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1033

This issue can be worked around by using `fetch_max_bytes` to control the time between KafkaClient.poll()s, rather than `max_poll_records` to control the time between KafkaConsumer.poll()s. This doesn't seem like an ideal solution, however. A lot of my consumers process very small messages - a LOT of them fit in the default 50MB `fetch_max_bytes`, so I have to set it a lot smaller, which seems like it could have adverse effects on performance?

(semi-related to https://github.com/dpkp/kafka-python/issues/948, but as I understand it even with background heartbeats you need to poll at least every `max.poll.interval.ms`, so would still be an issue)

dpkp:
Good catch - this was fixed originally in the java client via KAFKA-3627 . Since then they have switched to a background thread implementation. kafka-python has not implemented either approach yet. I've renamed this issue to mirror KAFKA-3627. See #948 re heartbeats via background thread.

dpkp:
Note that this only affects `KafkaConsumer.poll()` . The message iterator interface does not have the same problem.

braedon:
There doesn't seem to be any progress on implementing background threads in https://github.com/dpkp/kafka-python/issues/948.

It seems like it would be easier to fix this issue in the meantime?

jeffwidman:
[...]

PR #1235, opened 2017-10-05
Background tasks may not run during consumer.poll when using max_records

88manpreet:
Observation: While testing the KafkaConsumer using poll method for multiple instances and consuming messages at a low rate (~1Msg/sec) without any manual or auto commits, we observed that the rebalance was not getting triggered when a new instance was added or removed.

Steps to reproduce the issue:
a) Create consumer with auto-commit disabled, with at least 2 partitions.
b) Start 1st consumer instance
c) Consume messages at a very slow rate (~1Msg/sec), without any manual commits
d) Start another instance
No rebalance is triggered for the first instance. Both the instances are assigned from both the partitions. 

On debugging we found that on consumer.poll call the client.poll is [not](https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/group.py#L607) sent if there are previously fetched records. This lead to missing the handling of heartbeat tasks (every 3 seconds) for the running instance. As a result for any newly added or removed the instance, the consumer-group is treated as new one and assigned all partitions instead of partitioner dividing between them.

We also saw that the client.poll request in case the message_iterator method is used instead of poll.
The issue is not seen in the case of the iterator method.

The [fix is](https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/group.py#L607) to send client-poll with timeout 0 to avoid any performance impacts due to additional polls.

We tested the fix internally and tested for performance impact as minimal.

Is there a reason that client.poll is not called on every consumer.poll?

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -608,6 +608,9 @@ class KafkaConsumer(six.Iterator):
             # fetched records.
             if not partial:
                 self._fetcher.send_fetches()
+
+            # To handle any heartbeat tasks
+            self._client.poll(tiemout_ms=0)
             return records
 
         # Send any new fetches (won't resend pending fetches)


kafka/consumer/group.py
```diff
@@ -608,6 +608,9 @@ def _poll_once(self, timeout_ms, max_records):
             # fetched records.
             if not partial:
                 self._fetcher.send_fetches()
+
+            # To handle any heartbeat tasks
+            self._client.poll(tiemout_ms=0)
```
dpkp: spelling: `timeout_ms`

dpkp:
Dup of #1039. The problem with your approach is that if you call client.poll() you could trigger an autocommit before the user is given the records. This can cause message loss on failure, which is definitely not good. The java client (in 0.9.0.0) had called poll here, but it skipped delayed tasks and so it did not send heartbeats. The 0.10.0.0 java client refactored heartbeats into a background thread to avoid the issue. See #948 to track our progress on that feature.

88manpreet:
@dpkp Thanks for the explanation.

Issue #961, opened 2017-01-28
Throwing error when using invalid topic name


nikeee:
The library does not throw any exception or error when using a topic name that is not valid.
The only legal characters in a topic are:
`a-z`, `A-Z`, `0-9`,`.`, `_` and `-`
See: https://github.com/apache/kafka/blob/0.9.0/core/src/main/scala/kafka/common/Topic.scala#L25
This character constraint exists due to [this](https://issues.apache.org/jira/browse/KAFKA-495).

Maybe it would be a good idea to inform the library user that the topic is not valid. I just spent an hour of debugging because a topic contained a `/`. Therefore, the connection timed out with a `Failed to update metadata after %s secs.` error which does not seem to be related on first sight.

nikeee:
Fixed in #1238

PR #1238, opened 2017-10-05
Add method to ensure a valid topic name

nikeee:
As described in #961, the consumer does not check whether topic names are valid. It does however check if the topic name is a string.
This PR adds a check according to the current kafka source:
 https://github.com/apache/kafka/blob/39eb31feaeebfb184d98cc5d94da9148c2319d81/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L29

That means:
- A Topic name must not be empty
- It must not be `.` or `..`
- It must be shorter than 250 characters
- It must consist only of `a-z`, `A-Z`, '0-9', `.`, `-` and `_`

### DIFF
--- a/kafka/consumer/subscription_state.py
+++ b/kafka/consumer/subscription_state.py
@@ -142,13 +142,13 @@ class SubscriptionState(object):
         if not isinstance(topic, six.string_types):
             raise TypeError('All topics must be strings')
         if len(topic) == 0:
-            raise TypeError('All topics must be non-empty strings')
+            raise ValueError('All topics must be non-empty strings')
         if topic == '.' or topic == '..':
-            raise TypeError('Topic name cannot be "." or ".."')
+            raise ValueError('Topic name cannot be "." or ".."')
         if len(topic) > self._MAX_NAME_LENGTH:
-            raise TypeError('Topic name is illegal, it can\'t be longer than {0} characters, topic: "{1}"'.format(self._MAX_NAME_LENGTH, topic))
+            raise ValueError('Topic name is illegal, it can\'t be longer than {0} characters, topic: "{1}"'.format(self._MAX_NAME_LENGTH, topic))
         if not self._TOPIC_LEGAL_CHARS.match(topic):
-            raise TypeError('Topic name "{0}" is illegal, it contains a character other than ASCII alphanumerics, ".", "_" and "-"'.format(topic))
+            raise ValueError('Topic name "{0}" is illegal, it contains a character other than ASCII alphanumerics, ".", "_" and "-"'.format(topic))
 
     def change_subscription(self, topics):
         """Change the topic subscription.
--- a/kafka/consumer/subscription_state.py
+++ b/kafka/consumer/subscription_state.py
@@ -43,6 +43,10 @@ class SubscriptionState(object):
         " (2) subscribe to topics matching a regex pattern,"
         " (3) assign itself specific topic-partitions.")
 
+    # Taken from: https://github.com/apache/kafka/blob/39eb31feaeebfb184d98cc5d94da9148c2319d81/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L29
+    _MAX_NAME_LENGTH = 249
+    _TOPIC_LEGAL_CHARS = re.compile('^[a-zA-Z0-9._-]+$')
+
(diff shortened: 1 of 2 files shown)


kafka/consumer/subscription_state.py
```diff
@@ -120,6 +124,32 @@ def subscribe(self, topics=(), pattern=None, listener=None):
...
+        """
+
+        # See Kafka Source:
+        # https://github.com/apache/kafka/blob/39eb31feaeebfb184d98cc5d94da9148c2319d81/clients/src/main/java/org/apache/kafka/common/internals/Topic.java
+        if topic is None:
+            raise TypeError('All topics must not be None')
+        if not isinstance(topic, six.string_types):
+            raise TypeError('All topics must be strings')
+        if len(topic) == 0:
+            raise TypeError('All topics must be non-empty strings')
```
dpkp: I'd probably make the bottom 4 exceptions ValueError. (The first two make sense w/ TypeError)

dpkp:
Awesome!

kafka/consumer/subscription_state.py
nikeee: Done

kafka/consumer/subscription_state.py
```diff
@@ -120,6 +124,32 @@ def subscribe(self, topics=(), pattern=None, listener=None):
             raise TypeError('listener must be a ConsumerRebalanceListener')
         self.listener = listener
 
+    def _ensure_valid_topic_name(self, topic):
+        """ Ensures that the topic name is valid according to the kafka source.
+
+        Raises:
+            TypeError: if the topic is
```
jeffwidman: nit: now that you're raising `ValueError`, this docstring should also be updated. Either clarify what you raise, or just say that you raise on X condidtions, without specifying what you raise (I'm fine with either given that this is a somewhat buried docstring, other maintainers may have other opinions).

kafka/consumer/subscription_state.py
```diff
@@ -128,7 +158,11 @@ def change_subscription(self, topics):
 
         Raises:
             IllegalStateErrror: if assign_from_user has been used already
-            TypeError: if a non-str topic is given
+            TypeError: if a topic is
+                        - a non-str or
+                        - an empty string or
+                        - the topic name is '.' or '..' or
+                        - the topic name does not consist of ASCII-characters/'-'/'_'/'.'
```
jeffwidman: I would probably entirely remove this part of the docstring since the raising for invalid values now happens in `_ensure_valid_topic_name()`

kafka/consumer/subscription_state.py
nikeee: I tried to adapt to the style of the stuff of the project that I removed ([see this](https://github.com/dpkp/kafka-python/pull/1238/files/09db2238bad317717916941b68e75ff196d43134#diff-fa1fdd480849c0108501eb0ee33b6123L131)).

Now, the private method raises something. Shouldn't this be documented in the public method that uses this method (since it uses it only for that)?

kafka/consumer/subscription_state.py
nikeee: The internal method raises something. Shouldn't this be documented in the public method that uses this method (since it uses it only for that)? The checks are only encapsulated in a separate method to make it easier to reuse the checks later on.

dpkp: I would keep them in the docstring here, but update re ValueError

kafka/consumer/subscription_state.py
dpkp: I usually only add a docstring to a private method if it isn't obvious from the code what or why it is doing. The audience for private methods is primarily developers with access to the source code. But the public methods end up getting pulled out into HTML docs on kafka-python.readthedocs.org , so those are much more important.

jeffwidman: @dpkp right now the `SubscriptionState` docs are not visible on http://kafka-python.readthedocs.io/en/master/apidoc/modules.html, do you want to change that?

I'm fine with dropping the docstring from the private method, it reduces the risk of the docstring going out of sync with the code.

nikeee:
@dpkp @jeffwidman Can you take another look on this?

dpkp:
thanks!

Issue #1237, opened 2017-10-05
`KafkaConsumer.subscribe(pattern='x')` sometimes picks up topic but not partitions


jeffwidman:
There appears to be a race condition bug of some kind in `KafkaConsumer.subscribe(pattern='some pattern')`.

Normally the call works fine, the consumers picks up matching topics, assigns partitions to group members, etc. 

However, once in a blue moon I've observed that the consumer finds the matching topic, but never successfully assigns the topic partitions to the group members. Once it's in this state, it will call `poll()` for hours without returning messages because the consumer thinks it has no assigned partitions, and because the consumer's subscription already contains the topic, there's never a change that triggers a rebalance.

I'm embarrassed to say that I've spent 40+ hours over the past two weeks trying to figure this out as we hit it in production, but all I've managed to do is isolate is a semi-consistently reproducible example. Unfortunately that requires running a service that has a `KafkaConsumer` instance and has a bunch of associated docker containers, so I can't make this setup public. The wrapper service does use `gevent` which I'm not very familiar with, but I disabled all the service's other greenlets so I don't _think_ that should affect this at all.

Every time I try to isolate it down to a simple `kafka-python` script, I cannot reproduce it. But after spending hours stepping through the code, I'm reasonably certain it's a race condition in kafka-python and not the wrapper service. 

Here's what I know:

1. The issue doesn't show up the first time I run the service. If I kill the service (without calling `KafkaConsumer.close()`) and then restart it before the group coordinator evicts the consumer from the group, then I trigger the issue. If I then kill it, wait until I know the group coordinator has evicted all consumers, and then re-run it, it will work fine. Unfortunately, I have no idea if this behavior is related to the root cause, or just a trigger that makes the docker kafka container busy enough that it slows down its response times.
2. In the failure case, calling `KafkaConsumer.subscription()` returns the expected topic name, but calling `KafkaConsumer.assignment()` returns an empty set.
3. In the failure case, I can see that the cluster metadata object has both the topic and the list of partitions, so the cluster metadata is getting correctly updated, it's just not making it into the group assignments.
[...]

PR #1240, opened 2017-10-06
Explicitly check for `None` rather than False

jeffwidman:
If the group leader somehow gets in a state that it has an empty partition assignment, then `self._assignment_snapshot` will be `{}` which evaluates to `False`. So `self._subscription.mark_for_reassignment()` will never be triggered, even if `self._assignment_snapshot != self._metadata_snapshot`. 

Here is the relevant upstream Java code:
https://github.com/apache/kafka/blob/c5d26c4829583c95af7ca9e961a4d3954f8e09eb/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L197

Fixes the symptoms of #1237 although I suspect there's an additional bug in that case that triggers the condition of the the group leader getting an empty partition assignment.

I'm hesitant with this fix because I still don't fully understand the race condition that results in the group leader simultaneously having:
1. a valid topic subscription
2. an empty partition assignment
3.`subscription.needs_partition_assignment = False`.

Related: #1237 / #1241 / #1242

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -190,7 +190,7 @@ class ConsumerCoordinator(BaseCoordinator):
         # if we were the assignor, then we need to make sure that there have
         # been no metadata updates since the rebalance begin. Otherwise, we
         # won't rebalance again until the next metadata change
-        if self._assignment_snapshot and self._assignment_snapshot != self._metadata_snapshot:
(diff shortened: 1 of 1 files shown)


jeffwidman:
Discussed in person with @dpkp and he agreed it should be merged.

PR #1243, opened 2017-10-06
Expand metrics docs

jeffwidman:
I've never taken the time to experiment with them because they weren't documented, so decided to fix that.

There were a number of distinct changes, so I tried to keep them contained within separate commits so we can easily drop any that aren't wanted... probably easiest to review as distinct commits as well.

For example, I'm not sure if the Sphinx warning syntax will show up properly on RTD or not, as I'm not sure how to build the docs locally.

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -8,6 +8,8 @@ import threading
 import time
 import weakref
 
+from ..vendor import six
+
 from .. import errors as Errors
 from ..client_async import KafkaClient, selectors
 from ..metrics import MetricConfig, Metrics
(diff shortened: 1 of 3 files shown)


kafka/producer/kafka.py
```diff
@@ -8,6 +8,8 @@
 import time
 import weakref
 
+from ..vendor import six
```
jeffwidman: I noticed the import style was inconsistent between `kafka/producer/kafka.py` and `kafka/consumer/group.py`, not sure why...?

dpkp: I think I must have started kafka/consumer/group.py before deciding to use relative imports. I like this style better.

tvoinarovskyi:
Great!

PR #1244, opened 2017-10-08
Update to 0.11.0.1 for travis test fixture

dpkp:
Also updated test.fixtures to use 0.11.0.1 by default if KAFKA_VERSION is not set. Previously the default was 0.8.0


tvoinarovskyi:
LGTM

PR #1247, opened 2017-10-08
Add a Makefile

tvoinarovskyi:
With this, users which are less familiar with the project can easily build, generate docs and run tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,52 @@
+# Some simple testing tasks (sorry, UNIX only).
+
+FLAGS=
+KAFKA_VERSION=0.11.0.0
+SCALA_VERSION=2.11
+
+setup:
+	pip install -r requirements-dev.txt
+	pip install -Ue .
+
+# Test and produce coverage using tox. This is the same as is run on Travis
+test36:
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
@dpkp @jeffwidman Any commands you want to add?

Makefile
```diff
@@ -0,0 +1,52 @@
+# Some simple testing tasks (sorry, UNIX only).
+
+FLAGS=
+KAFKA_VERSION=0.11.0.0
```
dpkp: can update to 0.11.0.1 now

Makefile
```diff
@@ -0,0 +1,52 @@
...
+SCALA_VERSION=2.11
+
+setup:
+	pip install -r requirements-dev.txt
```
dpkp: probably want a virtualenv here, dont we? or perhaps --user ?

dpkp:
might be nice to also include a command to fetch / install server artifacts via build_integration.sh

Makefile
tvoinarovskyi: I would prefer to assume the virtualenv is already installed and activated. There are a lot of different ways to do that, like pyvenv, virtualenv, virtualenvwrapper and so on.

tvoinarovskyi: I can change the name to something like `setup-local`

PR #1258, opened 2017-10-16
Move callback processing from BrokerConnection to KafkaClient

dpkp:
This is a small refactor to how response callbacks are processed. It is designed to separate the callback chain processing from the network processing code. This will allow us to use a threading lock around client networking that can be released before processing callbacks to reduce the risk of deadlock. This approach is derived from KAFKA-3888 / KIP-62 (background heartbeat thread).

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -175,7 +175,8 @@ class SimpleClient(object):
 
             # Block
             while not future.is_done:
-                conn.recv()
+                for r, f in conn.recv():
+                    f.success(r)
 
             if future.failed():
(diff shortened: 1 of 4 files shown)


dpkp:
This PR handles the success case, but network failures in BrokerConnection.send / .recv still handle future.failure() inline. I am going to think a bit about how best to handle this so that both callbacks and errbacks can be handled in isolation (avoiding locking issues w/ background thread)

tvoinarovskyi:
Lgtm

tvoinarovskyi:
Merging this to lessen the burden of reviewing #1266

Issue #1130, opened 2017-06-22
example.py script does not cleanly shutdown producer / consumer instances


dakoller:
Hi,

I took the example script with both producer & consumer.

The producer code looks like this:

----
class Producer(threading.Thread):
    daemon = True

    def run(self):
        producer = KafkaProducer(value_serializer=lambda v: json.dumps(v).encode('utf-8'))

        while True:
            producer.send('test',  {'msg': 'latency bad'})
            time.sleep(random.random() * 4)
            producer.send('test',  {'msg': 'showing better'})
            time.sleep(random.random()*4)

--
The output is:

Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/Users/daniel/projects/kafka_message_processing/example.py", line 31, in run
    producer.send('test',  {'msg': 'latency bad'})
  File "/Library/Python/2.7/site-packages/kafka/producer/kafka.py", line 516, in send
    self.config['max_block_ms'])
  File "/Library/Python/2.7/site-packages/kafka/producer/record_accumulator.py", line 213, in append
    assert not self._closed, 'RecordAccumulator is closed'
AssertionError: RecordAccumulator is closed

...do you have any idea?

harsh29:
+1

dpkp:
Can you try enabling debug logs? It seems like something is closing your producer outside of the main thread. Possibly via interpreter shutdown? Do you only see this error when killing the script either via Ctrl-C or kill signal ?

alirezastack:
I see the same state when trying to use the sample code on `OS X`. I do not kill the script using CTRL+C or any kill signal.

dpkp:
Your debug logs suggest that you are closing the KafkaProducer while another thread is still trying to use the producer to send messages:
```
2017-09-17 22:16:58,075.75.5391120911:kafka.producer.kafka:140735279222784:INFO:5356:Closing the Kafka producer with 0 secs timeout.
```

alirezastack:
I'm new to Kafka, and used the installation at my first step, so I don't have any other script in between that would try to close the producer connection.

dpkp:
Ah -- sorry, I see that you are just running `example.py` . The threading code there does not handle clean shutdown, so the error is expected. It would be great if someone wanted to submit a PR that improves the example to avoid scary looking errors on close.

TheAtomicOption:
[...]

Issue #1148, opened 2017-07-14
Add `wakeup` to Consumer as in Java Client


tvoinarovskyi:
Java client has [wakeup()](https://kafka.apache.org/0101/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#wakeup()) functions that can be used to break out of a `poll()` and can be called from another thread. It's needed for some unique cases like the process pool processing example, where we have 1 consumer and a pool of workers.

dpkp:
kafka-python supports `wakeup` via a socketpair in KafkaClient. This allows threads sharing a KafkaClient instance that is blocked waiting for IO to signal that wakeup is desired for other processing.

There is also a related, but slightly different, "wakeup" feature in the java KafkaConsumer class. This allows threads that share a KafkaConsumer instance to trigger a WakeupException to be raised from the next consumer.poll() call. As far as I can tell this is used primarily to signal a looping consumer that it should shutdown. But I think the same could be done with simple external concurrency primitives, like a shared threading.Event() perhaps.

You filed this issue wrt the second (KafkaConsumer.wakeup()), correct?

dpkp:
The current behavior of `consumer.poll(timeout_ms=1000)` is to only block a maximum of `retry_backoff_ms` in the case that there are no fetchable partitions, e.g., all subscribed partitions are paused. So I don't think this feature is as useful in current releases.

But I could imagine a future change that allows an option to instead block the full timeout but support a `wakeup()` from a different thread using a `WakeupNotifier`. There are a lot of reasons why all partitions may be unfetchable though and we'd want to make sure those are covered somehow to avoid an unexpected deadlock.

PR #1268, opened 2017-10-18
Added controlled thread shutdown to example.py

TheAtomicOption:
Added better control over threads/processes, making them shutdown cleanly before the program terminates. Switched from the iterator to poll() style to do it.

Should solve this issue: https://github.com/dpkp/kafka-python/issues/1130

### DIFF
--- a/example.py
+++ b/example.py
@@ -6,29 +6,44 @@ from kafka import KafkaConsumer, KafkaProducer
 
 
 class Producer(threading.Thread):
-    daemon = True
+    def __init__(self):
+        threading.Thread.__init__(self)
+        self.stop_event = threading.Event()
+        
+    def stop(self):
+        self.stop_event.set()
 
     def run(self):
         producer = KafkaProducer(bootstrap_servers='localhost:9092')
 
-        while True:
+        while not self.stop_event.is_set():
             producer.send('my-topic', b"test")
             producer.send('my-topic', b"\xc2Hola, mundo!")
             time.sleep(1)
 
+        producer.close()
 
 class Consumer(multiprocessing.Process):
-    daemon = True
-
+    def __init__(self):
+        multiprocessing.Process.__init__(self)
+        self.stop_event = multiprocessing.Event()
+        
+    def stop(self):
+        self.stop_event.set()
+        
     def run(self):
         consumer = KafkaConsumer(bootstrap_servers='localhost:9092',
                                  auto_offset_reset='earliest')
         consumer.subscribe(['my-topic'])
 
-        for message in consumer:
-            print (message)
+        while not self.stop_event.is_set():
+            message = consumer.poll(timeout_ms=0, max_records=1)
+            if len(message) > 0:
+                print (message.values()[0])
 
-
+        consumer.close()
+        
+        
 def main():
     tasks = [
         Producer(),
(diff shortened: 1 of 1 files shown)


example.py
```diff
@@ -6,29 +6,44 @@
...
     def run(self):
         consumer = KafkaConsumer(bootstrap_servers='localhost:9092',
                                  auto_offset_reset='earliest')
         consumer.subscribe(['my-topic'])
 
-        for message in consumer:
-            print (message)
-
+        while not self.stop_event.is_set():
+            message = consumer.poll(timeout_ms=0, max_records=1)
```
dpkp: I admit that I am not a huge fan of the `poll` interface. We support it because that is how the java client works, but I think using the iterator is much more pythonic. Perhaps something like:
```
consumer = KafkaConsumer(...., consumer_timeout_ms=1000)
while True:
    for message in consumer:
        print(message)
        if self.stop_event.is_set():
            break
````

?

tvoinarovskyi: It will not work. Needs the wakeup method

TheAtomicOption: I agree the iterator looks more pythonic, and believe me I've tried to use it instead. But it simply isn't controllable as currently implemented. The main problem being that it blocks indefinitely if it's called when there won't be any new messages for it. 

consumer_timeout_ms is not a great solution either because it just statically sets the lifetime of the KafkaConsumer object. The consumer just closes on its own after that length of time even if you're actively pulling messages through it with the iterator. You can set consumer_timeout_ms for longer, but then you're just back to the same problem where the iterator call is blocking forever until that timer runs out. 

poll() is the only way I've found to put clean consumer closure under programmer control. It lets you use the same consumer as long as you like without reinitializing it, and then lets you shut it down very quickly as well. My code, which poll()'s for maxrecords=1 and loops if no new messages, is the closest you can currently get to a controllable iterator that allows the programmer's code to end the request loop at will.

If I've somehow missed a command to shut down the iterator's internal request loop please let me know.

TheAtomicOption:
When I ran it, I got a "Closing connection" line in log, and then another "Closing connection" later when calling consumer.close(). 

Even if I'm misreading that, it still requires you to decide ahead of time when you're going to shut down the request loop. If you want to iterate indefinitely until an event signals a stop, you'll need to somehow have another thread come in and set consumer_timeout_ms=1 because the threat the iterator is in is blocked by the iterator.

TheAtomicOption:
One solution might be to add an iterator_max_time_ms  parameter such that if iterator_max_time_ms is set, the iterator will return None (or maybe a zero length list as an empty "consumer record" if None would make the python loop end?) instead of polling internally forever. 

If the default value was infinite, this wouldn't break any existing implementations. And it would allow the iterator loop to execute every so often so that signals like my stop_event could be checked.

If this sounds good, I might be able to figure out how to make the pull request. What do you think?

TheAtomicOption:
I played around with your code above and got it to work. I was wrong about consumer_timeout_ms closing the consumer. Thanks for your patience with me.

I still think it's not _ideal_ to need to nest 2 loops and to check the stop_event twice, but it does work. I haven't seen the iterator used with an outer while loop like this anywhere else. I think showing this structure in more of the examples would prevent a lot of headaches. 

I'll edit the pull request real quick...

dpkp:
No problemo. I agree that it is not ideal. The way that that java consumer handles this is there is a consumer.wakeup() method that you can call from a different thread and it will cause the consumer to raise a WakeupException from a long-blocking poll(). We could implement something similar here, but it hasn't been a high priority. See #1148 for tracking that feature.

dpkp:
Thanks for the PR!

PR #1269, opened 2017-10-19
Explicitly check for None rather than falsey

jeffwidman:
Be pedantic about checking for identity rather than equality to avoid issues like #1237 / 411bc08f214b7afc36f11bde2047096c06467088

Noticed while reviewing #1266

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -539,7 +539,7 @@ class KafkaClient(object):
                         task_future.success(result)
 
             # If we got a future that is already done, don't block in _poll
-            if future and future.is_done:
+            if future is not None and future.is_done:
                 timeout = 0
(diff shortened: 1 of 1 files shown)


tvoinarovskyi:
And it's faster anyway)

kafka/client_async.py
```diff
@@ -681,8 +681,6 @@ def least_loaded_node(self):
         elif 'bootstrap' in self._conns:
             return 'bootstrap'
 
-        return None
```
dpkp: Slight preference for keeping this as an explicit return. Otherwise we still return None, it is just implicit when the function completes.

jeffwidman: No problem, reverted.

PR #1272, opened 2017-10-20
Check that timestamp_ms value is expected type

### LINKED ISSUE #1271: timestamp_ms that isn't integer throws confusing Kafka Producer struct.error during send
I am having a strange problem.   I have a django endpoint that I am trying to get to post of Kafka topic. 
It accepts JSON body,  byte encodes that json message then send to kafka via kafka producer.

However the message never makes it to Kafka, it fails silently.   After some investigation the producer is eating the error which is  struct.error.   The error does not had a message.    This happens during message encoding.

Any help is appreciated.

Using Python version 3.6.0  (Anaconda package) 
OS:  Windows 7 64bit
Kafka:  0.11.0.1  (bitnami ovm)

Curl command: 
curl -i -X POST -H "Content-Type: application/json" -b "{\"id\": \"1234\",  \"message\": \"test message 1\"}" http://localhost:8000/api/v1/log/

` 
# Python Code

KAFKA_BROKERLIST = '192.168.95.168:9092'
KAFKA_TOPIC = 'test'
KAFKA_ACKS = 1

def post(self, request):
        try:
            message = json.loads(request.body, encoding='utf-8')
            msg = None
            if 'id' not in message:  # we need to add an id to the message before we add to kafka
                id = uuid.uuid4()
                if isinstance(message , list): message.insert(0,id)
                elif isinstance(message, dict): message['id']=id
                elif isinstance(message, str): message = "id: {} {}".format(id,message)
                else: #we could not add id to the message for specified format
                    raise Exception("Id was not supplied, and could not generate id for message body.  message format ={}".format(type(message)))
            value = bytes(json.dumps(message),'utf-8') 
            self.producer.send(topic=settings.KAFKA_TOPIC, value=value, timestamp_ms=time())
            self.producer.flush()
            # except KafkaError as e:
            #     raise Exception(str(e))
        except Exception as e:
            if hasattr(e, 'message'): raise Exception(e.message)
            else: raise Exception(str(e))
        return ""

#Producer wrap
class loggingProducerView(View):

    _producer = None

    @property
    def producer(self):
        if self._producer is None:
            self._producer = KafkaProducer(
                client_id ='logging_service',
                bootstrap_servers = KAFKA_BROKERLIST,
                acks = KAFKA_ACKS
            )
        return self._producer

`

Debug output:
Starting the Kafka producer
Added sensor with name connections-closed
Added sensor with name connections-created
Added sensor with name select-time
Added sensor with name io-time
Bootstrapping cluster metadata from [('192.168.95.168', 9092, <AddressFamily.AF_INET: 2>)]
Attempting to bootstrap via node at 192.168.95.168:9092
Added sensor with name bytes-sent-received
Added sensor with name bytes-sent
Added sensor with name bytes-received
Added sensor with name request-latency
Added sensor with name node-bootstrap.bytes-sent
Added sensor with name node-bootstrap.bytes-received
Added sensor with name node-bootstrap.latency
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: creating new socket
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: setting socket option (6, 1, 1)
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: connecting to 192.168.95.168:9092
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: established TCP connection
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: Connection complete.
Node bootstrap connected
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092> Request 1: MetadataRequest_v0(topics=[])
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092> Response 1: MetadataResponse_v0(brokers=[(node_id=0, host='192.168.95.168', port=9092)], topics=[(error_code=0, topic='__consumer_offsets', partitions=[(error_code=0, partition=23, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=41, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=32, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=8, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=17, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=44, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=35, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=26, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=11, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=29, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=38, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=47, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=20, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=2, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=5, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=14, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=46, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=49, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=40, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=4, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=13, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=22, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=31, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=16, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=7, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=43, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=25, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=34, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=10, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=37, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=1, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=19, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=28, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=45, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=36, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=27, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=9, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=18, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=21, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=48, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=12, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=3, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=30, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=39, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=15, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=42, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=24, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=33, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=6, leader=0, replicas=[0], isr=[0]), (error_code=0, partition=0, leader=0, replicas=[0], isr=[0])]), (error_code=0, topic='logging', partitions=[(error_code=0, partition=0, leader=0, replicas=[0], isr=[0])])])
Updated cluster metadata to ClusterMetadata(brokers: 1, topics: 2, groups: 0)
Bootstrap succeeded: found 1 brokers and 2 topics.
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: Closing connection. 
<BrokerConnection node_id=bootstrap host=192.168.95.168/192.168.95.168 port=9092>: reconnect backoff 0.048303844089174645 after 1 failures
Initiating connection to node 0 at 192.168.95.168:9092
Added sensor with name node-0.bytes-sent
Added sensor with name node-0.bytes-received
Added sensor with name node-0.latency
<BrokerConnection node_id=0 host=192.168.95.168/192.168.95.168 port=9092>: creating new socket
<BrokerConnection node_id=0 host=192.168.95.168/192.168.95.168 port=9092>: setting socket option (6, 1, 1)
<BrokerConnection node_id=0 host=192.168.95.168/192.168.95.168 port=9092>: connecting to 192.168.95.168:9092
Node 0 connected
Broker version identifed as 0.11.0
Set configuration api_version=(0, 11, 0) to skip auto check_version requests on startup
Added sensor with name bufferpool-wait-time
Added sensor with name batch-size
Added sensor with name compression-rate
Added sensor with name queue-time
Added sensor with name produce-throttle-time
Added sensor with name records-per-request
Added sensor with name bytes
Added sensor with name record-retries
Added sensor with name errors
Added sensor with name record-size-max
Starting Kafka producer I/O thread.
Kafka producer started
Sending (key=None value=b'{"id": "1234", "message": "test message 1"}') to TopicPartition(topic='logging', partition=0)
Sending metadata request MetadataRequest_v1(topics=['logging']) to node 0
<BrokerConnection node_id=0 host=192.168.95.168/192.168.95.168 port=9092> Request 3: MetadataRequest_v1(topics=['logging'])
<BrokerConnection node_id=0 host=192.168.95.168/192.168.95.168 port=9092> Response 3: MetadataResponse_v1(brokers=[(node_id=0, host='192.168.95.168', port=9092, rack=None)], controller_id=0, topics=[(error_code=0, topic='logging', is_internal=False, partitions=[(error_code=0, partition=0, leader=0, replicas=[0], isr=[0])])])
Updated cluster metadata to ClusterMetadata(brokers: 1, topics: 1, groups: 0)
Allocating a new 16384 byte message buffer for TopicPartition(topic='logging', partition=0)
Exception occurred during message send: <class 'struct.error'>

jeffwidman:
I used an assert rather than an exception because if someone runs python with the `-o` flag where asserts are stripped out, they probably don't want to pay the penalty of this check, and if it fails they will get an error, it'll just be difficult to figure it out (as seen in #1271).

Fix #1271

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -530,6 +530,8 @@ class KafkaProducer(object):
         assert value is not None or self.config['api_version'] >= (0, 8, 1), (
             'Null messages require kafka >= 0.8.1')
         assert not (value is None and key is None), 'Need at least one: key or value'
+        assert timestamp_ms is None or isinstance(timestamp_ms, int), (
+            'timestamp_ms must be either None or an integer') 
         key_bytes = value_bytes = None
         try:
             # first make sure the metadata for the topic is


dpkp:
What about attempting to convert to int?
```
if timestamp_ms is not None and not isinstance(timestamp_ms, int):
    timestamp_ms = int(timestamp_ms)
```

jeffwidman:
While it's certainly more pythonic, I'm not too excited about ducktyping this:

1. Dates/times are complex enough that I expect there's a number of semi-valid inputs where the conversion won't blow up, but also won't coerce the value to what the user expects, resulting in silently corrupted timestamps. For example, what if someone thinks `timestamp_ms=True` causes the producer to append the current timestamp?
2. A common use case for message timestamps in Kafka is to support message re-processing when you realized the last deploy pushed a bug in your code. Just fix the bug and re-process all messages since the last deploy. However, if you aren't watching consumer lag based on timestamps, you may not realize that your timestamps are being coerced incorrectly until you're in the middle of firefighting. So it feels like database backups... if there's a chance of failure, I'd rather blow up early and let the human decide what to do.
3. Are there any negative performance impacts for the broker's time-based index if multiple producers are sending to a topic, and one of them is sending incorrect timestamps? For example, the `timestamp_ms=True` which coerces to `1`, does that mean the broker is constantly switching between appending the valid entries to the end of the index and then going back to the beginning for the erroneous entries? Perhaps this is fine, I'm not very familiar with the time-based offsets functionality/implementation, as we are still ~6 months away from turning them on at my day job

tvoinarovskyi:
Ehmm, did you try to reproduce the issue in master? It should raise a proper exception from LegacyRecordBatchBuilder.

tvoinarovskyi:
On master we already see a proper exception:
```
  File "test.py", line 21, in <module>
    fut.get()
  File "/Users/taras/workspace/kafka-python/kafka/producer/future.py", line 63, in get
    raise self.exception # pylint: disable-msg=raising-bad-type
TypeError: 00:00:00
```

jeffwidman:
> Ehmm, did you try to reproduce the issue in master?

Nope, I'm afraid I didn't. I should have, but I skipped it because I was in a hurry and didn't think any recent commits had affected this code path. Thanks for noticing this.

jeffwidman:
> Ehmm, did you try to reproduce the issue in master?

Nope, I'm afraid I didn't (looking sheepish). I should have, but I skipped it because I was in a hurry and didn't think any recent commits had affected this code path. I do remember checking the code and not seeing any checks for this, so I just assumed nothing was ever raised. Thanks for noticing this.

However, after looking at this again, the exception isn't actually raised directly to the user when they call `producer.send()`, it's only if they check the result of the future.  I'm not sure how reasonable it is to expect users to do this, I suspect many just fire and forget. Versus we do a couple of other checks of the inputs right up front and fast fail, see: https://github.com/dpkp/kafka-python/blob/4dbf34abce9b4addbb304520e2f692fbaef60ae5/kafka/producer/kafka.py#L530-L532

I guess I'm fine closing though, the asserts we do upfront are only for the most egregious mistakes, and I'm loathe to expand the number of checks there as this is a high-throughput code path.

tvoinarovskyi:
@jeffwidman Filled a followup for the problem of not raising the error to the user: #1274

PR #1273, opened 2017-10-21
Fix timestamp not passed to RecordMetadata

### LINKED ISSUE #1270: Degrade after LegacyRecords merge
As I moved the generation of a timestamp into LegacyRecordBatchBuilder the timestamp returned by future may now be None, as it was not generated yet.
This is more of a reminder for myself to fix it.

tvoinarovskyi:
Tried to fix #1270 and found, that it did not work properly in the first place. Was a bug in FutureRecordMetadata, that ignored timestamp=-1 being returned from the broker.

Also minor fixes include:
* Avoid getting current timestamp if magic==0 (broker <= 0.9), as it's not needed anyway. Will return -1 for this case in ```RecordMetadata```.
* Fixed timestamp error message if it's of improper type.
* Added more tests for LegacyBatch
* Added integration test for correct RecordMetadata fields, including `checksum`

### DIFF
--- a/kafka/record/legacy_records.py
+++ b/kafka/record/legacy_records.py
@@ -110,6 +110,8 @@ class LegacyRecordBase(object):
     LOG_APPEND_TIME = 1
     CREATE_TIME = 0
 
+    NO_TIMESTAMP = -1
+
 
 class LegacyRecordBatch(ABCRecordBatch, LegacyRecordBase):
 
@@ -333,10 +335,14 @@ class LegacyRecordBatchBuilder(ABCRecordBatchBuilder, LegacyRecordBase):
         # Check types
         if type(offset) != int:
             raise TypeError(offset)
-        if timestamp is None:
+        if self._magic == 0:
+            timestamp = self.NO_TIMESTAMP
+        elif timestamp is None:
             timestamp = int(time.time() * 1000)
         elif type(timestamp) != int:
-            raise TypeError(timestamp)
+            raise TypeError(
+                "`timestamp` should be int, but {} provided".format(
+                    type(timestamp)))
         if not (key is None or
                 isinstance(key, (bytes, bytearray, memoryview))):
             raise TypeError(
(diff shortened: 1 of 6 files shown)


test/test_producer.py
```diff
@@ -78,3 +78,54 @@ def test_kafka_producer_gc_cleanup():
...
+
+@pytest.mark.skipif(not version(), reason="No KAFKA_VERSION set")
+@pytest.mark.parametrize("compression", [None, 'gzip', 'snappy', 'lz4'])
+def test_kafka_producer_proper_record_metadata(kafka_broker, compression):
+    connect_str = ':'.join([kafka_broker.host, str(kafka_broker.port)])
+    producer = KafkaProducer(bootstrap_servers=connect_str,
+                             retries=5,
+                             max_block_ms=10000,
+                             compression_type=compression,
+                             value_serializer=str.encode)
```
dpkp: You're sending bytes below so no serializer needed here.

tvoinarovskyi: Copy-paste)

kafka/record/legacy_records.py
```diff
@@ -484,3 +490,37 @@ def estimate_size_in_bytes(cls, magic, compression_type, key, value):
                 cls.record_size(magic, key, value)
             )
         return cls.LOG_OVERHEAD + cls.record_size(magic, key, value)
+
+
+class LegacyRecordMetadata(object):
```
jeffwidman: I'm sorry, I am clearly missing something... What is the purpose of this class? It doesn't seem to do anything other than enforcing read-only on the attributes. Is there something else it does that I'm not realizing?

tvoinarovskyi: No, it's more or less as you said. It's just a class to store results of a function. I find this a bit more clean than `namedtuple` when it comes to optimization. This is a hot part.

It's a more convenient way of returning several values from a function. Returning a tuple will require you to parse it every time, but returning either `None` or an instance will only require a None check. See https://github.com/dpkp/kafka-python/pull/1273/files#diff-261f56705197d1ab9e6fca9249856556 and https://github.com/dpkp/kafka-python/pull/1273/files#diff-d98d7e54a13e15511c44d90e78f105e9R368.

PR #1281, opened 2017-10-25
Use socket timeout when blocking on send/recv

### LINKED ISSUE #1246: Set a timeout on `send` in connection
Spotted while reviewing #1230. The code for sending bytes into the network currently looks like this:
```python
            # In the future we might manage an internal write buffer
            # and send bytes asynchronously. For now, just block
            # sending each request payload
            self._sock.setblocking(True)
            total_sent = 0
            while total_sent < len(data):
                sent_bytes = self._sock.send(data[total_sent:])
                total_sent += sent_bytes
            assert total_sent == len(data)
            if self._sensors:
                self._sensors.bytes_sent.record(total_sent)
            self._sock.setblocking(False)
```

The problem here is, that if 1 of the nodes goes down only on write (will stop reading bytes on the other end) this will lead to a stop in application code, as the sending Thread will suspend indefinitely. This is not too uncommon in Kubernetics or other virtual environments, that do not perfectly handle socket proxying.
To avoid the issue we have to make the call asynchronous or at least set the timeout based on `request_timeout_ms` configuration. For example:
```python
            # In the future we might manage an internal write buffer
            # and send bytes asynchronously. For now, just block
            # sending each request payload
            self._sock.settimeout(self.config['request_timeout_ms'] / 1000.0)
            total_sent = 0
            while total_sent < len(data):
                sent_bytes = self._sock.send(data[total_sent:])
                total_sent += sent_bytes
            assert total_sent == len(data)
            if self._sensors:
                self._sensors.bytes_sent.record(total_sent)
            self._sock.setblocking(False)
```

dpkp:
Fixes #1246

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import collections
 import copy
@@ -491,7 +491,7 @@ class BrokerConnection(object):
                     self.config['sasl_mechanism']))
 
     def _send_bytes_blocking(self, data):
-        self._sock.setblocking(True)
+        self._sock.settimeout(self.config['request_timeout_ms'] / 1000)
         total_sent = 0
         try:
             while total_sent < len(data):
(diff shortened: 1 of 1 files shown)


kafka/conn.py
```diff
@@ -501,10 +501,10 @@ def _send_bytes_blocking(self, data):
                 raise ConnectionError('Buffer overrun during socket send')
             return total_sent
         finally:
-            self._sock.setblocking(False)
+            self._sock.settimeout(0.0)
```
tvoinarovskyi: You can leave it as setblocking(False). It's the same as per documentation.

dpkp: Yes, they are the same. But I decided to use settimeout here for symmetry with the original call.

PR #1293, opened 2017-11-13
Integration tests migration to pytest framework

asdaraujo:
In preparation for the changes I'll need to make to the integration tests to cover the GSSAPI authentication changes in PR #1283, I had to make a few changes to the existing test cases, which were written using unittest.TestCase.
    
After some discussions with @dpkp and @jeffwidman I decided to first move those test cases to pytest, which is the preferred test framework for the project.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -404,7 +404,7 @@ class SimpleClient(object):
         return [responses[tp] for tp in original_ordering]
 
     def __repr__(self):
-        return '<KafkaClient client_id=%s>' % (self.client_id)
+        return '<SimpleClient client_id=%s>' % (self.client_id)
 
     def _raise_on_response_error(self, resp):
 
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -545,6 +545,8 @@ class KafkaClient(object):
             timeout_ms = 100
         elif timeout_ms is None:
             timeout_ms = self.config['request_timeout_ms']
+        elif not isinstance(timeout_ms, (int, float)):
+            raise RuntimeError('Invalid type for timeout: %s' % type(timeout_ms))
 
         # Loop for futures, break after first loop if None
         responses = []
--- a/pylint.rc
+++ b/pylint.rc
@@ -1,5 +1,6 @@
 [TYPECHECK]
 ignored-classes=SyncManager,_socketobject
+generated-members=py.*
 
 [MESSAGES CONTROL]
 disable=E1129
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,38 +1,117 @@
 from __future__ import absolute_import
 
-import os
+import inspect
 
 import pytest
+from decorator import decorate
 
 from test.fixtures import KafkaFixture, ZookeeperFixture
-
+from test.testutil import kafka_version, random_string
 
 @pytest.fixture(scope="module")
 def version():
-    if 'KAFKA_VERSION' not in os.environ:
-        return ()
-    return tuple(map(int, os.environ['KAFKA_VERSION'].split('.')))
-
+    """Return the Kafka version set in the OS environment"""
+    return kafka_version()
 
 @pytest.fixture(scope="module")
-def zookeeper(version, request):
-    assert version
-    zk = ZookeeperFixture.instance()
-    yield zk
-    zk.close()
-
+def zookeeper():
+    """Return a Zookeeper fixture"""
+    zk_instance = ZookeeperFixture.instance()
+    yield zk_instance
+    zk_instance.close()
 
 @pytest.fixture(scope="module")
-def kafka_broker(version, zookeeper, request):
-    assert version
-    k = KafkaFixture.instance(0, zookeeper.host, zookeeper.port,
-                              partitions=4)
-    yield k
-    k.close()
+def kafka_broker(kafka_broker_factory):
+    """Return a Kafka broker fixture"""
+    return kafka_broker_factory()[0]
 
+@pytest.fixture(scope="module")
+def kafka_broker_factory(version, zookeeper):
+    """Return a Kafka broker fixture factory"""
+    assert version, 'KAFKA_VERSION must be specified to run integration tests'
+
+    _brokers = []
+    def factory(**broker_params):
+        params = {} if broker_params is None else broker_params.copy()
+        params.setdefault('partitions', 4)
+        num_brokers = params.pop('num_brokers', 1)
+        brokers = tuple(KafkaFixture.instance(x, zookeeper, **params)
+                        for x in range(num_brokers))
+        _brokers.extend(brokers)
+        return brokers
+
+    yield factory
+
+    for broker in _brokers:
+        broker.close()
+
+@pytest.fixture
+def simple_client(kafka_broker, request, topic):
+    """Return a SimpleClient fixture"""
+    client = kafka_broker.get_simple_client(client_id='%s_client' % (request.node.name,))
+    client.ensure_topic_exists(topic)
+    yield client
+    client.close()
+
+@pytest.fixture
+def kafka_client(kafka_broker, request):
+    """Return a KafkaClient fixture"""
+    (client,) = kafka_broker.get_clients(cnt=1, client_id='%s_client' % (request.node.name,))
+    yield client
+    client.close()
+
+@pytest.fixture
+def kafka_consumer(kafka_consumer_factory):
+    """Return a KafkaConsumer fixture"""
+    return kafka_consumer_factory()
+
+@pytest.fixture
+def kafka_consumer_factory(kafka_broker, topic, request):
+    """Return a KafkaConsumer factory fixture"""
+    _consumer = [None]
+
+    def factory(**kafka_consumer_params):
+        params = {} if kafka_consumer_params is None else kafka_consumer_params.copy()
+        params.setdefault('client_id', 'consumer_%s' % (request.node.name,))
+        _consumer[0] = next(kafka_broker.get_consumers(cnt=1, topics=[topic], **params))
+        return _consumer[0]
+
+    yield factory
+
+    if _consumer[0]:
+        _consumer[0].close()
+
+@pytest.fixture
+def kafka_producer(kafka_producer_factory):
+    """Return a KafkaProducer fixture"""
+    yield kafka_producer_factory()
+
+@pytest.fixture
+def kafka_producer_factory(kafka_broker, request):
+    """Return a KafkaProduce factory fixture"""
+    _producer = [None]
+
+    def factory(**kafka_producer_params):
+        params = {} if kafka_producer_params is None else kafka_producer_params.copy()
+        params.setdefault('client_id', 'producer_%s' % (request.node.name,))
+        _producer[0] = next(kafka_broker.get_producers(cnt=1, **params))
+        return _producer[0]
+
+    yield factory
+
+    if _producer[0]:
+        _producer[0].close()
+
+@pytest.fixture
+def topic(kafka_broker, request):
+    """Return a topic fixture"""
+    topic_name = '%s_%s' % (request.node.name, random_string(10))
+    kafka_broker.create_topics([topic_name])
+    return topic_name
 
 @pytest.fixture
 def conn(mocker):
+    """Return a connection mocker fixture"""
     from kafka.conn import ConnectionStates
     from kafka.future import Future
     from kafka.protocol.metadata import MetadataResponse
--- a/test/fixtures.py
+++ b/test/fixtures.py
@@ -4,28 +4,54 @@ import atexit
 import logging
 import os
 import os.path
-import shutil
+import random
+import socket
+import string
 import subprocess
-import tempfile
 import time
 import uuid
 
-from six.moves import urllib
+import py
+from six.moves import urllib, xrange
 from six.moves.urllib.parse import urlparse  # pylint: disable=E0611,F0401
 
+from kafka import errors, KafkaConsumer, KafkaProducer, SimpleClient
+from kafka.client_async import KafkaClient
+from kafka.protocol.admin import CreateTopicsRequest
+from kafka.protocol.metadata import MetadataRequest
 from test.service import ExternalService, SpawnedService
-from test.testutil import get_open_port
-
 
 log = logging.getLogger(__name__)
 
+def random_string(length):
+    return "".join(random.choice(string.ascii_letters) for i in xrange(length))
+
+def version_str_to_list(version_str):
+    return tuple(map(int, version_str.split('.'))) # e.g., (0, 8, 1, 1)
+
+def version():
+    if 'KAFKA_VERSION' not in os.environ:
+        return ()
+    return version_str_to_list(os.environ['KAFKA_VERSION'])
+
+def get_open_port():
+    sock = socket.socket()
+    sock.bind(("", 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    return port
 
 class Fixture(object):
     kafka_version = os.environ.get('KAFKA_VERSION', '0.11.0.2')
     scala_version = os.environ.get("SCALA_VERSION", '2.8.0')
-    project_root = os.environ.get('PROJECT_ROOT', os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
-    kafka_root = os.environ.get("KAFKA_ROOT", os.path.join(project_root, 'servers', kafka_version, "kafka-bin"))
+    project_root = os.environ.get('PROJECT_ROOT',
+                                  os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
+    kafka_root = os.environ.get("KAFKA_ROOT",
+                                os.path.join(project_root, 'servers', kafka_version, "kafka-bin"))
     ivy_root = os.environ.get('IVY_ROOT', os.path.expanduser("~/.ivy2/cache"))
+
+    def __init__(self):
+        self.child = None
 
     @classmethod
     def download_official_distribution(cls,
@@ -71,31 +97,34 @@ class Fixture(object):
     @classmethod
     def kafka_run_class_args(cls, *args):
         result = [os.path.join(cls.kafka_root, 'bin', 'kafka-run-class.sh')]
-        result.extend(args)
+        result.extend([str(arg) for arg in args])
         return result
 
     def kafka_run_class_env(self):
         env = os.environ.copy()
-        env['KAFKA_LOG4J_OPTS'] = "-Dlog4j.configuration=file:%s" % self.test_resource("log4j.properties")
+        env['KAFKA_LOG4J_OPTS'] = "-Dlog4j.configuration=file:%s" % \
+                                  self.test_resource("log4j.properties")
         return env
 
     @classmethod
     def render_template(cls, source_file, target_file, binding):
-        log.info('Rendering %s from template %s', target_file, source_file)
+        log.info('Rendering %s from template %s', target_file.strpath, source_file)
         with open(source_file, "r") as handle:
             template = handle.read()
             assert len(template) > 0, 'Empty template %s' % source_file
-        with open(target_file, "w") as handle:
+        with open(target_file.strpath, "w") as handle:
             handle.write(template.format(**binding))
             handle.flush()
             os.fsync(handle)
 
         # fsync directory for durability
         # https://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/
-        dirfd = os.open(os.path.dirname(target_file), os.O_DIRECTORY)
+        dirfd = os.open(os.path.dirname(target_file.strpath), os.O_DIRECTORY)
         os.fsync(dirfd)
         os.close(dirfd)
 
+    def dump_logs(self):
+        self.child.dump_logs()
 
 class ZookeeperFixture(Fixture):
     @classmethod
@@ -111,32 +140,36 @@ class ZookeeperFixture(Fixture):
         fixture.open()
         return fixture
 
-    def __init__(self, host, port):
+    def __init__(self, host, port, tmp_dir=None):
+        super(ZookeeperFixture, self).__init__()
         self.host = host
         self.port = port
 
-        self.tmp_dir = None
-        self.child = None
+        self.tmp_dir = tmp_dir
 
     def kafka_run_class_env(self):
         env = super(ZookeeperFixture, self).kafka_run_class_env()
-        env['LOG_DIR'] = os.path.join(self.tmp_dir, 'logs')
+        env['LOG_DIR'] = self.tmp_dir.join('logs').strpath
         return env
 
     def out(self, message):
         log.info("*** Zookeeper [%s:%s]: %s", self.host, self.port or '(auto)', message)
 
     def open(self):
-        self.tmp_dir = tempfile.mkdtemp()
+        if self.tmp_dir is None:
+            self.tmp_dir = py.path.local.mkdtemp() #pylint: disable=no-member
+        self.tmp_dir.ensure(dir=True)
+
         self.out("Running local instance...")
         log.info("  host    = %s", self.host)
         log.info("  port    = %s", self.port or '(auto)')
-        log.info("  tmp_dir = %s", self.tmp_dir)
+        log.info("  tmp_dir = %s", self.tmp_dir.strpath)
 
         # Configure Zookeeper child process
         template = self.test_resource("zookeeper.properties")
-        properties = os.path.join(self.tmp_dir, "zookeeper.properties")
-        args = self.kafka_run_class_args("org.apache.zookeeper.server.quorum.QuorumPeerMain", properties)
+        properties = self.tmp_dir.join("zookeeper.properties")
+        args = self.kafka_run_class_args("org.apache.zookeeper.server.quorum.QuorumPeerMain",
+                                         properties.strpath)
         env = self.kafka_run_class_env()
 
         # Party!
@@ -174,7 +207,7 @@ class ZookeeperFixture(Fixture):
         self.child.stop()
         self.child = None
         self.out("Done!")
-        shutil.rmtree(self.tmp_dir)
+        self.tmp_dir.remove()
 
     def __del__(self):
         self.close()
@@ -182,9 +215,11 @@ class ZookeeperFixture(Fixture):
 
 class KafkaFixture(Fixture):
     @classmethod
-    def instance(cls, broker_id, zk_host, zk_port, zk_chroot=None,
+    def instance(cls, broker_id, zookeeper, zk_chroot=None,
                  host=None, port=None,
-                 transport='PLAINTEXT', replicas=1, partitions=2):
+                 transport='PLAINTEXT', replicas=1, partitions=2,
+                 sasl_mechanism='PLAIN', auto_create_topic=True, tmp_dir=None):
+
         if zk_chroot is None:
             zk_chroot = "kafka-python_" + str(uuid.uuid4()).replace("-", "_")
         if "KAFKA_URI" in os.environ:
@@ -195,19 +230,29 @@ class KafkaFixture(Fixture):
             if host is None:
                 host = "localhost"
             fixture = KafkaFixture(host, port, broker_id,
-                                   zk_host, zk_port, zk_chroot,
+                                   zookeeper, zk_chroot,
                                    transport=transport,
-                                   replicas=replicas, partitions=partitions)
+                                   replicas=replicas, partitions=partitions,
+                                   sasl_mechanism=sasl_mechanism,
+                                   auto_create_topic=auto_create_topic,
+                                   tmp_dir=tmp_dir)
+
             fixture.open()
         return fixture
 
-    def __init__(self, host, port, broker_id, zk_host, zk_port, zk_chroot,
-                 replicas=1, partitions=2, transport='PLAINTEXT'):
+    def __init__(self, host, port, broker_id, zookeeper, zk_chroot,
+                 replicas=1, partitions=2, transport='PLAINTEXT',
+                 sasl_mechanism='PLAIN', auto_create_topic=True,
+                 tmp_dir=None):
+        super(KafkaFixture, self).__init__()
+
         self.host = host
         self.port = port
 
         self.broker_id = broker_id
+        self.auto_create_topic = auto_create_topic
         self.transport = transport.upper()
+        self.sasl_mechanism = sasl_mechanism.upper()
         self.ssl_dir = self.test_resource('ssl')
 
         # TODO: checking for port connection would be better than scanning logs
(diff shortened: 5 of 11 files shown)


asdaraujo:
@dpkp Is ipv6 enabled on these build VMs? I'm getting the following error in the travis build:
```
kafka.common.KafkaException: Socket server failed to bind to 0:0:0:0:0:0:0:1:52941: Protocol family unavailable.
```

I haven't changed anything related to the Kafka broker bind address, so I'm wondering why of these errors... All my local tests have passed and bound to ipv6 addresses without any problems.

kafka/protocol/types.py
```diff
@@ -100,7 +100,7 @@ def decode(cls, data):
 
     @classmethod
     def repr(cls, value):
-        return repr(value[:100] + b'...' if len(value) > 100 else b'')
+        return repr('<None>' if value is None else value[:100] + b'...' if len(value) > 100 else b'')
```
jeffwidman: You don't need to cast `None` to a string, `repr()` will do it automatically and it'll be less confusing (at least to me) to see the naturally expected string representation of `None`. Also, if you prepend the None check to the existing length check, then that will get evaluated first, preventing the `len()` from throwing the exception.

I also think there's a bug here and the final `b''` should actually be `value` as the whole point is to print the entire thing if it's less than 100 characters. But IIRC @tvoinarovskyi wrote this and he normally knows what he's doing, so perhaps there's a reason for this?

So all that put together probably looks like:

    value[:100] + b'...' if value is not None and len(value) > 100 else value

kafka/client.py
```diff
@@ -404,7 +404,7 @@ def _send_consumer_aware_request(self, group, payloads, encoder_fn, decoder_fn):
         return [responses[tp] for tp in original_ordering]
 
     def __repr__(self):
-        return '<KafkaClient client_id=%s>' % (self.client_id)
+        return '<SimpleClient client_id=%s>' % (self.client_id)
```
jeffwidman: 👍

kafka/client_async.py
```diff
@@ -526,6 +526,8 @@ def poll(self, timeout_ms=None, future=None, delayed_tasks=True):
             timeout_ms = 100
         elif timeout_ms is None:
             timeout_ms = self.config['request_timeout_ms']
+        elif not isinstance(timeout_ms, int) and not isinstance(timeout_ms, float):
```
jeffwidman: `isinstance()` supports checking against multiple types in one call:

    isinstance(timeout_ms, (int, float))

kafka/client_async.py
```diff
@@ -526,6 +526,8 @@ def poll(self, timeout_ms=None, future=None, delayed_tasks=True):
...
+            raise RuntimeError('Invalid type for timeout: %s' % (type(timeout_ms),))
```
jeffwidman: The argument to `%` doesn't have to be a tuple if it's just a single argument:

    >>> 'Invalid type for timeout: %s' % (type(5),)
    "Invalid type for timeout: <type 'int'>"
    >>> 'Invalid type for timeout: %s' % type(5)
    "Invalid type for timeout: <type 'int'>"

I'd argue that in this case the latter form is more readable.

servers/trunk/resources/log4j.properties
```diff
@@ -19,6 +19,6 @@ log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
 
-log4j.logger.kafka=DEBUG, stdout
+log4j.logger.kafka=INFO, stdout
```
jeffwidman: Why?

test/conftest.py
```diff
@@ -1,35 +1,166 @@
...
 def version():
-    if 'KAFKA_VERSION' not in os.environ:
-        return ()
-    return tuple(map(int, os.environ['KAFKA_VERSION'].split('.')))
-
+    return kafka_version()
 
 @pytest.fixture(scope="module")
-def zookeeper(version, request):
-    assert version
```
jeffwidman: I wonder why these were in here in the first place...?

test/conftest.py
```diff
@@ -1,35 +1,166 @@
...
+
+def set_params(**params):
+    global _params
+
+    def real_set_params(func, *args, **kwargs):
+        _params[func.__name__] = params
+        ret = decorate(func, _set_params)
+        return ret
+
+    return real_set_params
```
jeffwidman: I don't understand what is the purpose of these setters?

test/fixtures.py
```diff
@@ -164,7 +186,6 @@ def open(self):
         else:
             raise Exception('Failed to start Zookeeper before max_timeout')
         self.out("Done!")
-        atexit.register(self.close)
```
jeffwidman: Why is it safe to remove this?

test/fixtures.py
```diff
@@ -239,20 +269,30 @@ def __init__(self, host, port, broker_id, zk_host, zk_port, zk_chroot,
...
+        return '%s:%d' % (self.host, self.port)
+
     def kafka_run_class_env(self):
         env = super(KafkaFixture, self).kafka_run_class_env()
         env['LOG_DIR'] = os.path.join(self.tmp_dir, 'logs')
         return env
 
     def out(self, message):
-        log.info("*** Kafka [%s:%s]: %s", self.host, self.port or '(auto)', message)
+        log.info("*** Kafka [%s:%s]: %s", self.host, self.port or '(auto)', message.decode('utf-8') if isinstance(message, bytes) else message)
```
jeffwidman: @dpkp do we want to know the actual bytes here for debugging purposes?

test/fixtures.py
```diff
@@ -239,20 +269,30 @@ def __init__(self, host, port, broker_id, zk_host, zk_port, zk_chroot,
...
         if self.running:
             self.out("Instance already running")
             return
 
-        self.tmp_dir = tempfile.mkdtemp()
+        # Create directories
+        if self.tmp_dir is None:
+            self.tmp_dir = tempfile.mkdtemp()
+            os.mkdir(os.path.join(self.tmp_dir, "logs"))
+            os.mkdir(os.path.join(self.tmp_dir, "data"))
```
jeffwidman: IIRC pytest has a built-in tmp dir fixture that can be used... is that relevant here at all? I'm not sure how these directories are actually used...

test/fixtures.py
```diff
@@ -317,21 +353,153 @@ def open(self):
...
             raise Exception('Failed to start KafkaInstance before max_timeout')
+
+        (self._client,) = self.get_clients(1, '_internal_client')
+
         self.out("Done!")
         self.running = True
-        atexit.register(self.close)
 
-    def __del__(self):
-        self.close()
```
jeffwidman: Why is it safe to remove this?

test/test_client_async.py
```diff
@@ -7,14 +7,14 @@
     # vendored backport module
     import kafka.vendor.selectors34 as selectors
 
+import pytest
```
jeffwidman: this should not be moved per pep8, `pytest` should be in 3p library imports separate from stdlib imports

test/test_client_async.py
```diff
@@ -393,13 +393,34 @@ def test_maybe_refresh_metadata_cant_send(mocker, client):
...
+    sleep_secs = 0.5
+    task_result = 'task_succeeded'
+    def task():
+        return task_result
+    start_time = time.time()
+    future = client.schedule(task, start_time + task_delay_secs)
+    while not future.is_done:
+        time.sleep(sleep_secs)
+        client.poll(timeout_ms=100, delayed_tasks=True)
+        assert time.time() < start_time + tolerance_secs
```
jeffwidman: Can we add parens to make the order of operations more obvious?

    time.time() < (start_time + tolerance_secs)

test/test_client_async.py
```diff
@@ -393,13 +393,34 @@ def test_maybe_refresh_metadata_cant_send(mocker, client):
...
+    assert future.exception is None
+
+
+def test_unschedule(mocker, client):
+    def task():
+        return 'anything'
+    future = client.schedule(task, time.time() + 1)
+    client.unschedule(task)
+    assert future.is_done
+    assert future.value == None
```
jeffwidman: should be `is None`

test/test_integration_consumer.py
```diff
@@ -0,0 +1,233 @@
+import logging
+import pytest
+import six
+import time
+
+from six.moves import xrange
```
jeffwidman: per pep8, the above imports should be split into a section for stdlib and a section for 3p imports

kafka/client_async.py
asdaraujo: Nice. I'll change it.

kafka/client_async.py
asdaraujo: Yes, I know. But I usually prefer to make it explicit as a tuple to avoid ambiguity with the case where the argument is a tuple. It's not the case here, since we know that `type(5)` will never be a tuple. But I usually try to use the same style for consistency.

I'm happy to change this since the style for this project is the latter form.

kafka/protocol/types.py
asdaraujo: Agreed on all counts. I'll wait for @tvoinarovskyi response for the final change in regards to the `b''` case.

servers/trunk/resources/log4j.properties
asdaraujo: My bad. This was just for some tests and slipped into the commit. I'll revert this.

test/conftest.py
asdaraujo: I can't tell. But having the check on the Kafka fixture seems enough to me... ZK shouldn't depend on that for anything.

test/conftest.py
asdaraujo: When I started coding and testing the fixtures I noticed that the fixtures needed different test methods may need different configurations. For example, some tests need a consumer with `auto_offset_reset=earliest` while others need `auto_offset_reset=latest`.

One way to enable that flexibility would be to let each test to explicitly instantiate the fixtures, without using pytest injection. Another way, would be to provide the pytest fixtures the parameters for each instantiation.

These setters were my way to implement that. Using those, the tests could be coded like this:
```
@set_params(simple_consumer={'auto_offset_reset':'earliest'})
def test_consumer_with_earliest(simple_consumer):
    ...

@set_params(simple_consumer={'auto_offset_reset':'latest'})
def test_consumer_with_latest(simple_consumer):
    ...
```
The parameters set this way are exposed to the fixture method as an injected parameter:
```
@pytest.fixture
def simple_consumer(...some_params..., simple_consumer_params):
    ....
```
Which I though to be simpler and easier to code than handle explicit instantiation and destruction on each test method.

This works only for fixtures with `function` scope.

Thoughts?

test/fixtures.py
asdaraujo: Because I'm handling the destruction at the fixture method:

```
+def zookeeper():
     zk = ZookeeperFixture.instance()
     yield zk
     zk.close() # <--
```

test/fixtures.py
asdaraujo: mmm... I hadn't thought of that. It makes sense to have the raw bytes, rather than a nicely formatted message. I'll revert this

test/fixtures.py
asdaraujo: My idea is that these should all be instantiated by pytest fixtures, and I'm handling the destruction in the fixture methods, as explained above.

test/test_client_async.py
asdaraujo: Noted. I'll change

test/test_client_async.py
asdaraujo: Sure

test/test_client_async.py
asdaraujo: (y)

test/test_integration_consumer.py
asdaraujo: Will change... And will re-read PEP8!

asdaraujo:
@jeffwidman Thanks for the review.
Agreed on the smaller chunks for the PRs. I'll break this down and update the PR with the first changes.

asdaraujo:
@jeffwidman Would you know anything about the "protocol not available" errors? I see that other builds in Travis were also affected by the same issue. Even documentation changes fail with that error, so I assume it's an infrastructure problem.

test/test_client_async.py
jeffwidman: btw, you can directly put the thumbsup on the comment (via the smiley face/emoticon dropdown) rather than leaving a separate comment... reduces spam for watchers of the project when all you want to do is say "i agree/support this"

test/fixtures.py
asdaraujo: Yes, it's relevant. I had a look at the pytest tmpdir fixture and will change the code to use it instead

asdaraujo:
@jeffwidman I've reduced the scope of this PR and per your suggestion and will build up from here after this one is approved and merged.

The CI builds, though, keep failing, but is again due to those "protocol not available" errors. I tested the changes on my test box for all the Python and Kafka versions and it works well.

pylint.rc
```diff
@@ -1,5 +1,6 @@
 [TYPECHECK]
 ignored-classes=SyncManager,_socketobject
+generated-members=py.*
```
jeffwidman: What does this do?

test/fixtures.py
```diff
@@ -317,21 +336,55 @@ def open(self):
...
+        log.info("  host       = %s", self.host)
+        log.info("  port       = %s", self.port or '(auto)')
+        log.info("  transport  = %s", self.transport)
+        log.info("  broker_id  = %s", self.broker_id)
+        log.info("  zk_host    = %s", self.zk_host)
+        log.info("  zk_port    = %s", self.zk_port)
+        log.info("  zk_chroot  = %s", self.zk_chroot)
+        log.info("  replicas   = %s", self.replicas)
+        log.info("  partitions = %s", self.partitions)
+        log.info("  tmp_dir    = %s", self.tmp_dir.strpath)
```
jeffwidman: I wonder whether it's better to have these as separate log entries or as a single log entry that just has newlines? I'm not actually sure.

test/fixtures.py
```diff
@@ -317,21 +336,55 @@ def open(self):
             tries += 1
         else:
             raise Exception('Failed to start KafkaInstance before max_timeout')
-        self.out("Done!")
         self.running = True
+        self.out("Kafka started!")
+
+    def open(self):
+        if self.running:
+            self.out("Instance already running")
```
jeffwidman: Is there a reason to use `stdout` here rather than `log.info()`? 

I noticed throughout this code that some messages are going to the logs and others going to stdout which seems confusing to me, but perhaps there's a reason or unix convention for this... @dpkp your input?

test/fixtures.py
```diff
@@ -71,28 +90,29 @@ def test_resource(cls, filename):
     @classmethod
     def kafka_run_class_args(cls, *args):
         result = [os.path.join(cls.kafka_root, 'bin', 'kafka-run-class.sh')]
-        result.extend(args)
+        result.extend([str(x) for x in args])
```
jeffwidman: at first glance, this change makes sense to me, but I'm curious why not casting to strings wasn't causing problems before?

test/fixtures.py
```diff
@@ -4,27 +4,46 @@
...
+import random
+import socket
+import string
 import subprocess
-import tempfile
 import time
 import uuid
 
-from six.moves import urllib
+import py
```
jeffwidman: What benefit do we get from this new dependency `py`?

test/fixtures.py
```diff
@@ -4,27 +4,46 @@
...
+from six.moves import urllib, xrange
```
jeffwidman: Personally, I prefer using the python3 syntax and then let `six` fill in the python 2 equivalent. So use `range` rather than `xrange` here--they are functionally the same.

test/conftest.py
jeffwidman: So I agree with the design goal, but can't you skip the setters in favor of `pytest`'s built-in `@pytest.mark.parametrize()` decorator? 

It provides the flexibility to both dynamically assign the fixture params for a test, and also to run a single test multiple times using different input permutations.

Am I missing some additional functionality that the custom setters provide?

jeffwidman:
Also, I'm not sure what the protocol error is about... @dpkp is this a known issue?

pylint.rc
asdaraujo: The `py` module seems to have some dynamically generated members and `pylint` complains about not finding the ones used by the module:
```
Module 'py' has no 'path' member (no-member)
```

This is the same issue reported [here](https://github.com/PyCQA/pylint/issues/640). Adding this option to pylint just silences this error.

This will be temporary anyway, as I explain below.

test/fixtures.py
asdaraujo: I had a look at replacing `tempdir` with the pytest's `tmpdir` fixture as you suggested and I liked the idea.

The pytest's `tmpdir` fixture is an instance of the `py.path.local` class. Once I finish converting all the test cases to `pytest`, the `zookeeper` and `kafka` fixtures will automatically get a `tmpdir` fixture from `pytest` and will pass it to the `ZookeeperFixture` and `KafkaFixture` constructors, respectively.

Before the conversion is complete, though, the constructors` will still not being passed `tmpdir` as a parameter and will have to instantiate it internally.

Hence, this import is only needed temporarily to allow the `unittest` code to continue working while we're going through this conversion process. Once everything is converted to `pytest`, we can remove this since the `tmpdir` will always be provided by the `pytest` fixtures.

test/fixtures.py
asdaraujo: In the changes that I'll be submitting next I changed the `create_topic` method to be more "robust" and don't rely on the SimpleClient API or on `auto_create_topic=true` to be able to create a topic.

The method tries to use the API, if possible (Kafka > 0.10.1), to create a topic but, if not, falls back to using `kafka.admin.TopicCommand` to ensure the topic is created successfully.

This command takes numeric command line arguments for the `--partitions` and `--replication-factor` options and this is what cause me to hit an exception at this point of the code.

I could've just passed them as strings to `kafka_run_class_args` instead, but I found this was more elegant and could help avoid further issues later.

test/fixtures.py
asdaraujo: I noticed that as well, but I don't know why. I just kept it the way it had been.
Happy to change it, if needed.

test/fixtures.py
asdaraujo: I personally prefer the way it is. It makes the code more readable, IMO.

test/conftest.py
asdaraujo: That's a good question. At first glance the `@set_params` and `@pytest.mark.parametrize` decorators seem equivalent but they work in very different ways. I had a look at the latter before I implemented the former and realized it didn't suit the purpose I had in mind.

`@pytest.mark.parametrize` is used to inject different values into a test method's parameter. Each value will trigger a new execution of the test. For example, consider the following test method:
```
@pytest.mark.parametrize('auto_offset_reset', ['earliest', 'latest'])
@pytest.mark.parametrize('enable_auto_commit', [True, False])
def test_consumer(kafka_consumer, auto_offset_reset, enable_auto_commit)
    ...
```
This method will run 4 times, each time with the following values for the `(auto_offset_reset, enable_auto_commit)` variables: `('earliest', True)`, `('earliest', False)`, `('latest', True)`, `('latest', False)`.
The `kafka_consumer` fixture, however, **will be instantiated the same way every time**. The parameters above are injected into the test method, but not the fixture itself.

The `@set_params` decorator does not cause multiple test executions. Instead, it just passes parameters for the instantiation of the specified fixture. In the example below, the test is run only once and the parameters specified in the `@set_params` call are injected into the `kafka_consumer` fixture, not the test method:
```
@set_params(kafka_consumer={'auto_offset_reset':'earliest', 'enable_auto_commit':True})
def test_consumer(kafka_consumer)
    ... kafka_consumer is intantiated with the parameters specified above^^
```

asdaraujo: @jeffwidman You mentioned that `@pytest.mark.parametrize` "provides the flexibility to ... dynamically assign the fixture params for a test".

I can see how that would work if the fixture is explicitly instantiated within the test method. I don't know how the parameters could be passed to the fixture instantiation if `pytest` is injecting it into the test method. Is there a way to do that? I may be missing something here.

asdaraujo: I just had another look at the pytest documentation. I see that the fixtures can be parametrized through the `@pytest.fixture()` decorator. This still doesn't seem to suit the purpose here, since different tests may need different fixture configurations.

I think there's a way to do what we need using the `pytest_generate_tests` method. I'll have a look into this and will post back here.

dpkp:
Overall it looks good to me. Tests pass locally after fixing merge conflicts and removing the tox.ini py version pin.

test/fixtures.py
jeffwidman: @dpkp can you comment here?

jeffwidman: nudge @dpkp

jeffwidman:
@asdaraujo This looks good to me, and I'd like to see your hard work merged. Do you mind rebasing to fix merge conflicts?

test/conftest.py
jeffwidman: Re-reading this, I understand the problem now. This makes sense why `@pytest.mark.parametrize()` won't work... I don't know if there's a better way or not. If not, can you add a comment to the code explaining why you had to do this? Just briefly summarize your above comments...  Otherwise I can totally see myself down the road forgetting the rationale here and spending a couple of hours trying to change it before I realize the issue.

tox.ini
```diff
@@ -20,6 +20,9 @@ deps =
     lz4
     xxhash
     py26: unittest2
+    decorator
+    # sugar plugin depends on specifc versions of py
+    py<1.5,>=1.4.33
```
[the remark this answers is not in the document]
jeffwidman: @asdaraujo can you also address this point?

asdaraujo:
@jeffwidman Will do! Just got back from PTO and will get back to this asap.

asdaraujo:
@jeffwidman I've addressed the comments and rebased the changes

tvoinarovskyi:
Hi there, sorry for no input on this before, had my hands full. @asdaraujo could you explain the purpose of this PR, as far as I could tell there's no actual migration of test cases themselves to `pytest`. For example, the `set_params` is implemented and never used...

tvoinarovskyi:
The problem I have with this is that it's probably not tested and will require changes later on anyway. For example, the `set_params` fixture could work only if you have 1 instance of the consumer. In several cases, you will want to create 2 instances to work as a group and test reconnection, those 2 will have different params.
I would rather have a `kafka_consumer_factory` fixture, that adds params such as broker_connection and those can be changed by providing `kwargs` to the fixture.

jeffwidman:
> as far as I could tell there's no actual migration of test cases themselves to pytest. For example, the set_params is implemented and never used...

@tvoinarovskyi this was because in https://github.com/dpkp/kafka-python/pull/1293#pullrequestreview-76326384 I requested he break his changes (which were a massive PR) into several smaller changesets. 

Your other point sounds valid, at least at first glance.

tvoinarovskyi:
@jeffwidman Got it, as this PR does not hinder test runs, we can merge it. If some changes will be needed after, we can do it in following PR's.

jeffwidman:
@asdaraujo before I merge, what do you think of the suggestion to have a `kafka_consumer_factory` fixture?

> For example, the set_params fixture could work only if you have 1 instance of the consumer. In several cases, you will want to create 2 instances to work as a group and test reconnection, those 2 will have different params.
I would rather have a kafka_consumer_factory fixture, that adds params such as broker_connection and those can be changed by providing kwargs to the factory.

asdaraujo:
Thanks, @dpkp. I like the `kafka_consumer_factory` suggestion. Let me review this and I'll push the changes.

I'm happy to open all the PRs once this is defined.

And I'll still try to stay away from buses...

asdaraujo:
@tvoinarovskyi @jeffwidman I've had a look into this and I think most of the work is already there and I can simplify things a bit. Looking for your thoughts here before I make any changes.

The `kafka_broker` fixture (a `KafkaFixture` object) already has a consumer and a producer factory methods: `get_consumers` and `get_producers`, respectively.

For the sake of simplicity, I'm thinking on the following changes:
* Get rid of the `set_params` trickery
* Modify the `kafka_consumer` and `kafka_producer` fixtures to always return the default consumer and producer (no kwargs). This should cater for most of the simple tests that don't need specific setting for consumers and producers
* Add similar factory methods to `KafkaFixture` to create simple producers and consumers

With this, if the default consumer/producer is needed the test can simple use the related consumer/producer fixture. If any specific settings are required, the test can use the `kafka_broker` fixture and use its factory methods to create what's needed for the tests.

Thoughts?

jeffwidman:
👍 This sounds like a good plan to me. Let's go ahead and do that and then get it merged so that you're unblocked for the rest of the code changes you are making. Thanks again for your patience and willingness to go through multiple revisions on this.

jeffwidman:
Any update @asdaraujo?

asdaraujo:
@jeffwidman Sorry for the delay. This month is kinda crazy for me and I've been flat out.
But I gave this a push today and I'm just running some local tests now before I update the pull request.

I made the changes along the lines we had discussed before, although I've had to change a few things from the original plan. But the overall idea is the same and the `@set_params` method has been replaced with factory fixtures.

I was also concerned with the earlier comment from @tvoinarovskyi that things were being committed without testing. So after the changes I migrated two test methods to use `pytest`: one consumer and one producer test. Once this is merged I'll work on the next PRs to convert the rest of the tests.

So, as you'll see, the size of the PR has increased a little bit, but hopefully not overwhelmingly.

jeffwidman:
Sounds good, I started to review but looks like you're still pushing updates... let me know when you're done and I'll give it a final review.

asdaraujo:
@jeffwidman I just noticed a small issue that caused Python 3.x tests to fail. Just fixed it and tests are queued again.

Not planning to change anything else, unless tests fail again.

test/fixtures.py
jeffwidman: Looks like this got missed?

test/testutil.py
```diff
@@ -1,36 +1,20 @@
...
-import logging
 import operator
 import os
-import random
 import socket
-import string
 import time
 import uuid
 
-from six.moves import xrange
```
asdaraujo: @jeffwidman Yes, I missed this. I'll take care of this in the upcoming PRs, if you don't mind.

test/fixtures.py
asdaraujo: Yep, I missed. Will take care in the upcoming PRs

test/testutil.py
jeffwidman: No worries, at all, I'm happy to clean it up myself after merging

jeffwidman:
So I'm afraid there are a couple of test failures this time around that look like legit race conditions... 

I'm not sure if they're problems with the test harness (which I don't mind ignoring other than raising an issue to fix in later PRs) or race condition in `kafka-python` (which we can fix in another PR, but should do ASAP)... 

You're now more familiar with the test harness than I am... Any idea?

asdaraujo:
I'll look into those errors

asdaraujo:
@jeffwidman I believe I've found the cause of the racing condition.

The client's `ensure_topic_exists` only checks for the presence of the metadata for the topic. The actual logs for the topic, though are created asynchronously by the brokers and by the time `ensure_topic_exists` returns some of the partitions' logs may have not been created yet.

If the client sends a message to the topic in that state, the broker will throw internally a `NotAssignedReplicaException`, which will surface to the `kafka-python` client as a `UnknownTopicOrPartitionError` exception.

I've patched the `KafkaIntegrationTestCase` by adding the following after the call to `ensure_topic_exists`:
```
        timeout = time.time() + 30
        for partition in self.client.get_partition_ids_for_topic(self.topic):
            while True:
                try:
                    req = OffsetRequestPayload(self.topic, partition, -1, 100)
                    self.client.send_offset_request([req])
                    break
                except (NotLeaderForPartitionError, UnknownTopicOrPartitionError, FailedPayloadsError) as e:
                    if time.time() > timeout:
                        raise KafkaTimeoutError('Timeout loading topic metadata!')
                    time.sleep(.1)
```

The offset request will throw the same error as the produce request if the log(s) creation is completed. In my tests I noticed that I sporadically got a few exceptions in the first loop iterations, which stopped usually in less than 1 second, after the brokers finished the creation of the topic' logs.

I'm still running tests, but this seems to fix the issues. I think it might be a good idea to move the check above to the `ensure_topic_exists` method later (in a different PR). If my tests complete successfully I'll update the PR.

I couldn't find anything in this PR that would've had introduced this though. This issues seems to have existed before. I'm still not sure why it started manifesting now.

jeffwidman:
@asdaraujo when you have time, please put up the next PR here so we can keep pushing this migration across the finish line. I'll try to be faster next time to turn this around. :-)

tvoinarovskyi:
Damn, too fast =)
@asdaraujo As this is merged already will not be too picky, but please remove `py` and `decorator` dependency. `py` module is not supported actively:

```
NOTE: this library is in maintenance mode and should not be used in new code.
```

As for `decorator`, I simply don't see any need for the package.
If you really need those packages, please elaborate why.

jeffwidman:
> please remove `py` and `decorator` dependency. `py` module is not supported actively:

The `py` dependency is a temporary import as explained here: https://github.com/dpkp/kafka-python/pull/1293#discussion_r151291693 I think it's fine to do temporarily as long as @asdaraujo has time to push through this whole migration. He's been consistent and reliable about showing up to work on this PR, even though I kept asking for revisions, so I trust him to follow through.

I'm not sure about the `decorator` dependency, don't recall if I inquired about that, possibly I forgot to.

asdaraujo:
@tvoinarovskyi @jeffwidman Yes, the `py` dependency is temporary and will just exist while we're in the process of migrating the legacy tests. I'll remove it as soon as the legacy stuff is converted.

As for `decorator`, that was a miss. It was used by my previous `@set_params` decorator, which I removed, but forgot to remove the dependency. I'll remove that in the next PR.

@jeffwidman I've started working on the new PRs but I have a big personal deadline at the end of next week, so I'll probably be quiet here until early March.

jeffwidman:
Sounds good @asdaraujo thanks for the update. We all very much appreciate the help on this, will make everyone's lives much smoother.

dpkp:
Let's please also make sure that repl fixtures continue to work without requiring a virtualenv. I very frequently run this in a shell for local testing / debugging / etc:
```
from test.fixtures import ZookeeperFixture, KafkaFixture
zk = ZookeeperFixture.instance()
k = KafkaFixture.instance(0, zk)
import logging; logging.basicConfig(level=logging.DEBUG)
consumer = KafkaConsumer(bootstrap_servers="{}:{}".format(k.host, k.port))
...
```

The above now requires that I have `py` installed + available.

jeffwidman:
> The above now requires that I have py installed + available.

Makes sense, I didn't even think about that. Thankfully, as noted in https://github.com/dpkp/kafka-python/pull/1293#issuecomment-367761594, this should be temporary:
> Yes, the py dependency is temporary and will just exist while we're in the process of migrating the legacy tests. I'll remove it as soon as the legacy stuff is converted.

jeffwidman:
@asdaraujo any idea when you'll be able to pick this up again? 

Be nice to get the rest of the tests migrated and the GSSAPI stuff under test coverage...

jeffwidman:
Bump @asdaraujo any chance you'll be able to finish migrating the legacy tests? 

I'm hoping to move toward a big 2.0 release with both #1540 and #1196, but for #1196 we need the legacy tests migrated... You understand this testing framework better than anyone else right now, so if you have any spare cycles to finish this migration we'd certainly appreciate it.

PR #1301, opened 2017-11-17
changed for loop to use enumerate()

TheAtomicOption:
enumerate() is cleaner and allows removing an assignment statement.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -747,13 +747,12 @@ class BrokerConnection(object):
             return ()
 
         # augment respones w/ correlation_id, future, and timestamp
-        for i in range(len(responses)):
+        for i, response in enumerate(responses):
             (correlation_id, future, timestamp) = self.in_flight_requests.popleft()
             latency_ms = (time.time() - timestamp) * 1000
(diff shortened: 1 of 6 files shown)


dpkp:
I chose not to use enumerate because the loop explicitly modifies the list itself and that typically doesn't play well with in place iterators

TheAtomicOption:
It's not changing the count or order of records, just the values so that _shouldn't_ be an issue in python as [the example in the PEP itself shows doing exactly that](https://docs.python.org/2.3/whatsnew/section-enumerate.html). 

That said, all these build checks failed, which really surprised me since it seemed to work on my clone. The only issue I expected on seeing the code was maybe a performance drag from the change in element size of the data from record --> (record, stuff), but that was there to start with.

dpkp:
Apologies -- all test fixtures were failing due to an unrelated issue. Please pull latest changes from master and re-run tests.

TheAtomicOption:
Yay it passes, and I am no longer confused! :)

dpkp:
Thanks for the fixup and link to PEP w/ example!

PR #1317, opened 2017-12-07
Minor Exception cleanup

jeffwidman:


kafka/metrics/metric_name.py
```diff
@@ -50,9 +50,9 @@ def __init__(self, name, group, description=None, tags=None):
             tags (dict, optional): Additional key/val attributes of the metric.
         """
         if not (name and group):
-            raise Exception('name and group must be non-empty.')
+            raise ValueError('name and group must be non-empty.')
         if tags is not None and not isinstance(tags, dict):
-            raise Exception('tags must be a dict if present.')
+            raise ValueError('tags must be a dict if present.')
```
jeffwidman: Alternatively, both of these could be changed to `KafkaConfigurationError`

kafka/consumer/fetcher.py
```diff
@@ -478,8 +478,8 @@ def _unpack_message_set(self, tp, records):
         # caught by the generator. We want all exceptions to be raised
         # back to the user. See Issue 545
         except StopIteration as e:
-            log.exception('StopIteration raised unpacking messageset: %s', e)
-            raise Exception('StopIteration raised unpacking messageset')
+            log.exception('StopIteration raised unpacking messageset')
+            raise RuntimeError('StopIteration raised unpacking messageset')
```
jeffwidman: Potentially this should be a `KafkaError` although I'm unclear when to reach for that instead of `RuntimeError`

tvoinarovskyi: RuntimeError is what would be raised by the convention on Python3 if we have a not handled StopIteration raised inside of a generator.

PR #1320, opened 2017-12-08
Give better struct errors

### LINKED ISSUE #1318: Handling of struct errors doesn't print the specific error message
When a `struct.error` is thrown during `_pack()` or `_unpack()`, we catch and re-raise as a `ValueError`: https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/types.py#L11-L12

However, we're shadowing the word `error` so we lose a handle on the specific exception and cannot print the specific error message.

jeffwidman:
Stop shadowing the word `error` because we want to know what the specific exception message was.

Also give more details on exactly which value failed. We don't know who submitted the value, but perhaps it's unique enough that we can debug it better.

Fix #1318

### DIFF
--- a/kafka/protocol/types.py
+++ b/kafka/protocol/types.py
@@ -8,16 +8,20 @@ from .abstract import AbstractType
 def _pack(f, value):
     try:
         return pack(f, value)
-    except error:
-        raise ValueError(error)
+    except error as e:
+        raise ValueError("Error encountered when attempting to convert value: "
+                        "{} to struct format: '{}', hit error: {}"
+                        .format(value, f, e))
 
 
 def _unpack(f, data):
     try:
         (value,) = unpack(f, data)
         return value
-    except error:
-        raise ValueError(error)
+    except error as e:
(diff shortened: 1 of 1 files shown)


kafka/protocol/types.py
```diff
@@ -8,16 +8,20 @@
 def _pack(f, value):
     try:
         return pack(f, value)
-    except error:
-        raise ValueError(error)
+    except error as e:
+        raise ValueError("Error encountered when attempting to convert value: "
+                        "{} to struct format: '{}', hit error: {}"
```
tvoinarovskyi: Maybe `{!r}` just in case str is overridden.

tvoinarovskyi:
Looks good.

dpkp:
+1

kafka/protocol/types.py
jeffwidman: I forgot to do this earlier, changed in https://github.com/dpkp/kafka-python/pull/1373

PR #1333, opened 2017-12-28
Fix typo in _try_authenticate_plain

everpcpc:
Returned data was lost, thus `if data != b'\x00\x00\x00\x00':` always fails .


dpkp:
Thanks!

Issue #1331, opened 2017-12-26
MetadataResponse received by _handle_group_coordinator_response


dpkp:
I saw this error during local testing. It looks like there is a thread synchronization issue that leads to responses being processed incorrectly (here a MetadataResponse was received by a group coordinator request future).
```
1514315653.454923 conn.py                 MainThread-heartbeat Connect attempt to <BrokerConnection node_id=0 host=localhost/127.0.0.1 port=65231> returned error 61. Disconnecting.
1514315653.455129 client_async.py         MainThread-heartbeat Node 0 connection failed -- refreshing metadata
1514315653.583284 client_async.py         MainThread-heartbeat Node 1 connection failed -- refreshing metadata
Exception in thread MainThread-heartbeat:
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Users/dpowers/Code/kafka-python/kafka/coordinator/base.py", line 891, in run
    self._run_once()
  File "/Users/dpowers/Code/kafka-python/kafka/coordinator/base.py", line 922, in _run_once
    if not self.coordinator.lookup_coordinator().is_done:
  File "/Users/dpowers/Code/kafka-python/kafka/coordinator/base.py", line 284, in lookup_coordinator
    self._find_coordinator_future = self._send_group_coordinator_request()
  File "/Users/dpowers/Code/kafka-python/kafka/coordinator/base.py", line 650, in _send_group_coordinator_request
    _f.add_callback(self._handle_group_coordinator_response, future)
  File "/Users/dpowers/Code/kafka-python/kafka/future.py", line 52, in add_callback
    self._call_backs('callback', [f], self.value)
  File "/Users/dpowers/Code/kafka-python/kafka/future.py", line 83, in _call_backs
    raise e
  File "/Users/dpowers/Code/kafka-python/kafka/future.py", line 79, in _call_backs
    f(value)
  File "/Users/dpowers/Code/kafka-python/kafka/coordinator/base.py", line 657, in _handle_group_coordinator_response
    error_type = Errors.for_code(response.error_code)
AttributeError: 'MetadataResponse_v0' object has no attribute 'error_code'
```

dpkp:
I can only reproduce this error on PR #1332 , so I suspect it is caused by returning multiple responses from conn.recv

dpkp:
[...]

PR #1339, opened 2018-01-10
Improve KafkaConsumer cleanup

dpkp:
After digging into #1331 a bit it I found a few issues:

(1) heartbeat thread is not stopped unless KafkaConsumer is closed explicitly (or python interpreter exits).
(2) most connection objects have a circular reference preventing garbage collection prior to interpreter shutdown
(3) the client object also has a circular reference preventing gc prior to shutdown

This PR should fix these issues by using weakrefs to break circular references and by shutting down local resources like the heartbeat thread during gc via `__del__`

It appears that there are still circular references preventing gc of KafkaMetrics and possibly SubscriptionState. These do not prevent the consumer, coordinator, client, or connection objects from being cleaned up, so I'm going to leave further fixes for a later PR.

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -6,6 +6,7 @@ import functools
 import logging
 import random
 import threading
+import weakref
 
 # selectors in stdlib as of py3.4
 try:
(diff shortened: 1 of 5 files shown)


jeffwidman:
Nice catches!

PR #1344, opened 2018-01-12
Avoid KeyError when filtering fetchable partitions

dpkp:
I noticed this failure in a failed travis run. If we attempt to remove a partition that is not currently fetchable, the set will raise a KeyError. To avoid, we test that the element is in the set before removing it.

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -638,11 +638,9 @@ class Fetcher(six.Iterator):
     def _fetchable_partitions(self):
         fetchable = self._subscriptions.fetchable_partitions()
         if self._next_partition_records:
-            if self._next_partition_records.topic_partition in fetchable:
-                fetchable.remove(self._next_partition_records.topic_partition)
+            fetchable.discard(self._next_partition_records.topic_partition)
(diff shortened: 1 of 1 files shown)


tvoinarovskyi:
jfyi, `discard` can be used here, as it's a set.
```python
>>> x = set([1, 2, 3])
>>> x.remove(0)
  File "<stdin>", line 1, in <module>
KeyError: 0
>>> x.discard(0)
>>> x
{1, 2, 3}
>>> x.discard(1)
>>> x
{2, 3}
```

dpkp:
TIL!

On Jan 11, 2018 6:49 PM, "Taras Voinarovskyi" <notifications@github.com>
wrote:

jfyi, discard can be used here, as it's a set.

>>> x = set([1, 2, 3])>>> x.remove(0)
  File "<stdin>", line 1, in <module>KeyError: 0>>> x.discard(0)>>> x
{1, 2, 3}>>> x.discard(1)>>> x
{2, 3}

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dpkp/kafka-python/pull/1344#issuecomment-357128297>,
or mute
the thread

PR #1345, opened 2018-01-12
Name heartbeat thread with group_id; use backoff when polling

dpkp:
Another set of small improvements that I stashed while working on other PRs. To help with log analysis we can name the heartbeat thread using the consumer group_id .  I've also added a short timeout to the internal heartbeat poll to avoid tight loops.


dpkp:
This is causing problems on pypy for some reason. Needs investigation.

dpkp:
A bit too trigger happy

dpkp:
I'm not sure what's wrong with me today. The thread rename should be merged. The timeout change should not. I'm just going to push that manually.

Issue #1184, opened 2017-08-25
KafkaTimeoutError


hugoren:
the err:
       KafkaTimeoutError: Batch for TopicPartition(topic='test', partition=0) containing 1 record(s) expired: 30 seconds have passed since batch creation plus linger time

my code:

    def kafka_producer():
         producer = KafkaProducer(bootstrap_servers='{0}:{1}'.format(HOST, PORT),
                             api_version=(0, 10, 1)
                             )
       try:
           future = producer.send('test', b'hello kafka')
           producer.flush()
           record_metadata = future.get(timeout=10)
           print(record_metadata)
    except KafkaError as e:
           print(e)

dpkp:
Yes, KafkaTimeout is an exception you may get when using the asynchronous producer.

tylerjharden:
@dpkp That's great and all, but how do you resolve it / why does it occur?

skskcco2o17:
I am facing the same issue. My Kafka Broker (v 2.4.0) is running in OpenShift container.  My Python producer code is similar to the above and that is causing the same issue. 

```
KafkaTimeoutError: Batch for TopicPartition(topic='my_test_topic', partition=0) containing 1 record(s) expired: 31 seconds have passed since batch creation plus linger time
   File "C:\Users\SUPALCHOWDHURY\Anaconda3\lib\site-packages\kafka\producer\future.py", line 65, in get
    raise self.exception # pylint: disable-msg=raising-bad-type
```

My kafka-python version is  2.0.1

I am fighting on the issue for last two days.  I am frustrated. Please assist how do you resolve it / why does it occur?

I have tested using the Kafka sh script to send the topic it's working. So where is the problem in python?

sushen-indifi:
Having the same issue

```
KafkaTimeoutError: Batch for TopicPartition(topic='test', partition=0) containing 1 record(s) expired: 30 seconds have passed since last append
```

skskcco2o17:
Why Kafka-python team is not responding this sever issue, which lot's of  development team are facing? If team is not supporting then how we rely  and use this library to the enterprise application for commercialization?

tvoinarovskyi:
@skskcco2o17 Kafka-python is supported by volunteers and does not have any commercial structure funding the project, so we all do it in our free time. If you need production enterprise support - there are a lot of companies (like Confluent) providing it.

0Jvang:
i solve it

0Jvang:
[...]

PR #1348, opened 2018-01-18
Remove assertion with side effect

bgedik:
When running Python code with PYTHONOPTIMIZE=1 assertions are removed. Needless to say, an assertion should never have a side effect that is critical for the program's execution. @dpkp You seem to have ignored the following: https://github.com/dpkp/kafka-python/issues/1184 but I was getting the same symptoms and it took quite some time to figure out the cause.


dpkp:
Well done!

bgedik:
I don't thin the failure is relevant.

PR #1352, opened 2018-01-23
fix reconnect_backoff_max_ms default config bug in KafkaProducer

YaoC:
According to the [Kafka Document](http://kafka.apache.org/documentation.html#producerconfigs) and the usage in [client_async.py](https://github.com/dpkp/kafka-python/blob/master/kafka/client_async.py) ，the config  `reconnect_backoff_max` should be `reconnect_backoff_max_ms` 
: )


dpkp:
Great catch, thanks!

PR #1353, opened 2018-01-24
Fix for Python 3 byte string handling

christophelec:
Due to differences in byte string handling between Python 2 and 3, this fix is required to be able to use Kerberos authentication with Python 3


dpkp:
Test failure is not related (503 when downloading apache kafka tarball). Thanks for the fix!

PR #1362, opened 2018-01-31
Use absolute imports everywhere

kevinkjt2000:


.gitignore
```diff
@@ -13,4 +15,5 @@ docs/_build
 .cache*
 .idea/
 integration-test/
-tests-env/
\ No newline at end of file
+tests-env/
+.pytest_cache/
```
jeffwidman: These changes should probably be in a separate PR as they are unrelated to import changes

kevinkjt2000: I have removed that by amending my commit.

kevinkjt2000:
@jeffwidman Python has trouble with relative imports when libraries are being imported from a zip file that is not extracted.  These changes fix an issue in a new product of the company I develop for.

dpkp:
I'm fine with this if/when tests pass.

kevinkjt2000:
The test failure seems strange to me https://travis-ci.org/dpkp/kafka-python/jobs/335863422
Looks like version 0.10.2.1 of kafka failed only with pypy.

jeffwidman:
There's some flaky tests... I restarted the failing test job.

> Python has trouble with relative imports when libraries are being imported from a zip file that is not extracted.

To clarify, you're saying that you are importing python code straight from a zip file without extracting it? I did not realize python let you do this...

kevinkjt2000:
@jeffwidman Yes.  https://docs.python.org/2/library/zipimport.html  Inserting a .zip file into the sys path allows python to import files from it.  Which, after the trouble I have seen here, I would not recommend doing.  All .so files also cannot load from a .zip file.  Fun stuff.

dpkp:
have you looked at https://github.com/pantsbuild/pex ?

kevinkjt2000:
@dpkp pex looks promising!  I will recommend it to my co-workers that work on that codebase more often.

PR #1364, opened 2018-02-01
KAFKA-3949: Avoid race condition when subscription changes during rebalance

### LINKED ISSUE #1241: KAFKA-3949: Fix race condition between group rebalance and metadata update
Details in https://issues.apache.org/jira/browse/KAFKA-3949

Note that the fix refactored a fair bit of code for managing subscription state: https://github.com/apache/kafka/pull/1762

And then KIP-70 (tracked in #1242) further modified this code.

Related: #1237 / #1240 / #1242

dpkp:
Fix #1241

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -84,6 +84,8 @@ class ConsumerCoordinator(BaseCoordinator):
                 self.config[key] = configs[key]
 
         self._subscription = subscription
+        self._is_leader = False
+        self._joined_subscription = set()
         self._metadata_snapshot = self._build_metadata_snapshot(subscription, client.cluster)
         self._assignment_snapshot = None
         self._cluster = client.cluster
@@ -119,6 +121,7 @@ class ConsumerCoordinator(BaseCoordinator):
         self.consumer_sensors = ConsumerCoordinatorMetrics(
             metrics, self.config['metric_group_prefix'], self._subscription)
 
+        #self._handle_metadata_update(self._cluster)
         self._cluster.request_update()
         self._cluster.add_listener(WeakMethod(self._handle_metadata_update))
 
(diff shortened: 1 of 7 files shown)


kafka/coordinator/consumer.py
```diff
@@ -119,6 +121,7 @@ def __init__(self, client, subscription, metrics, **configs):
         self.consumer_sensors = ConsumerCoordinatorMetrics(
             metrics, self.config['metric_group_prefix'], self._subscription)
 
+        #self._handle_metadata_update(self._cluster)
```
jeffwidman: Why add this if it's commented out?

kafka/coordinator/consumer.py
```diff
@@ -132,11 +135,22 @@ def protocol_type(self):
...
+        # why? because we are using this strange method group_protocols,
+        # which is seemingly innocuous, to set internal state (_joined_subscription)
+        # that is later used to check whether metadata has changed since we joined a group
+        # but there is no guarantee that this method, group_protocols, will get called
+        # in the correct sequence or that it will only be called when we want it to be.
+        # So this really should be moved elsewhere, but I don't have the energy to
+        # work that out right now. If you read this at some later date after the mutable
+        # state has bitten you... I'm sorry! It mimics the java client, and that's the
+        # best I've got for now.
+        self._joined_subscription = set(self._subscription.subscription)
```
jeffwidman: 😃 loved this comment

tvoinarovskyi: Yea, I know. Hate it, as the method is basically a getter. They put it there, just cause it should be in `ConsumerCoordinator` but not in `AbstractCoordinator` and only this method was overridable... Basically a big hack. BTW the method is called `metadata()` nowdays. https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L148

dpkp:
> Wow, this is a lot more work than I remembered when I investigated how Java fixed this... is that because you're also porting in other updates from the Java client? Or is my memory just faulty?

This is just 3949. The upstream PR is here: https://github.com/apache/kafka/pull/1762

kafka/coordinator/consumer.py
dpkp: Will delete -- thanks

Issue #1383, opened 2018-02-16
Large CPU usage due to busy waiting by KafkaConsumer.poll()


braedon:
`KafkaConsumer.poll()` is effectively busy waiting, as `KafkaConsumer._poll_once()` doesn't respect `timeout_ms`.

`_poll_once()` sets `timeout_ms` of `KafkaClient.poll()` to:
```
min(timeout_ms, self._coordinator.time_to_next_poll())
```
However, `ConsumerCoordinator.time_to_next_poll()` returns the time in seconds, not milliseconds. The effective timeout is always much lower than `timeout_ms`, causing `KafkaConsumer.poll()` to spin, consuming a lot of CPU.

PR #1384, opened 2018-02-16
Correctly respect timeouts in consumer poll interface

braedon:
`ConsumerCoordinator.time_to_next_poll()` returns seconds, so need to convert to milliseconds.

Fixes https://github.com/dpkp/kafka-python/issues/1383


tvoinarovskyi:
LGTM will try to get the build green and ready to go 🚢

tvoinarovskyi:
In kafka-python we have the pattern all over the place. Ie, it returns seconds, not milliseconds. I think it's ok.

dpkp:
LGTM

PR #1386, opened 2018-02-18
Add Admin CreatePartitions API call

alexef:
I've been using it for some time in a personal project.

Protocol should match:https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/common/requests/CreatePartitionsRequest.java and  https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/common/requests/CreatePartitionsResponse.java

### DIFF
--- a/kafka/protocol/__init__.py
+++ b/kafka/protocol/__init__.py
@@ -42,4 +42,6 @@ API_KEYS = {
     31: 'DeleteAcls',
     32: 'DescribeConfigs',
     33: 'AlterConfigs',
+    36: 'SaslAuthenticate',
+    37: 'CreatePartitions',
 }
--- a/kafka/protocol/admin.py
+++ b/kafka/protocol/admin.py
@@ -256,7 +256,6 @@ class DescribeGroupsResponse_v1(Response):
     )
 
 
-
 class DescribeGroupsRequest_v0(Request):
     API_KEY = 15
     API_VERSION = 0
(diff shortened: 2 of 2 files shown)


kafka/protocol/__init__.py
```diff
@@ -42,4 +42,6 @@
     31: 'DeleteAcls',
     32: 'DescribeConfigs',
     33: 'AlterConfigs',
+    36: 'SaslAuthenticate',
```
alexef: I saw it was missing, but anyway where is API_KEYS used?

tvoinarovskyi: It's not used atm.

kafka/protocol/admin.py
```diff
@@ -333,3 +332,33 @@ class SaslAuthenticateRequest_v0(Request):
 
 SaslAuthenticateRequest = [SaslAuthenticateRequest_v0]
 SaslAuthenticateResponse = [SaslAuthenticateResponse_v0]
+
+
+class CreatePartitionsResponse_v0(Response):
+    API_KEY = 37
+    API_VERSION = 0
+    SCHEMA = Schema(
+        ('topic_error_codes', Array(
```
tvoinarovskyi: Seems like you missed `throttle_time_ms` field. Lets set the name to `topic_errors` to be exact to docs also. See https://kafka.apache.org/protocol#The_Messages_CreatePartitions.

alexef: Ouch, my bad, pushed a fix.

alexef:
@tvoinarovskyi thank you for the review! Is there something else I need to do to get this merged, or just wait for more approval?

jeffwidman:
Looked good to me, so I merged it. Thanks!

kafka/protocol/admin.py
Vijay-P: Is Line 335 in kafka/admin/client.py a residual mistake from this change?

PR #1389, opened 2018-02-19
Use hardware accelerated CRC32C function if available

tvoinarovskyi:
@dpkp @jeffwidman Hehe, I'm an idiot for this =) So, here I was pushing my butt to accelerate varint implementation, but did not look properly at the profiler.
Long story short, while accelerating the V2 implementation for `aiokafka` I found that `_crc32c.py` is super slow and takes more time for 1 call than 100 `append`. So as a fast fix I propose to use https://github.com/ICRAR/crc32c if available on the system. I will look into `_crc32c.py` script later to make it a bit faster.

Benchmark without HW crc32:
```
.....................
batch_read_v0: Mean +- std dev: 1.42 ms +- 0.03 ms
.....................
batch_read_v1: Mean +- std dev: 1.44 ms +- 0.03 ms
.....................
batch_read_v2: Mean +- std dev: 3.10 ms +- 0.06 ms
```

With HW crc32:
```
.....................
batch_read_v0: Mean +- std dev: 1.48 ms +- 0.05 ms
.....................
batch_read_v1: Mean +- std dev: 1.49 ms +- 0.05 ms
.....................
batch_read_v2: Mean +- std dev: 1.09 ms +- 0.04 ms
```

So now V2 is relatively the same speed as V1 and V0. Now we can move on to TransactionalProducer =)

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -70,3 +70,16 @@ Install the `python-snappy` module
 .. code:: bash
 
     pip install python-snappy
+
+
+Optional crc32c install
+***********************
+Highly recommended if you are using Kafka 11+ brokers. For those `kafka-python`
+uses a new message protocol version, that requires calculation of `crc32c`,
+which differs from `zlib.crc32` hash implementation. By default `kafka-python`
+calculates it in pure python, which is quite slow. To speed it up we optionally
+support https://pypi.python.org/pypi/crc32c package if it's installed.
+
+.. code:: bash
+
+    pip install crc32c
--- a/kafka/record/util.py
+++ b/kafka/record/util.py
@@ -1,6 +1,10 @@
 import binascii
 
 from kafka.record._crc32c import crc as crc32c_py
+try:
+    from crc32c import crc32 as crc32c_c
+except ImportError:
+    crc32c_c = None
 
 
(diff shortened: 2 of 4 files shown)


jeffwidman:
Oh interesting. Two thoughts: 
1. if we go down this route of fixing by adding an optional dependency, there should be a note about it in the docs as an optional way to speedup `kafka-python`.
2. It's not clear to me what the other options are beyond adding an optional dependency... since this isn't pure python, does that create problems with directly vendoring it in like we do a few other libs?

tvoinarovskyi:
@jeffwidman Yea, will patch the docs.
Yes, you can't easiely bundle a C extension module, you will need to correctly build it. As for alternitives, in aiokafka I plan to use the C implementation taken from librdkafka...
I will also try to make the python script at least somewhat faster.

docs/install.rst
```diff
@@ -70,3 +70,16 @@ Install the `python-snappy` module
 .. code:: bash
 
     pip install python-snappy
+
+
+Optional crc32c install
```
tvoinarovskyi: @jeffwidman Could you read this through, good enough?

jeffwidman: Looks great

PR #1395, opened 2018-02-26
Support alternative lz4framed

everpcpc:
Add support for https://github.com/Iotic-Labs/py-lz4framed

### DIFF
--- a/kafka/codec.py
+++ b/kafka/codec.py
@@ -27,6 +27,11 @@ except ImportError:
     lz4f = None
 
 try:
+    import lz4framed
+except ImportError:
+    lz4framed = None
+
+try:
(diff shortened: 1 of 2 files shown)


jeffwidman:
Are there tests that should also be updated to run using this new library as well?

everpcpc:
Tests added, but maybe it's too much to add this for all python envs?

everpcpc:
I replaced `lz4` with `py-lz4framed` in the test and travis-ci seems OK. I'll remove this later.
It's maybe not necessary to add tests for this. Because `lz4f` is also not test. And it takes too long based on current tests.

dpkp:
Looks great, thanks!

PR #1403, opened 2018-03-02
Fix error var name in _unpack

j2gg0s:


j2gg0s:
This is a bug. Maybe we should merge it. @dpkp

dpkp:
Agree - unfortunately i'm travelling and dont have access to my laptop. I
plan to merge when i return home in a few days

tvoinarovskyi:
Merged it, don't see any reason not to =)

tvoinarovskyi:
@j2gg0s GJ

jeffwidman:
oops, my bad!

PR #1411, opened 2018-03-08
Add BrokerConnection.connect_blocking()

dpkp:
This PR adds a blocking connect to BrokerConnection that will attempt to connect to all ip/ports found via a single dns lookup. We can use this during bootstrap and version probes to handle brokers that advertise a hostname that has multiple interfaces but the broker only listens on one of them. This can happen when a host has both an IPv4 and an IPv6 interface (but the kafka broker binds to only one of them). This will also help in contexts where the bootstrap_servers list includes a multi-address dns entry where not all results are available/healthy.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -71,17 +71,7 @@ class SimpleClient(object):
             )
 
         conn = self._conns[host_key]
-        conn.connect()
-        if conn.connected():
-            return conn
-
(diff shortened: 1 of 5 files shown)


PriyaranjanMaratheDish:
How to handle this when I am trying to consume on my mac? Seeing the issue and when I add the api version I don't get any results.

PR #1422, opened 2018-03-09
Make BrokerConnection .host / .port / .afi immutable

dpkp:
Separating this from #1411. The underlying issue is that KafkaClient checks whether the broker metadata has changed when deciding whether to "recycle" an existing BrokerConnection or not (see `_maybe_connect`). The check examines .host and .port directly, which means that we are creating false positives when we assign new resolved values to these attributes. This means that we are not recycling BrokerConnection objects the way we want to at best, and at worst it means that we are not dropping the objects before trying all possible resolved connection addresses.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -78,6 +78,14 @@ except ImportError:
     gssapi = None
     GSSError = None
 
+
+AFI_NAMES = {
+    socket.AF_UNSPEC: "unspecified",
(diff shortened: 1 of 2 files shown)


jeffwidman:
> at worst it means that we are not dropping the objects before trying all possible resolved connection addresses.

I'm struggling to parse this... should the word "not" be omitted from the above?

PR #1426, opened 2018-03-09
Defer version check until after bootstrap succeeds

dpkp:
Separated from #1411 . This PR moves version checking from `__init__` into `_bootstrap` . This means that if a cluster is down and bootstrapping fails, version checking will not cause an exception to be raised immediately. Instead, we wait until the initial bootstrap succeeds and then run version checks. In order to handle this I've changed the "unset" version from `None` to `(0, 0)`. This also allows us to simplify version checking code internally. `(0, 0)` should cause us to use the lowest available version (likely 0.8.2) if needed but still flag that version checking has not yet happened.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1426)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -196,6 +196,9 @@ class KafkaClient(object):
             assert self.config['api_version'] in self.API_VERSIONS, (
                 'api_version [{0}] must be one of: {1}'.format(
                     self.config['api_version'], str(self.API_VERSIONS)))
+        else:
+            # This should get updated after a successful bootstrap
+            self.config['api_version'] = (0, 0)
 
         self.cluster = ClusterMetadata(**self.config)
         self._topics = set()  # empty set will fetch all topic metadata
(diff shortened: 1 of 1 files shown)


kafka/client_async.py
```diff
@@ -821,10 +825,10 @@ def check_version(self, node_id=None, timeout=2, strict=False):
             This is only possible if node_id is None.
 
         Returns: version tuple, i.e. (0, 10), (0, 9), (0, 8, 2), ...
+            (0, 0) returned if the version cannot be determined,
+            typically due to networking.
 
         Raises:
-            NodeNotReadyError (if node_id is provided)
-            NoBrokersAvailable (if node_id is None)
```
jeffwidman: So what happens if bootstrapping eventually fails? Will these still get raised somewhere else in the stack?

I know there's code in the wild that is expecting these exceptions, as I used it myself when I updated the Datadog kafka consumer lag check to use the async client instead of the old SimpleClient. 

To clarify, I'm not against this change, I just want to understand the new behavior.

jeffwidman:
I like this PR's goal of separating the "liveness" check from the "which version" check.

jeffwidman:
What's needed to push this across the line?

dpkp:
Going to leave this out of 1.4.X patch release as it seems like a largish change. Will land for next release.

kurtmckee:
Did this land in 2.0?

dpkp:
Both bootstrap and check_version have changed substantially since I worked on this PR. Closing. This feature will require a different PR.

PR #1438, opened 2018-03-10
KAFKA-4160: Ensure rebalance listener not called with coordinator lock

dpkp:
See https://issues.apache.org/jira/browse/KAFKA-4160 . The original fix was intended to support rebalance listeners that may run longer than the session timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1438)

### DIFF
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -338,7 +338,36 @@ class BaseCoordinator(object):
                 return float('inf')
             return self.heartbeat.time_to_next_heartbeat()
 
+    def _reset_join_group_future(self):
+        with self._lock:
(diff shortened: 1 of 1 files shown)


jeffwidman:
Any reason not to merge @dpkp ?

dpkp:
This should not be merged until locking / deadlock issues are addressed. At least it should be rebased onto master after latest lock changes, and then I think we need to look closely at the ensure_active_group changes to make sure locks are maintained.

wbarnha (changes requested):
Waiting to merge until deadlock is resolved.

Issue #1256, opened 2017-10-13
Document how to add callbacks/errbacks to KafkaProducer message Futures


jeffwidman:
We should document how to do this: https://stackoverflow.com/questions/46388116/how-to-add-a-failure-callback-for-kafka-python-kafka-kafkaproducersend

If my answer is amiss at all, please feel free to correct it.

jeffwidman:
Fixed in https://github.com/dpkp/kafka-python/pull/1441

PR #1441, opened 2018-03-13
Adds add_callback/add_errback example to docs

Berkodev:
After spending quite some time figuring out how to properly add callbacks/errbacks to the producer.send future obj, decided to add a basic example to the docs (Issue #1256 ).

I didn't really know where to put it as there seemed to be no examples in the docs outside of "Usage Overview", so I just put it there.

I am willing to expand upon this if needed, but I'll need to know where exactly to put it in and some guidance as to what you want to see in terms of explanation/example.

Have a good day!

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -91,6 +91,18 @@ KafkaProducer
     for _ in range(100):
         producer.send('my-topic', b'msg')
 
+    def on_send_success(record_metadata):
+        print(record_metadata.topic)
+        print(record_metadata.partition)
+        print(record_metadata.offset)
+
+    def on_send_error(excp):
+        log.exception()
+        # handle exception
+
+    # produce asynchronously with callbacks 
+    producer.send('my-topic', b'raw_bytes').add_callback(on_send_success).add_errback(on_send_error)
+
     # block until all async messages are sent
     producer.flush()
 


Berkodev:
No worries, I'm actually on a vacation myself next week, so will be available the following week if needed.

Enjoy the rest of your vacation!

docs/usage.rst
```diff
@@ -91,6 +91,18 @@ KafkaProducer
     for _ in range(100):
         producer.send('my-topic', b'msg')
 
+    def on_send_success(record_metadata):
+        print(record_metadata.topic)
+        print(record_metadata.partition)
+        print(record_metadata.offset)
+
+    def on_send_error(excp):
+        log.exception()
```
jeffwidman: Just doublechecking, since I'm not sure how this works:
Is the send exception always going to be the most recent one raised? If you screwed up your callback somehow and threw an exception, it would hide the send exception.

 Perhaps it'd be better to illustrate how to pass the `excp` into `log.error()`? This is honestly less of a `kafka-python` thing, and more just showing clearly how python logging works, as we often get questions on the issue tracker about that as well... 

Just a thought, I do not have a strong opinion, and I am not even 100% sure how this works in this particular example as I haven't played with it...

dpkp: I agree -- `log.exception()` may not always work here because we don't guarantee that the errback is evaluated within the context of a exception handler. It would be better to pass the exception explicitly:
```
def on_send_error(excp):
    log.error('I am an errback', exc_info=excp)
```

dpkp:
Thanks!

Issue #1218, opened 2017-09-15
Should Kafka stop logging during shutdown?


mauhcs:
I keep having this issue with kafka:
```
Exception ignored in: <bound method KafkaProducer.__del__ of <kafka.producer.kafka.KafkaProducer object at 0x7f03d57c1518>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/kafka/producer/kafka.py", line 400, in __del__
  File "/usr/local/lib/python3.5/dist-packages/kafka/producer/kafka.py", line 413, in close
  File "/usr/lib/python3.5/logging/__init__.py", line 1279, in info
  File "/usr/lib/python3.5/logging/__init__.py", line 1415, in _log
  File "/usr/lib/python3.5/logging/__init__.py", line 1425, in handle
  File "/usr/lib/python3.5/logging/__init__.py", line 1487, in callHandlers
  File "/usr/lib/python3.5/logging/__init__.py", line 855, in handle
  File "/usr/lib/python3.5/logging/__init__.py", line 1047, in emit
  File "/usr/lib/python3.5/logging/__init__.py", line 1037, in _open
NameError: name 'open' is not defined
```
If you go to the lines 400 and 413 in the Kafka repo (you can use the link below), you will see that kafka is logging during shutdown:

https://github.com/dpkp/kafka-python/blob/b1ae45c10d46e881492b0fd37f0919cac6d79224/kafka/producer/kafka.py#L399

This discussion in Python bug tracker suggests that logging during shutdown should be avoided:
https://docs.python.org/3/library/weakref.html

I made the following simple example to reproduce the issue. It appears when I am trying to define my own logger (setting the filename in the basicConfig):

```python
#!/usr/bin/python3
'''
Demonstrate issue:
'''
from kafka import KafkaProducer
import logging

# set up logging to file - see previous section for more details
logging.basicConfig(level=logging.INFO,
                    filename='/tmp/my_log_file.log')

prod = KafkaProducer(bootstrap_servers="localhost")

def main():
    prod.send("test",b"my message.")

if __name__ == "__main__":
    main()

prod.close() #Edit from comments
```

Copy the code and save in a file called `test.py` and in the terminal run: ```python -m "import test" ```
You will get the issue I got. (if you have a kafka instance running in the localhost)

The problem with this is that I cannot define my own loggings and gracefully terminate the program. 

Should we add some try/except in de close method in kafka? 

Or I am missing something?

[...]

PR #1445, opened 2018-03-16
Deleted __del__ as fix for Issue #1218

matthew-mcateer:
There was previously an issue with kafka logging during shutdown, after all of the global variables had been set to `None`. Most of the suggestions to a fix involved using `atexit.register` at the end of the constructor. After deleting the `__del__` function, looking over the rest of this code, and having no trouble with the Kafka producer logging afterwards, I concluded that this fixed the logging issue. Plus, ever since Python 3.4, it generally seems unwise to use __del__.

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -420,9 +420,6 @@ class KafkaProducer(object):
                     pass
         self._cleanup = None
 
-    def __del__(self):
-        self.close(timeout=0)
-
     def close(self, timeout=None):
         """Close this producer.
 


jeffwidman:
@matthew-mcateer FYI if you include the #1218 directly in your commit message or your pull request description, github will cross-link this with the original issue. You tried to do that by putting it in the PR title, but unfortunately, github only cross-links if the issue number is in the body, not in the title.

Additionally, if you just put `Fix #1218` at the end of your PR comment, then when this is merged github will automagically close the associated issue.

dpkp:
I would prefer that we continue calling `close` on `__del__` in order to avoid leaks.

PR #1447, opened 2018-03-18
Add codec validators to record parser and builder for all formats.

### LINKED ISSUE #1443: In 1.4.1, unsupported codec in consumer no longer gives clear error
In 1.3.5, attempting to consume a message encoded with an unavailable codec (in my case, lz4) would give you a clear error message `LZ4 decompression unsupported`. However, in 1.4.1 I'm getting a far more confusing error:
```
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/consumer/group.py", line 603, in poll
    records = self._poll_once(remaining, max_records)
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/consumer/group.py", line 652, in _poll_once
    records, _ = self._fetcher.fetched_records(max_records)
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/consumer/fetcher.py", line 337, in fetched_records
    self._next_partition_records = self._parse_fetched_data(completion)
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/consumer/fetcher.py", line 788, in _parse_fetched_data
    unpacked = list(self._unpack_message_set(tp, records))
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/consumer/fetcher.py", line 455, in _unpack_message_set
    for record in batch:
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/record/default_records.py", line 249, in __iter__
    self._maybe_uncompress()
  File "/usr/lib/python2.7/site-packages/kafka_python-1.4.1-py2.7.egg/kafka/record/default_records.py", line 165, in _maybe_uncompress
    uncompressed = lz4_decode(data.tobytes())
TypeError: 'NoneType' object is not callable
```
I haven't looked too closely, but it looks like the new `kafka.record` codepath doesn't check `has_lz4()` before calling `lz4_decode`, causing the above.

tvoinarovskyi:
Also removed `decorator` dependency on tests
Fixes #1443

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import
 import inspect
 
 import pytest
-from decorator import decorate
 
 from test.fixtures import KafkaFixture, ZookeeperFixture
 from test.testutil import kafka_version, random_string
@@ -73,6 +72,7 @@ def kafka_consumer_factory(kafka_broker, topic, request):
     def factory(**kafka_consumer_params):
         params = {} if kafka_consumer_params is None else kafka_consumer_params.copy()
         params.setdefault('client_id', 'consumer_%s' % (request.node.name,))
+        params.setdefault('auto_offset_reset', 'earliest')
         _consumer[0] = next(kafka_broker.get_consumers(cnt=1, topics=[topic], **params))
         return _consumer[0]
 
--- a/test/test_consumer_integration.py
+++ b/test/test_consumer_integration.py
@@ -1,6 +1,9 @@
 import logging
 import os
 import time
+from mock import patch
+import pytest
+import kafka.codec
 
 from six.moves import xrange
 import six
@@ -13,7 +16,7 @@ from kafka import (
 from kafka.consumer.base import MAX_FETCH_BUFFER_SIZE_BYTES
 from kafka.errors import (
     ConsumerFetchSizeTooSmall, OffsetOutOfRangeError, UnsupportedVersionError,
-    KafkaTimeoutError
+    KafkaTimeoutError, UnsupportedCodecError
 )
 from kafka.structs import (
     ProduceRequestPayload, TopicPartition, OffsetAndTimestamp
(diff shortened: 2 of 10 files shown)


test/conftest.py
```diff
@@ -73,6 +72,7 @@ def kafka_consumer_factory(kafka_broker, topic, request):
     def factory(**kafka_consumer_params):
         params = {} if kafka_consumer_params is None else kafka_consumer_params.copy()
         params.setdefault('client_id', 'consumer_%s' % (request.node.name,))
+        params.setdefault('auto_offset_reset', 'earliest')
```
jeffwidman: do we really want this? the default in the consumers is normally `'latest'` and I'm a little concerned that this sets a default for test purposes that does not match normal behavior.

tvoinarovskyi: It gave me problems, at least. In tests, you set up by sending messages before consuming. So having a `latest` setting require you to connect before set up, which is not what you want in several places.

tvoinarovskyi:
@jeffwidman Valid point actually. I do like it symmetrical, but really quite confusing to have a code, that should raise an error, but some assert is raised instead.

test/test_consumer_integration.py
```diff
@@ -47,6 +51,23 @@ def test_kafka_consumer(simple_client, topic, kafka_consumer_factory):
     kafka_consumer.close()
 
 
+def test_kafka_consumer_unsupported_encoding(
```
dpkp: please also add a skip if no version is set -- see #1453

test/conftest.py
jeffwidman: Understood.

For my in-house tests of our internal wrapper, I did the connect-then-send messages and while slightly more verbose, it hasn't been too onerous.

That said, I'm fine with going this route instead.

dpkp:
Looks like this one needs conflicts resolved (sorry!)

dpkp:
Merged manually

PR #1448, opened 2018-03-18
Added a test for MemoryRecordsBuilder and fixed some problems there

### LINKED ISSUE #1442: Producer: AttributeError - 'tuple' object has no attribute 'size'
I've had an `AttributeError` - `'tuple' object has no attribute 'size'` raised from producer.send in method `ProducerBatch.try_append`. I won't post a stack trace as I already located the primary issue:

https://github.com/dpkp/kafka-python/blob/master/kafka/record/memory_records.py#L131

`MemoryRecordsBuilder.append` returns `None, 0` in case `self._closed`, but the code using this function expects the return value to be either `None` or some metadata object. Also the documentation of the method falsely claims the return type is `(int, int)`, it all probably stems in unfinished refactoring.

Furthermore, I'm scared what will happen when that bug is fixed, because if it properly returns None and it gets to the following code, it will raise `Exception()`, which just sounds incredibly wrong, I don't think it's handled by the library.
https://github.com/dpkp/kafka-python/blob/master/kafka/producer/record_accumulator.py#L266

I'm afraid I don't have the time to make a PR with tests and also probably wouldn't have proper understanding of what I'm doing anyway.

tvoinarovskyi:
Fixes #1442

### DIFF
--- a/kafka/record/__init__.py
+++ b/kafka/record/__init__.py
@@ -1,3 +1,3 @@
-from kafka.record.memory_records import MemoryRecords
+from kafka.record.memory_records import MemoryRecords, MemoryRecordsBuilder
 
-__all__ = ["MemoryRecords"]
+__all__ = ["MemoryRecords", "MemoryRecordsBuilder"]
--- a/kafka/record/memory_records.py
+++ b/kafka/record/memory_records.py
@@ -135,7 +135,7 @@ class MemoryRecordsBuilder(object):
             (int, int): checksum and bytes written
         """
         if self._closed:
-            return None, 0
+            return None
 
         offset = self._next_offset
         metadata = self._builder.append(offset, timestamp, key, value, headers)
@@ -166,7 +166,7 @@ class MemoryRecordsBuilder(object):
 
     def compression_rate(self):
         assert self._closed
-        return self.size_in_bytes() / self._bytes_written
+        return self.size_in_bytes() / float(self._bytes_written)
 
     def is_full(self):
         if self._closed:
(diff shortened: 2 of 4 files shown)


kafka/record/__init__.py
```diff
@@ -1,3 +1,3 @@
-from kafka.record.memory_records import MemoryRecords
+from kafka.record.memory_records import MemoryRecords, MemoryRecordsBuilder
 
-__all__ = ["MemoryRecords"]
+__all__ = ["MemoryRecords", "MemoryRecordsBuilder"]
```
jeffwidman: I'm not very familiar with this code, but I don't understand why this import is needed? Reading the bug description, it looks like the problems were all due to the code fixed in `kafka/record/memory_records.py` below... If the import were just forgotten, I would have thought it would have thrown problems before?

tvoinarovskyi: No particular reason, just used in tests. We can remove it, yea

kafka/record/memory_records.py
```diff
@@ -166,7 +166,7 @@ def size_in_bytes(self):
 
     def compression_rate(self):
         assert self._closed
-        return self.size_in_bytes() / self._bytes_written
+        return self.size_in_bytes() / float(self._bytes_written)
```
dpkp: This is only required for python2; python3 division returns a float by default. I think a cleaner way to handle this compatibility is with `__future__`:
```
from __future__ import division
```

With that you shouldn't need to cast to float

kafka/record/__init__.py
jeffwidman: I'm fine leaving it in if you think it should be there, I just didn't understand if there was something I was missing.

kafka/record/memory_records.py
```diff
@@ -135,7 +135,7 @@ def append(self, timestamp, key, value, headers=[]):
             (int, int): checksum and bytes written
```
petee-d: Return type in docstring is incorrect.

dpkp:
Merged manually

Issue #1451, opened 2018-03-22
Python 3.7 support


mikmatko:
**```async```** is a reserved keyword in Python 3.7.
**```async```** is used a bit in kafka-python, so this currently breaks Python 3.7 compatibility. See e.g. ```producer/base.py```

Using something like **```async_```** would be fine though, so quickest way to resolve this would be just to rename a few things from **```async```** to **```async_```** or something like that.

Thanks!

dpkp:
This requires a breaking change to the `SimpleProducer` interface (renaming the public `async=True` kwarg to something else). So we will need to be careful about how we roll this out.

tvoinarovskyi:
Just use kwargs...

dpkp:
We wont be able to keep the `async` kwarg:
```
Python 3.7.0b2 (default, Mar 22 2018, 10:48:35)
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(async=True):
  File "<stdin>", line 1
    def foo(async=True):
                ^
SyntaxError: invalid syntax
>>>
```

Which means we are going to have to break the current interface...

tvoinarovskyi:
I meant something like:
```
>>> def send(self, **kwargs):
...     if 'async' in kwargs:
...         print(kwargs['async'])
... 
>>> send(1)
>>> send(1, async=True)
True
```

dpkp:
Nice! So passing `async=True` would still work on older pythons. Great call.

dpkp:
Fixed in #1454

ATajadod94:
Issue persists on pip install as of today

IgorBrehm:
> Issue persists on pip install as of today

You need to pip uninstall kafka and then pip install kafka-python
The package named kafka is still with an older version without the fix, but the package kafka-python is up to date...

PR #1454, opened 2018-03-22
Change SimpleProducer to use async_send (async is reserved in py37)

dpkp:
`async` is a reserved word in python 3.7. I've renamed the SimpleProducer async to `async_send` and updated the docs. Applications that pass `async=` will get a deprecation warning in the logs but we'll continue to allow for now by checking **kwargs.

Ref #1451

### DIFF
--- a/docs/simple.rst
+++ b/docs/simple.rst
@@ -49,7 +49,7 @@ Asynchronous Mode
 
     # To send messages asynchronously
     client = SimpleClient('localhost:9092')
-    producer = SimpleProducer(client, async=True)
+    producer = SimpleProducer(client, async_send=True)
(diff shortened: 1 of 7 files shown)


Amr-Abouelleil-bose:
Ha this fix been integrated into a python-kafka release? The documentation page still indicates python-kafka is only compatible with 3.6 and lower.

https://kafka-python.readthedocs.io/en/master/

dpkp:
It has been merged, yes. But we have not successfully configured the travis
test suite to run on python 3.7. Once that happens and we have automated
3.7 testing, then I will feel more comfortable declaring compatibility.

Issue #1461, opened 2018-03-27
Consumer threads can deadlock in connection errors


magaman384:
After running some time my Kafka consumer got stuck in a deadlock. I found following messages in the logs.
```
kafka/conn.py <BrokerConnection node_id=1 host=... port=9092>: Closing connection. CorrelationIdError: Correlation IDs do not match: sent 873966, recv 873967
kafka/client_async.py Node 1 connection failed -- refreshing metadata
kafka/coordinator/base.py Error sending OffsetCommitRequest_v2 to node 1 [CorrelationIdError: Correlation IDs do not match: sent 873966, recv 873967]
```
I have tried connecting with gdb to the stuck consumer, here are the stack traces from threads waiting for locks.
```
#1  0x000000000056aae4 in PyThread_acquire_lock (lock=0x43466c0, waitflag=1) at ../Python/thread_pthread.h:324
324	../Python/thread_pthread.h: No such file or directory.
(gdb) pydown
Undefined command: "pydown".  Try "help".
(gdb) py-bt
#4 Frame 0x441f800, for file /usr/lib/python2.7/threading.py, line 173, in acquire (self=<_RLock(_Verbose__verbose=False, _RLock__owner=139894790764288, _RLock__block=<thread.lock at remote 0x7f3bce3c8d70>, _RLock__count=1) at remote 0x7f3bc8bd3a10>, blocking=1, me=139895040968512)
    rc = self.__block.acquire(blocking)
#8 Frame 0x4d0d880, for file /usr/lib/python2.7/threading.py, line 285, in __enter__ (self=<_Condition(_Condition__lock=<_RLock(_Verbose__verbose=False, _RLock__owner=139894790764288, _RLock__block=<thread.lock at remote 0x7f3bce3c8d70>, _RLock__count=1) at remote 0x7f3bc8bd3a10>, acquire=<instancemethod at remote 0x7f3bcbc32f00>, _is_owned=<instancemethod at remote 0x7f3bcbc32690>, _release_save=<instancemethod at remote 0x7f3bcbc32500>, release=<instancemethod at remote 0x7f3bcbc32460>, _acquire_restore=<instancemethod at remote 0x7f3bcbc329b0>, _Verbose__verbose=False, _Condition__waiters=[]) at remote 0x7f3bc8bd39d0>)
    return self.__lock.__enter__()
[...]

PR #1464, opened 2018-03-28
Always acquire client lock before coordinator lock to avoid deadlocks

dpkp:
This is an attempt to address the deadlock issue described in #1461 . The first option discussed, delaying errback processing until after release of the client lock, would require a lot of work. I've instead tried to change the lock acquisition paths so that the coordinator never attempts to acquire the client lock after it holds the coordinator lock. This is done by either releasing the coordinator lock before calling functions that acquire the client lock, or by preemptively acquiring the client lock before acquiring the coordinator lock.

### DIFF
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -231,20 +231,19 @@ class BaseCoordinator(object):
 
         Returns: the current coordinator id or None if it is unknown
         """
-        with self._lock:
-            if self.coordinator_id is None:
-                return None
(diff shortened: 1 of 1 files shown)


dpkp:
Any comments on this? If not, I'm going to merge. I have been able to reproduce the deadlock locally on master (though not consistently).  With this patch applied I have not seen a deadlock.

dpkp:
Yes, I think you're right!

Issue #1468, opened 2018-04-03
BrokerConnection slow initialization in version >= 1.4.0 when using broker-version auto-detection


romulorosa:
Hello everyone,

I've updated my kafka-python from 1.3.5 to 1.4.0 and could notice that a producer initialization in 1.4.0 is slower than in 1.3.5. 

Using 1.3.5
```
In [1]: from kafka.producer.kafka import KafkaProducer
   ...: import time
   ...: 
   ...: start = time.time()
   ...: p = KafkaProducer(bootstrap_servers='kafka:9092')
   ...: end = time.time()
   ...: print(end - start)
   ...: 
0.12534785270690918
```

Using 1.4.0
```
In [1]: from kafka.producer.kafka import KafkaProducer
   ...: import time
   ...: 
   ...: start = time.time()
   ...: p = KafkaProducer(bootstrap_servers='kafka:9092')
   ...: end = time.time()
   ...: print(end - start)
   ...: 
1.1414685249328613
```

Also I've performed some tests using python profile and the result shows that `poll` method is taking more time to run.

Using 1.3.5

![screenshot from 2018-04-03 16-04-54](https://user-images.githubusercontent.com/15314978/38270273-407962fa-3759-11e8-8b93-1547b8127b13.png)

Using 1.4.0

![screenshot from 2018-04-03 16-04-17](https://user-images.githubusercontent.com/15314978/38270285-48ed1346-3759-11e8-87a3-eeb9db716c92.png)

Does anyone know the reason of this overhead? 

Thanks.

romulorosa:
Hi guys,

I've found the reason of the ovehead. Basically it was introduced by calling `selector.select(1)` in `kafka.conn.BrokerConnection#check_version` (https://github.com/dpkp/kafka-python/blob/master/kafka/conn.py#L954). 
I don't have enough context to know preciselly why we have this call always, eventhough the first attempt to get the responses of api requests was sucessfully completed. So i've promoted some changes and i would like to know if it makes sense. Basically i check if the previous attempt it's OK before call `selector.select(1)`.

https://github.com/dpkp/kafka-python/compare/master...romulorosa:chore-improve-producer-initialization?expand=1

@dpkp Please, can you take a look into these changes?

dpkp:
Thanks -- merged and fixed.

PR #1475, opened 2018-04-15
Improve BrokerConnection initialization

romulorosa:
This should improve the BrokerConnection initialization performance avoiding call `selector.select(1)` without data to fetch in socket.
This PR is related to https://github.com/dpkp/kafka-python/issues/1468


dpkp:
Thanks!!

PR #1490, opened 2018-05-09
Create kafka-producer-csv-to-json example

Obazure:
simple way convert csv-json example (kafka-producer-csv-to-json).
Tested with data from https://www.kaggle.com/sohier/30-years-of-european-solar-generation


jeffwidman:
Thanks, but I'm inclined to pass on this. It doesn't add much to show how to use kafka-python, it primarily shows how to do data-munging in Python. And StackOverflow is probably a better fit for that.

Obazure:
Ok, i got you. thanks for response.

PR #1492, opened 2018-05-10
Stop shadowing `ConnectionError`

jeffwidman:
In Python3, `ConnectionError` is a native exception. So rename our custom one to `KafkaConnectionError` to prevent accidentally shadowing the native one.

Note that there are still valid uses of `ConnectionError` in this code.
They already expect a native Python3 `ConnectionError`, and also already
handle the Python2 compatibility issues.

### DIFF
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -11,7 +11,7 @@ import select
 from kafka.vendor import six
 
 import kafka.errors
-from kafka.errors import (UnknownError, ConnectionError, FailedPayloadsError,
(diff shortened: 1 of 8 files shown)


jeffwidman:
Note that this is a breaking change for anyone who did `from kafka.errors import ConnectionError`.

jeffwidman:
@tvoinarovskyi / @dpkp can one of you review? I don't trust the test suite is executing all these code paths...

tvoinarovskyi:
Sorry for being so late, it's a good PR. 1 moment thou, we don't want to break existing code, so we should make an alias for ConnectionError in errors.py

PR #1495, opened 2018-05-11
Fix #1494: "Consumer fails to reconnect on kafka restart in case of mutiple topics"

### LINKED ISSUE #1494: Consumer fails to reconnect on kafka restart in case of mutiple topics
Steps to reproduce:
1. Run consumer, please see code below
2. Restart kafka
3. Wait for consumer reconnecting to kafka. It fails.

In case of the single topic consumer reconnects successfully.

The reason is update_metadata() function in kafka/cluster.py has the correct code only for a single topic case (lines 208-211). It reconnects fine for multiple topics if I replace the condition to "len(metadata.topics) >= 1" in line 208.

```
   198      def update_metadata(self, metadata):
   199          """Update cluster state given a MetadataResponse.
   200  
   201          Arguments:
   202              metadata (MetadataResponse): broker response to a metadata request
   203  
   204          Returns: None
   205          """
   206          # In the common case where we ask for a single topic and get back an
   207          # error, we should fail the future
   208          if len(metadata.topics) == 1 and metadata.topics[0][0] != 0:
   209              error_code, topic = metadata.topics[0][:2]
   210              error = Errors.for_code(error_code)(topic)
   211              return self.failed_update(error)
   212  
   213          if not metadata.brokers:
   214              log.warning("No broker metadata found in MetadataResponse")
   215  
   216          _new_brokers = {}
```

In case of multiple topics it resets the brockers list to empty list and therefore never updates the meta and brockers list as well.

Sample consumer code:

```
import logging
import sys


logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)


def run_listener():
    logger = logging.getLogger('kafka')
    logger.setLevel(logging.DEBUG)

    consumer = KafkaConsumer(bootstrap_servers='localhost:9092',
                             auto_offset_reset='earliest',
                             group_id="group1",
                             consumer_timeout_ms=1000)
    topics = ['topic1', 'topic2', 'topic3']
    consumer.subscribe(topics)
    while True:
        try:
            for message in consumer:
                print message.value
        except Exception, e:
            print e


run_listener()

```

vasily-novikov:
Fix #1494: "Consumer fails to reconnect on kafka restart in case of mutiple topics"

Prevent `update_metadata()` from moving to unrecoverable state of empty broker list.

### DIFF
--- a/kafka/cluster.py
+++ b/kafka/cluster.py
@@ -212,6 +212,7 @@ class ClusterMetadata(object):
 
         if not metadata.brokers:
             log.warning("No broker metadata found in MetadataResponse")
+            return Errors.NoBrokersAvailable()
 
         _new_brokers = {}
         for broker in metadata.brokers:


dpkp:
Thanks for submitting this. I've modified slightly and added a test in #1506

PR #1506, opened 2018-05-25
Ignore MetadataResponses with empty broker list

dpkp:
Fix for #1494 . The MetadataResponse should be ignored and treated as a failure if the broker list is empty (same as the java client).

### DIFF
--- a/test/test_cluster.py
+++ b/test/test_cluster.py
@@ -1,0 +1,24 @@
+# pylint: skip-file
+from __future__ import absolute_import
+
+import pytest
+
+from kafka.cluster import ClusterMetadata
+from kafka.protocol.metadata import MetadataResponse
+
+import kafka.common as Errors
+
+
+def test_empty_broker_list():
+    cluster = ClusterMetadata()
+    assert len(cluster.brokers()) == 0
+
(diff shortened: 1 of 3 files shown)


test/test_cluster.py
```diff
@@ -0,0 +1,24 @@
+# pylint: skip-file
+from __future__ import absolute_import
+
+import pytest
+
+from kafka.cluster import ClusterMetadata
+from kafka.protocol.metadata import MetadataResponse
+
+import kafka.common as Errors
```
jeffwidman: This is confusing... why import common as Errors? Why not directly import errors as errors or common as common?

dpkp: good catch! I removed that line as it is not needed. But generally re: kafka.common -- that line was copy-pasta from a file that was written a long time ago before kafka.common was split into kafka.errors and kafka.structs . The only reason to keep kafka.common is to prevent breaking user code that was written before the change. But for our library code we should use kafka.errors directly.

jeffwidman: That's what I thought... I finished the library cleanup here: https://github.com/dpkp/kafka-python/pull/1509 so hopefully this won't happen again...

PR #1507, opened 2018-05-25
Improve connection handling when bootstrap list is invalid

### LINKED ISSUE #1473: Invalid bootstrap servers list blocks during startup
```python
>>> kafka.KafkaProducer(bootstrap_servers='can-not-resolve:9092')
# never returns
```

dpkp:
Fixes #1473 . This change drops the DNS retries that I had originally included. This should return us to the prior behavior where attempting to start a kafka-python client with invalid bootstrap metadata (or a failing DNS system) will raise an exception.

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -292,11 +292,7 @@ class BrokerConnection(object):
         # First attempt to perform dns lookup
         # note that the underlying interface, socket.getaddrinfo,
         # has no explicit timeout so we may exceed the user-specified timeout
-        while time.time() < timeout:
-            if self._dns_lookup():
-                break
-        else:
-            return False
+        self._dns_lookup()
 
         # Loop once over all returned dns entries
         selector = None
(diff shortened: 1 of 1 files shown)


jeffwidman:
What do you think about making this configurable?

My day job has a flaky DNS server that's owned by another team and every few hours it will drop some queries... I'd prefer not to have to wrap kafka-python or fork it just to make sure those retries continue to happen. Fixing the root cause here ultimately will require swapping out the DNS server which is a massive project for us, so unlikely to happen anytime soon.

I also understand the perspective of folks who want this to fail loudly and immediately in case they typo'd.

So would it be possible to add a timeout config that defaults to failing quickly, but could be passed a larger value to keep retrying for a period of time?

If you want a PR for this, I'm more than happy to do so, but I wanted to discuss the API first.

svvitale:
Hi @dpkp, thanks for this update.  I'm working through integrating it now, but I'm finding that when I instantiate a `KafkaProducer` with invalid bootstrap information, it does *not* raise an exception.  It returns a valid instance, and I can't tell that anything is wrong until I actually try to send data.  From your description, it sounds like this should be happening earlier.  Which behavior is correct?

PR #1509, opened 2018-05-28
Don't use `kafka.common` internally

jeffwidman:
This finishes the split from `kafka.common` to `kafka.errors`/`kafka.structs`.


kafka/structs.py
```diff
@@ -93,7 +93,3 @@
 # Limit value: int >= 0, 0 means no retries
 RetryOptions = namedtuple("RetryOptions",
     ["limit", "backoff_ms", "retry_on_timeouts"])
-
-
-# Support legacy imports from kafka.common
-from kafka.errors import *
```
jeffwidman: This should no longer be needed as I verified that anywhere that imports `kafka.structs` is not indirectly importing anything from `kafka.errors`

PR #1514, opened 2018-06-05
Include missing import for msgpack

AtomsForPeace:


jeffwidman:
Thanks, but personally I'd rather not have this in the docs at the top level because it implies that all usages of kafka-python require `msgpack`, when that's not true... and if we go this route, we'd need to add `import json` etc that are used in the docs. 

I would be fine if instead right above where `msgpack` is used in the example if you added a comment that said "# requires `import msgpack`" to help clarify that this is a 3p library being used in the example

PR #1540, opened 2018-07-12
Add KafkaAdmin class

llamahunter:
Requires cluster version > 0.10.0.0, and uses new wire protocol
classes to do many things via broker connection that previously
needed to be done directly in zookeeper.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1540)

### DIFF
--- a/kafka/admin/config_resource.py
+++ b/kafka/admin/config_resource.py
@@ -1,0 +1,36 @@
+from __future__ import absolute_import
+
+# enum in stdlib as of py3.4
+try:
+    from enum import IntEnum  # pylint: disable=import-error
+except ImportError:
+    # vendored backport module
+    from kafka.vendor.enum34 import IntEnum
+
+
+class ConfigResourceType(IntEnum):
+    """An enumerated type of config resources"""
+
+    BROKER = 4,
+    TOPIC = 2
+
+
+class ConfigResource(object):
+    """A class for specifying config resources.
+    Arguments:
+        resource_type (ConfigResourceType): the type of kafka resource
+        name (string): The name of the kafka resource
+        configs ({key : value}): A  maps of config keys to values.
+    """
+
+    def __init__(
+            self,
+            resource_type,
+            name,
+            configs=None
+    ):
+        if not isinstance(resource_type, (ConfigResourceType)):
+            resource_type = ConfigResourceType[str(resource_type).upper()] # pylint: disable-msg=unsubscriptable-object
+        self.resource_type = resource_type
+        self.name = name
+        self.configs = configs
--- a/kafka/__init__.py
+++ b/kafka/__init__.py
@@ -18,6 +18,7 @@ except ImportError:
 logging.getLogger(__name__).addHandler(NullHandler())
 
 
+from kafka.admin import KafkaAdmin
 from kafka.consumer import KafkaConsumer
 from kafka.consumer.subscription_state import ConsumerRebalanceListener
 from kafka.producer import KafkaProducer
@@ -46,6 +47,7 @@ class KafkaClient(SimpleClient):
 
 
 __all__ = [
+    'KafkaAdmin',
     'KafkaConsumer', 'KafkaProducer', 'KafkaClient', 'BrokerConnection',
     'SimpleClient', 'SimpleProducer', 'KeyedProducer',
     'RoundRobinPartitioner', 'HashedPartitioner',
--- a/kafka/admin/__init__.py
+++ b/kafka/admin/__init__.py
@@ -1,0 +1,10 @@
+from __future__ import absolute_import
+
+from kafka.admin.config_resource import ConfigResource, ConfigResourceType
+from kafka.admin.kafka import KafkaAdmin
+from kafka.admin.new_topic import NewTopic
+from kafka.admin.new_partitions import NewPartitions
+
+__all__ = [
+    'ConfigResource', 'ConfigResourceType', 'KafkaAdmin', 'NewTopic', 'NewPartitions'
+]
--- a/kafka/admin/kafka.py
+++ b/kafka/admin/kafka.py
@@ -1,0 +1,505 @@
+from __future__ import absolute_import
+
+import copy
+import logging
+import socket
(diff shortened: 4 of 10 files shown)


llamahunter:
Not really a python pro, so please provide feedback on conventions, etc.  Also, not really sure how to do extensive unit testing.  The existing tests seem very integration-centric, and don't actually run successfully for me on my machine.

dpkp:
This is great work! Thanks so much for submitting. My only top-level question is how do you think we should handle api version selection in a multi-version cluster? I.e., during a cluster upgrade. In the current non-admin case we provide a way for users to pass in an explicit "broker version". A more robust approach might be to always check the version of the broker we are sending the request to, but that gets significantly more complicated. Finally, maybe the last option is to advertise in the documentation that it is not safe to use while the cluster is in mixed-version / rolling upgrade state. What do you think?

llamahunter:
Well, I'm using the ApiVersions data structure returned by the broker to select the max version that both the library and the broker supports for each individual API call.  I think hard coding that "0.11.0.2 means XYZ parameters on operations ABC are supported" is error prone, especially when the brokers *already* provide us with exactly what versions of each API call they support.

 I think that if someone is doing an admin operation on a mixed version cluster, they should be sure that they don't use any of the optional parameters that aren't supported by all brokers in the cluster. Tho, I think that most (all?) admin calls go to the cluster controller, so it really only matters what that broker version is at.

llamahunter:
Hmm.  I was fetching the ApiVersions map from whatever happened to be the 'least loaded node', but that's not necessarily the controller.  All the admin calls go to the controller, so really need to check what ApiVersions it supports.  Moved the check_version() call there.

llamahunter:
So, looking for some feedback on whether I should augment the API with return types for the calls.  Right now, it's sort of just delivering back whatever Kafka wire protocol object.  But, it looks like that's only got python attributes for the top level, and subobjects are just tuples or something, making them difficult to use.  Thoughts?  Is there a way to make the subobjects returned by the Schema parser have attributes?  Should I unmarshal into typed python objects?

kafka/admin/config_resource.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import absolute_import
+
+from enum import IntEnum
```
dpkp: We have a pretty strict rule against mandatory external dependencies. I know this is standard in py3, but it isn't in py2. Can we avoid using it? If not, we should vendor the py27 version (see kafka/vendor)

llamahunter: ugh... people are still using python 2?  Isn't it EOL?  Hrmph...  lemme look at it.

jeffwidman: Any progress @llamahunter ?

And yes, I know of multiple companies who heavily rely on this library and still have large codebases in python 2. Most are planning to migrate to python 3, but still in the early stages. I'd guess it'll be another 2-3 years before I'd be okay with saying that we as a core infra library are going to stop supporting python 2 in our releases. I look forward to that day, it's just not quite here yet.

sbreidbach-edr: FWIW, I'd love to see this be part of the project as well.

jeffwidman:
Hey @llamahunter I'd really like to land this. I think it'd pair well with #1196 as of a 2.0 release... 

Do you think you'll have time to look at the `enum` part? 

Personally, I'd probably just vendor it, since it's in the stdlib. That way whenever we do eventually drop python 2 support it'll be an easy change, plus in the meantime we can make additional changes throughout the code to use `enum`.

llamahunter:
Sorry about the delay. Will look into this today. Still would like feedback about the return format. Should there be structured data structures rather than the loosely typed property bags? Is there some better way of using the Schema class to make the properties easier to access?

jeffwidman:
I just added `enum34` in #1604--use it as a fallback: 

    # enum in stdlib as of py3.4
    try:
        import enum  # pylint: disable=import-error
    except ImportError:
        # vendored backport module
        from kafka.vendor import enum34 as enum

Re: the schema stuff--good question that I haven't had time to look into yet... @dpkp / @tvoinarovskyi your thoughts?

tox.ini
```diff
@@ -21,6 +21,7 @@ deps =
     crc32c
     py26: unittest2
     decorator
+    py27: enum34
```
jeffwidman: since this is vendored, no need to include this

jeffwidman:
@llamahunter let's just ship as is for now to land the existing changes and then leave figuring out a cleaner design for a later PR... I would not be surprised if we get a few bug reports about this, so I'd rather get it live in the hands of users on master sooner rather than later, even if the interface changes slightly in a later release.

Besides, with the 2.0 release impending once I get #1196, it will be an convenient place to do a breaking change in this interface if needed. 

So if you can rebase and use the vendored `enum34` when `enum` is not available, than I think that's all that is needed to get this merged.

llamahunter:
Ah, sorry if I conflicted.  I rebased and removed the tox.ini change.  Had some weird local problems running the tests, tho.  Hope it works on travis.

jeffwidman:
Also, I think the test failures are my fault from https://github.com/dpkp/kafka-python/pull/1605 so I will get those fixed in a separate PR and then merge this...

Github was down, so travis wasn't running the tests correctly and I didn't realize it til now

llamahunter:
Ok... so, what needs doing here?  Do I need to add
```
# enum in stdlib as of py3.4
try:
    import enum  # pylint: disable=import-error
except ImportError:
    # vendored backport module
    from kafka.vendor import enum34 as enum
```
for my use of IntEnum type?

jeffwidman:
No, I already got that change in there. I was waiting until I had a chance to fix the `pylint` issue which was breaking tests... that is handled in https://github.com/dpkp/kafka-python/pull/1609 / 0d766cf5c16cbc68d42732d044f8faa0af318b0f, so I just rebased and am just waiting for tests to pass and then I will merge this.

jeffwidman:
Tests took forever to get green due to repeated flakiness upstream of travis--failed downloads etc.

PR #1541, opened 2018-07-13
add support for on-demand topic metadata fetch

andyxning:
This PR adds support for on-demand topic metadata fetching. For now, a KafkaClient will fetch metadata for all topics in the initialization. This will have some performance problem when run with a large Kafka cluster.

Adding a `topics` (type `set`) key-value parameter to `KafkaProducer`. This should be all the topics that `send` will be invoked on. For example, if `send` will be invoked on `send('a', ...)` and `send('b', ...)`, then the value for `topics` will be `{'a', 'b'}`.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1541)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -144,6 +144,7 @@ class KafkaClient(object):
 
     DEFAULT_CONFIG = {
         'bootstrap_servers': 'localhost',
+        'bootstrap_topics_filter': set(),
         'client_id': 'kafka-python-' + __version__,
         'request_timeout_ms': 30000,
         'connections_max_idle_ms': 9 * 60 * 1000,
@@ -231,9 +232,15 @@ class KafkaClient(object):
         self._last_bootstrap = time.time()
 
         if self.config['api_version'] is None or self.config['api_version'] < (0, 10):
-            metadata_request = MetadataRequest[0]([])
+            if self.config['bootstrap_topics_filter']:
+                metadata_request = MetadataRequest[0](list(self.config['bootstrap_topics_filter']))
+            else:
+                metadata_request = MetadataRequest[0]([])
         else:
-            metadata_request = MetadataRequest[1](None)
+            if self.config['bootstrap_topics_filter']:
+                metadata_request = MetadataRequest[1](list(self.config['bootstrap_topics_filter']))
+            else:
+                metadata_request = MetadataRequest[1](None)
 
         for host, port, afi in hosts:
             log.debug("Attempting to bootstrap via node at %s:%s", host, port)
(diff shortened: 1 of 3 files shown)


dpkp:
I see the problem you're trying to solve and understand your approach here, but I'm not excited about the additional conceptual complexity here. I think the broker version probing issue could be avoided for now by setting an explicit `api_version` . And I wonder if we should just stop asking for topic metadata during _bootstrap at all. The java client, for example, "bootstraps" locally by creating fake broker objects derived from the bootstrap.servers list. As I understand the official client, it does not actually make a MetadataRequest until a Consumer or Producer action requires it for a specific topic (or all topics when consuming with a pattern match, etc). So I think it might be better to consider following this path. As a first step we could not request any topics when that is supported (MetadataRequest v1+). And then later we might think about how we could change the bootstrap process entirely to not make a MetadataRequest and just populate internal data structures based on the bootstrap_servers data. What do you think?

andyxning:
> And I wonder if we should just stop asking for topic metadata during _bootstrap at all. The java client, for example, "bootstraps" locally by creating fake broker objects derived from the bootstrap.servers list. As I understand the official client, it does not actually make a MetadataRequest until a Consumer or Producer action requires it for a specific topic (or all topics when consuming with a pattern match, etc). So I think it might be better to consider following this path. 

This is a good question. AFAIK, sarama will not request any topic metadata before a real producer send operation. I like the lazy initialization.

>  I think the broker version probing issue could be avoided for now by setting an explicit api_version

This is not a doable way to resolve this problem. This just works around it. In case we operate many different Kafka version clusters, it is hard to let the users set api_version correctly themselves.

> As a first step we could not request any topics when that is supported (MetadataRequest v1+). And then later we might think about how we could change the bootstrap process entirely to not make a MetadataRequest and just populate internal data structures based on the bootstrap_servers data. What do you think?

I agree with this approach. But it may take a long time, imo.

> but I'm not excited about the additional conceptual complexity here.

Yes. i also think it is not so grace. But it is implemented optionally and have no backward incompatibility. I think it can resolve the problem quickly and even later we support to no do any metadata prefetching, it should be alright and no backward incompatibility will be made.

andyxning:
Ping @dpkp

andyxning:
Ping @dpkp

andyxning:
ping @dpkp

dpkp:
Sorry - Ive been quite busy. I hope to take another look over the weekend.

andyxning:
@dpkp Big thanks for your answer and waiting for your comments. :)

andyxning:
Ping @dpkp

andyxning:
Ping @dpkp

andyxning:
Sorry to periodically ping on this PR. @dpkp

andyxning:
Ping @dpkp

andyxning:
Ping @dpkp

andyxning:
Ping @dpkp

dpkp:
If we're going to land this I think we should name the config something more specific. Perhaps `bootstrap_topics_filter` .

andyxning:
@dpkp Done. Please take a look at this.

andyxning:
@dpkp Big thanks!

Issue #1539, opened 2018-07-12
Connecting to secure kafka error


the-sea:
error traceback:

```
Traceback (most recent call last):
  File "/home/apps/pycharm-community-2017.1.5/helpers/pydev/pydevd.py", line 1591, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "/home/apps/pycharm-community-2017.1.5/helpers/pydev/pydevd.py", line 1018, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/home/apps/pycharm-community-2017.1.5/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/projects/pylearn/py_kafka.py", line 40, in <module>
    value_serializer=lambda v: json.dumps(v).encode('utf-8'))
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/producer/kafka.py", line 362, in __init__
    **self.config)
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/client_async.py", line 214, in __init__
    self._bootstrap(collect_hosts(self.config['bootstrap_servers']))
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/client_async.py", line 245, in _bootstrap
    if not bootstrap.connect_blocking():
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/conn.py", line 301, in connect_blocking
    self.connect()
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/conn.py", line 404, in connect
    if self._try_authenticate():
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/conn.py", line 488, in _try_authenticate
    raise ex  # pylint: disable-msg=raising-bad-type
  File "/home/anaconda3/lib/python3.6/site-packages/kafka/conn.py", line 585, in _try_authenticate_gssapi
    output_token = client_ctx.step(received_token)
  File "<decorator-gen-126>", line 2, in step
  File "/home/anaconda3/lib/python3.6/site-packages/gssapi/_utils.py", line 167, in check_last_err
    return func(self, *args, **kwargs)
  File "<decorator-gen-116>", line 2, in step
  File "/home/anaconda3/lib/python3.6/site-packages/gssapi/_utils.py", line 127, in catch_and_return_token
    return func(self, *args, **kwargs)
[...]

PR #1542, opened 2018-07-13
add kerberos domain name config for gssapi sasl mechanism handshake

the-sea:
Fix the related issue #1539

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1542)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -145,6 +145,8 @@ class KafkaClient(object):
             Default: None
         sasl_kerberos_service_name (str): Service name to include in GSSAPI
             sasl mechanism handshake. Default: 'kafka'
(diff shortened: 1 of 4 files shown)


kafka/client_async.py
```diff
@@ -179,6 +181,7 @@ class KafkaClient(object):
         'sasl_plain_username': None,
         'sasl_plain_password': None,
         'sasl_kerberos_service_name': 'kafka',
+        'kerberos_domain_name': None
```
dpkp: Let's maintain the `sasl_` prefix here (so `sasl_kerberos_domain_name`)

the-sea: OK, thank you.

PR #1553, opened 2018-07-19
Verbose logging on messages signalling data loss in producer

sibiryakov:
Recently, I've spend some human days trying to find where in our multi step pipeline we were having a data loss. It turned out one of our producers were generating batches too often resulting to their expiration. But there were no signs of any problems in logs. This PR attempts to fix this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1553)

### DIFF
--- a/kafka/producer/record_accumulator.py
+++ b/kafka/producer/record_accumulator.py
@@ -69,7 +69,8 @@ class ProducerBatch(object):
         return future
 
     def done(self, base_offset=None, timestamp_ms=None, exception=None):
-        log.debug("Produced messages to topic-partition %s with base offset"
+        level = logging.DEBUG if exception is None else logging.WARNING
+        log.log(level, "Produced messages to topic-partition %s with base offset"
                   " %s and error %s.", self.topic_partition, base_offset,
                   exception)  # trace
         if self.produce_future.is_done:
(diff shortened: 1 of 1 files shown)


dpkp:
You can also achieve this in application code by adding error logging to your produce messages via an callback / errback. Would that approach work for you instead?

sibiryakov:
Yes, I can @dpkp but this is awkward and non-obvious. Look from the user prospective: When user is starting to use the library it is assuming that Producer is reliable, and default settings are sufficient. Then he pushes too much small messages and it could take him days to debug where the problem is, depending on the infrastructure complexity. From my experience when messages are silently dropped in Kafka-based sequence of workers its an extremely hard to find out where the problem is, and implies checking the whole sequence.

So my opinion that even there is a way be notified of some problem, it will be rarely used and therefore result to a bad user experience.

sibiryakov:
>But for checking if an application is healthy, the first thing I see most devs do is check the logs and if no errors then they assume everything is fine...
And this is natural. I consider design poor if it requires code to be rewritten to diagnose that there is a data drop happens.

dpkp:
+1 -- thanks!!

Issue #1559, opened 2018-07-30
KafkaConsumer is documented as returning a future, but returns None


ekimekim:
in `KafkaConsumer` in `kafka.consumer.group`, the `commit_async` call is documented as returning a `kafka.future.Future`. While not stated outright now that I read it carefully, the obvious implication is that you can check that future to determine if your commit operation has succeeded or failed.

However, my observed behaviour is that I always get a None returned instead.

The code calls:
```
        future = self._coordinator.commit_offsets_async(
            offsets, callback=callback)
        return future
```

`self._coordinator` is a `kafka.coordinator.consumer.ConsumerCoordinator`. The relevant method `commit_offsets_async` does not return anything (hence, implicitly, it always returns None).

It does, however, call `self._do_commit_offsets_async`, which _does_ return a Future. In another code path it instead gets a future and sets some callbacks on it.

It looks like the intent was for these futures in each code path to be returned from `commit_offsets_async`, but that code is missing.

Could you please either fix this to return the future as implied, or remove the erroneous documentation / make it clear the only way to confirm whether a commit was successful is to add a callback?

ekimekim:
Checking the history, looks like this was introduced by cfddc6bd179e236874e00a899e9349d5c9a54400

PR #1560, opened 2018-07-30
Return future from commit_offsets_async

ekimekim:
The callers of commit_offsets_async seem to imply it should return a future
(eg. future = commit_offsets_async(...)), but it does not. This causes a mismatch
between KafkaConsumer.commit_async()'s documentation and its behaviour.

This fixes that by returning futures that were already available,
we just had to actually return them.

Fixes https://github.com/dpkp/kafka-python/issues/1559

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1560)


dpkp:
Thanks!

Issue #1340, opened 2018-01-11
Fix circular references in KafkaMetrics


dpkp:
It looks like there are several circular references within the metrics implementation. These prevent garbage collection when the consumer or producer object completes. These should be updated to use weakrefs where possible.

kishorenc:
@dpkp We're witnessing memory leaks in `KafkaProducer` via these `KafkaMetric` objects. We periodically `close` the producer object and re-create the producer, and when we do that the garbage collector does not seem to free these metric objects: we were able to verify that with a memory dump. When we don't reuse the producer object without re-initializing it, we see no leaks.

PR #1569, opened 2018-08-10
Clear the metrics dictionary on close (fixes a memory leak).

kishorenc:
Without this, the dictionary is not being garbage collected and is leaking memory. After applying this patch, the memory leak I described [here](https://github.com/dpkp/kafka-python/issues/1340#issuecomment-411712947) was fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1569)


dpkp:
Great catch!

kkxue:
witness (version 1.4.2)

PR #1577, opened 2018-08-17
set socket timeout for the wake_w

Fleurer:
hi Dana,

we're running kafka-python on a SNS site in china, yesterday we've met an issue about the kafka host hangs because of some kernel gotchas, and the application containers hangs after the kafka, thus the site is down before we restarted the kafka.

after some investigation, we found the writing side of the `wake_w` socketpair is blocking, and with no socket timeout. if the sender thread can not execute `wake_r.receive` but got blocked, `wake_w.sendall(b'x')` would block the `producer.send()` method without timeout. we have a config option about `max_block_ms` which works on ` buffer is full or metadata unavailable`, but have not worked on the `wake_w.send` yet.

we'd prefer the `wake_w.send` method fails fast with a timeout error instead of blocking forever on the producer thread, so we can save ourselves from hanging all the containers by a CircuitBreaker.

thank you

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1577)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -151,6 +151,7 @@ class KafkaClient(object):
         'bootstrap_servers': 'localhost',
         'client_id': 'kafka-python-' + __version__,
         'request_timeout_ms': 30000,
+        'max_block_ms': 60000,
         'connections_max_idle_ms': 9 * 60 * 1000,
         'reconnect_backoff_ms': 50,
         'reconnect_backoff_max_ms': 1000,
@@ -198,6 +199,7 @@ class KafkaClient(object):
         self._bootstrap_fails = 0
         self._wake_r, self._wake_w = socket.socketpair()
         self._wake_r.setblocking(False)
+        self._wake_w.settimeout(self.config['max_block_ms'] / 1000.0)
         self._wake_lock = threading.Lock()
 
         self._lock = threading.RLock()
@@ -847,6 +849,9 @@ class KafkaClient(object):
         with self._wake_lock:
             try:
                 self._wake_w.sendall(b'x')
+            except socket.timeout:
+                log.warning('Timeout to send to wakeup socket!')
+                raise Errors.KafkaTimeoutError()
             except socket.error:
                 log.warning('Unable to send to wakeup socket!')
 
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -151,7 +151,7 @@ class KafkaClient(object):
         'bootstrap_servers': 'localhost',
         'client_id': 'kafka-python-' + __version__,
         'request_timeout_ms': 30000,
-        'max_block_ms': 60000,
+        'wakeup_timeout_ms': 3000,
         'connections_max_idle_ms': 9 * 60 * 1000,
         'reconnect_backoff_ms': 50,
         'reconnect_backoff_max_ms': 1000,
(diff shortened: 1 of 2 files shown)


kafka/client_async.py
```diff
@@ -198,6 +199,7 @@ def __init__(self, **configs):
         self._bootstrap_fails = 0
         self._wake_r, self._wake_w = socket.socketpair()
         self._wake_r.setblocking(False)
+        self._wake_w.settimeout(self.config['max_block_ms'] / 1000.0)
```
tvoinarovskyi: I remember we set setblocking(True) in code sometimes. That operation resets timeout if I remember correctly. Could you make sure we don't do it on this socket?

Fleurer: yes, here is all the `setblocking()` calls in 1.4.3:

```
$ git grep 'setblocking'
kafka/client_async.py:        self._wake_r.setblocking(False)
kafka/conn.py:            self._sock.setblocking(False)
kafka/vendor/socketpair.py:        # setblocking(False) that prevents us from having to create a thread.
kafka/vendor/socketpair.py:                csock.setblocking(False)
kafka/vendor/socketpair.py:                csock.setblocking(True)
```

the csock in socketpair.py is setblocking(True) before returning as `self._wake_w` in `KafkaClient.__init__()`. 

```
import logging
import sys

logging.basicConfig(stream=sys.stdout, level=logging.INFO)

KAFKA_CONFIG = {
    'batch_size': 163840,
    'linger_ms': 10000,  # enlarge the linger_ms to run into the batch_is_full condition
    'max_block_ms': 10,
    'reconnect_backoff_ms': 300,
}

producer = KafkaProducer(bootstrap_servers='kafka03.dev.rack.xxxxxxx.com:3390', **KAFKA_CONFIG)
i = 0

while True:
    producer.send('log.franz.test', b'some_message_bytes')
    i += 1
    if i % 1000 == 0:
        print('1000')

```

this is the script to reproduce the hangs, when this script is looping, close the wifi connection. and we will see it hangs, and get this traceback on pressing ^C:

screen shot 2018-08-21 at 10 52 05 am

after setting this socket timeout, this script can quit immediately after `max_block_ms`.

dpkp:
Thanks for the PR! I have a few thoughts here:

(1) We should definitely add a timeout to the `_wake_w` socket! But I'm not sure that we should inherit the `max_block_ms`. I would prefer that we use a different name in the client, like `wakeup_block_ms`, and have the KafkaProducer pass that config to KafkaClient (probably still based on max_block_ms). This will help if/when we need to also support in the KafkaConsumer. I also think the default at the client level can and should be much lower than 60000.

(2) Are you expecting the KafkaProducer to raise an exception if there is a connection error? That is not our current design and I am hesitant to add that without a lot of thought. I would prefer to log a warning, the same as we do on `socket.error`

(3) But so in normal operation the wake socket pair should be drained very frequently and we should not see blocked writes. So the fact that this happens at all is strange. It suggests that the background producer thread is blocked (or crashed) somewhere as well. I would love to figure out where that is and fix it, because that seems like the root cause. In your local testing can you print `sys._current_frames()` when you detect a blocked socket before raising the exception and post the results?

Thanks again for helping improve this library!

jeffwidman:
nudge @dpkp ^^

PR #1591, opened 2018-09-20
Expose ConsumerRebalanceListener in __all__

ben-harack:
This solves a warning in linters like PyCharm, which warns that a line like:
```
from kafka import ConsumerRebalanceListener
```
is actually accessing a protected member of a class or module. Adding it to `__all__` should solve this.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1591)


jeffwidman:
Thanks!

Issue #1598, opened 2018-10-10
gevent monkey patch + ssl exception


genbits:
Trying to establish SSL connection under gevent monkey-patched modules throws exception:

    from gevent import monkey;monkey.patch_all()
    from kafka import KafkaProducer
    producer = KafkaProducer(bootstrap_servers=[...], security_protocol='SSL', ssl_cafile='...', ssl_certfile='...', ssl_keyfile='...', ssl_password='...')

This immediately raises the following exception:

    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/myenv/site-packages/kafka/producer/kafka.py", line 362, in __init__
        **self.config)
      File "/myenv/site-packages/kafka/client_async.py", line 214, in __init__
        self._bootstrap(collect_hosts(self.config['bootstrap_servers']))
      File "/myenv/site-packages/kafka/client_async.py", line 245, in _bootstrap
        if not bootstrap.connect_blocking():
      File "/myenv/site-packages/kafka/conn.py", line 301, in connect_blocking
        self.connect()
      File "/myenv/site-packages/kafka/conn.py", line 392, in connect
        if self._try_handshake():
      File "/myenv/site-packages/kafka/conn.py", line 455, in _try_handshake
        self._sock.do_handshake()
      File "/myenv/site-packages/gevent/_sslgte279.py", line 569, in do_handshake
        self._sslobj.do_handshake()
    AttributeError: 'NoneType' object has no attribute 'do_handshake'

This does not happen when not monkey-patching.

My environment:
macOS 10.13.6
Python 2.7.10 / PyPy 5.8.0
gevent 1.2.2
kafka-python 1.4.3

Issue #1515, opened 2018-06-07
Should we proactively raise an error in producer/consumer if gevent is monkeypatching threading?


jeffwidman:
We use `gevent` at my day job, and I hit a tricky issue where gevent was starving the `KafkaConsumer` threads, so they were timing out of the consumer groups. (There's more to it, but that's the gist.)

As a result, I am adding the following to my internal wrapper code:

    try:
        from gevent.monkey import is_module_patched
        if is_module_patched('thread'):
            raise RuntimeError("""gevent is reporting that the thread/threading
                modules are patched. This can lead to starvation of the
                kafka-python background heartbeat thread, ultimately resulting in
                the consumer being evicted from the kafka consumer group. To
                prevent this, explicitly skip monkey patching thread:
                    from gevent import monkey
                    monkey.patch_all(thread=False)
                """)
    except ImportError:  # they aren't using gevent
        pass

Wondering if we should add this directly to kafka-python as well?

We could lower it to a warning rather than a full-blown exception, in case someone has a valid reason for doing this.

This would probably be useful on `KafkaProducer` in addition to `KafkaConsumer` to minimize the chances of unexpected behavior (and subtle bugs being reported on the issue tracker).

But before I put up a PR, I thought I'd check what folks think of doing this...??

tvoinarovskyi:
Omg, that adds up... How come I never thought of it)

dpkp:
Agree. Strong hypothesis!

jeffwidman:
So what's confusing me currently is that we changed to `from gevent import monkey;         monkey.patch_all(thread=False)` and now I'm getting some rather obtuse errors from gevent:

    Traceback (most recent call last):
    File "/opt/ns/nsenv/lib/python2.7/site-packages/gevent/_socket2.py", line 271, in recv
        sock = self._sock  # keeping the reference so that fd is not closed during waiting
    AttributeError: 'function' object has no attribute '_sock'
    <io at 0x7fb6f81f6c08 fd=5 events=READ> failed with AttributeError

    Traceback (most recent call last):
    File "/usr/lib/python2.7/threading.py", line 65, in _note
        if self.__verbose:
    AttributeError: 'function' object has no attribute '_Verbose__verbose'
    <io at 0x7fb6f857de60 fd=10 events=READ active callback=<bound method PollResult.add_event of <gevent.select.PollResult object at 0x7fb6f4a5d1d0>> args=(gevent.core.EVENTS, 10)> failed with AttributeError

[...]

PR #1599, opened 2018-10-11
moved setblocking(False) after connection is established

genbits:
Fixes issue #1598 
This commit fixes the exception raised when using SSL connection with gevent monkey patched code.
It's done by moving the `setblocking(False)` command to AFTER the connection has been established.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1599)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -78,6 +78,13 @@ except ImportError:
     gssapi = None
     GSSError = None
 
+# needed for setblocking()
+try:
+    from gevent import monkey
+    is_gevent_patched = monkey.is_module_patched('ssl')
+except ImportError:
+    is_gevent_patched = False
+
 
 AFI_NAMES = {
     socket.AF_UNSPEC: "unspecified",
@@ -337,6 +344,7 @@ class BrokerConnection(object):
                 log.debug('%s: setting socket option %s', self, option)
                 self._sock.setsockopt(*option)
 
+            self._sock.setblocking(is_gevent_patched)
             self.state = ConnectionStates.CONNECTING
             if self.config['security_protocol'] in ('SSL', 'SASL_SSL'):
                 self._wrap_ssl()
@@ -401,7 +409,8 @@ class BrokerConnection(object):
                     self.state = ConnectionStates.CONNECTED
                 self.config['state_change_callback'](self)
 
-        self._sock.setblocking(False)
+        if is_gevent_patched:
+            self._sock.setblocking(False)
 
         if self.state is ConnectionStates.AUTHENTICATING:
             assert self.config['security_protocol'] in ('SASL_PLAINTEXT', 'SASL_SSL')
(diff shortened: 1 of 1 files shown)


dpkp:
This will make SSL handshakes blocking for all users, not just those using gevent. Can you come up with something that has less impact on other users?

genbits:
I added a check if the 'ssl' module is gevent monkey-patched.
setblocking(False) will be done after the handshake only in such a case.

kafka/conn.py
```diff
@@ -78,6 +78,13 @@ class SSLWantWriteError(Exception):
     gssapi = None
     GSSError = None
 
+# needed for setblocking()
+try:
+    from gevent import monkey
+    is_gevent_patched = monkey.is_module_patched('ssl')
```
jeffwidman: Can you rename `is_gevent_patched` to clarify that you're only checking whether `ssl` is patched? Maybe `ssl_is_patched_by_gevent`...

Otherwise it's ambiguous whether you're checking for any monkey-patching at all...  Also, FYI monkey-patching threading/sockets can lead to surprising behavior on the background heartbeat thread.

genbits: I renamed the flag as requested.
Regarding heartbeat, I haven't looked at the implementation, but from my experience with gevent monkey-patching, it works quite well with threading and sockets.
Only rarely have I encountered problems with these modules.

jeffwidman: Here's more details on the potential problem with gevent and the background heartbeat: https://github.com/dpkp/kafka-python/issues/1515#issuecomment-397147496

genbits: Well, it does make sense that in a gevent monkey-patched env, if the main thread loop is CPU bound (haven't looked at the code so I just take your word for it) it will not yield to other coroutines, unless yielding is done explicitly.
I think that given the popularity of gevent and its monkey-patching capabilities, it would be fair to handle this situation in the main loop.
Maybe do a check to see if the threading module is patched, and if so, `gevent.sleep(0)` during the main loop.
If this is ok with you, let me know and I'll add that to this PR.

genbits:
After playing with the code and patches, I have decided to take another approach, therefor I will close this pull request.

jeffwidman:
Care to share what your other approach is? And why you moved away from this approach?

genbits:
Sure.
Currently, our component (some kind of a server) uses Redis as our PubSub broker. Due to product request, we need to switch to Kafka.
So I tried to integrate the kafka client in the existing code, but it didn't work due to gevent monkey-patching. After playing with patching this package (like in this PR + some other patches), I didn't want to take a risk that threads will hang/timeout, etc.
So the new approach is to continue publishing to Redis, and to create a simple process that subscribes to the Redis channels, consumes the messages and publishes them to Kafka.
No gevent involved, and no risky patches done :)

srish13:
Need the changes in the PR for my application using gevent monkey.patch_all() and taking to Kafka producer

PR #1602, opened 2018-10-22
Bump vendored `six` to `1.11.0`

jeffwidman:
Bump `six` to `1.11.0`. Most changes do not affect us, but it's good to
stay up to date. Also, we will likely start vendoring `enum34` in which
case https://github.com/benjaminp/six/pull/178 is needed.

Note that this preserves the `kafka-python` customization from https://github.com/dpkp/kafka-python/pull/979
which has been submitted upstream as https://github.com/benjaminp/six/pull/176 but not yet merged.

### DIFF
--- a/kafka/vendor/six.py
+++ b/kafka/vendor/six.py
@@ -1,7 +1,6 @@
 # pylint: skip-file
-"""Utilities for writing code that runs on Python 2 and 3"""
 
-# Copyright (c) 2010-2015 Benjamin Peterson
+# Copyright (c) 2010-2017 Benjamin Peterson
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


kafka/vendor/six.py
```diff
@@ -71,7 +72,9 @@ def __len__(self):
             # 64-bit
             MAXSIZE = int((1 << 63) - 1)
 
-        # Don't del it here, cause with gc disabled this "leaks" to garbage
+        # Don't del it here, cause with gc disabled this "leaks" to garbage.
+        # Note: This is a kafka-python customization, details at:
+        # https://github.com/dpkp/kafka-python/pull/979#discussion_r100403389
```
dpkp: fwiw -- if upstream doesn't want to merge this change, I don't think we should keep it.

dpkp:
to the future!

kafka/vendor/six.py
jeffwidman: Agreed. I left a comment on the upstream PR to bump it so we'll see what happens...

PR #1605, opened 2018-10-22
Vendor `six` consistently

jeffwidman:
Use vendored `six`, and also `six.moves.range` rather than `xrange`

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1605)


jeffwidman:
This broke `pylint`, which I fixed in https://github.com/dpkp/kafka-python/pull/1609 / 0d766cf5c16cbc68d42732d044f8faa0af318b0f

PR #1608, opened 2018-10-23
create round-robin partitioner starting from a random partition

MrTomerLevi:
This partitioner is useful while using multiple producers that produce limited amount of records.

Instead of using RoundRobinPartitioner which always starts from the first partition (causing unbalanced partitions) this one will pick a random partition to start from.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1608)


jeffwidman:
I'm pretty hesitant to take this as the `RoundRobinPartitioner` is a leftover from the `SimpleProducer` days... in fact I'm planning on removing it in #1196.

If you use the new `KafkaProducer`, then the default partitioner it ships with will use a random partition by default if the `key` is not provided... would that achieve what you're looking for?

jeffwidman:
Sorry, but going to reject based on the above comment.... 

We love PRs, but you'll want to do them against the new `Kafka*` interfaces.

MrTomerLevi:
Thanks!

PR #1611, opened 2018-10-25
Stop pinning `pylint`

jeffwidman:
We have many deprecation warnings in the travis logs for things that are fixed in newer versions of `pylint` or `pylint`'s dependencies.

So opening a PR to see if `master` passes on whatever version travis pulls in. Note that `pylint` >= 2.0 does not support python 2, so this will result in different versions of pylint running for python 2 vs python 3. Personally, I am just fine with this.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1611)

### DIFF
--- a/kafka/record/_crc32c.py
+++ b/kafka/record/_crc32c.py
@@ -139,5 +139,7 @@ def crc(data):
 
 if __name__ == "__main__":
     import sys
-    data = sys.stdin.read()
+    # TODO remove the pylint disable once pylint fixes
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Cool, thanks. Those errors were bugging me)

jeffwidman:
The `pylint` team confirmed it's a bug, but I expect will be a little while until they fix it, so I opened https://github.com/dpkp/kafka-python/pull/1612 to workaround it for now...

Issue #1549, opened 2018-07-17
Python3.7 ValueError: attempt to connect already-connected SSLSocket!


dwintergruen:
Try to establish a producer with:

(python3.7 Mac OS high Sierra)

producer = KafkaProducer(bootstrap_servers='SERVER:9093', ssl_password="XXXXX",
                              ssl_cafile="/etc/kafka_clients/ca-cert", security_protocol="SSL",
                              ssl_certfile="/etc/kafka_clients/cert-signed",
                              ssl_keyfile="/etc/kafka_clients/private-key")

always get:

 File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/kafka/producer/kafka.py", line 362, in __init__
    **self.config)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/kafka/client_async.py", line 214, in __init__
    self._bootstrap(collect_hosts(self.config['bootstrap_servers']))
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/kafka/client_async.py", line 245, in _bootstrap
    if not bootstrap.connect_blocking():
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/kafka/conn.py", line 301, in connect_blocking
    self.connect()
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/kafka/conn.py", line 354, in connect
    ret = self._sock.connect_ex(self._sock_addr)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 1146, in connect_ex
    return self._real_connect(addr, True)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 1118, in _real_connect
    raise ValueError("attempt to connect already-connected SSLSocket!")
ValueError: attempt to connect already-connected SSLSocket!

freeyoung:
I'm getting exactly the same exception. Python 3.6 & 3.5 work prefectly.

bunnyc1986:
I am getting the same error. Python 3.7 as well.

[...]

PR #1614, opened 2018-10-27
Add python 3.7 support

jeffwidman:
Add Python 3.7 to the tests.

Document 3.7 support on PyPi.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1614)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ cache:
     - servers/
 
 before_install:
+    - source travis_java_install.sh
     - ./build_integration.sh
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
   apt:
     packages:
       - libsnappy-dev
+      - openjdk-8-jdk
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+dist: xenial    # required for Python >= 3.7 (travis-ci/travis-ci#9069)
+sudo: required  # required for Python >= 3.7 (travis-ci/travis-ci#9069)
+
 language: python
 
 python:
@@ -5,20 +8,8 @@ python:
     - 3.4
     - 3.5
     - 3.6
-    # - 3.7  # TODO uncomment once Travis supports 3.7 w/o the `matrix` workaround used below
+    - 3.7
     - pypy
-
-# TODO: workaround pulled from:
-# https://github.com/travis-ci/travis-ci/issues/9815#issue-336465122
-# It only runs a single 3.7 test job against one version of Kafka but
-# that's good enough for now since we test all the other permutations,
-# especially since this workaround should be removed in the near future.
-# Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
-matrix:
-  include:
-    - python: 3.7
-      dist: xenial
-      sudo: true
 
 env:
     - KAFKA_VERSION=0.8.2.2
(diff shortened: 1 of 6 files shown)


jeffwidman:
This passes locally, but seems like it's always timing out on Travis, probably due to having to use the VM builds rather than the docker container builds for python 3.7. This is what `sudo: yes` is implicitly doing here:
https://github.com/dpkp/kafka-python/pull/1614/files#diff-354f30a63fb0907d4ad57269548329e3R21

jeffwidman:
See also https://github.com/dpkp/kafka-python/issues/1549, this seems to be a valid issue but doesn't appear until 3.7.

jeffwidman:
> I feel like we have a bit too many jobs for our test runners.

And we need to add the 2.X Kafka brokers, so it will only get worse... I have thought several times about cutting down, but I am hesitant to do so on the broker versions as many of those are in production and they do have significant differences... maybe we should drop 3.4/3.5 from the travis matrix as most folks on python 3 tend to stay reasonably up to date

> a more recent Kafka version (>=0.11), that may (maybe) help with timeouts.

Oh, I hadn't thought of that, good idea

jeffwidman:
> run all brokers for 2.7 and latest python + 1 broker per other python versions.

That sounds reasonable to me. I may not get to it soon, so feel free to do it if you get time (and no worries if you don't).

Also, to put this in context... I typically look at the PR and then decide whether / which tests actually matter... ie, if it's straightforward and just needs to be tested against all brokers, then if some random python 3 version times out, I will often immediately merge.

So IMO it's not too bad to have exhaustive test matrix as long as we're not too hung up on retrying builds that timed out just to make them green if we're comfortable making the judgement call that the scope of the PR doesn't create an edge case that matters for it.

dpkp:
Test failure seems like there's a different java setup for this new test configuration and it is preventing Zookeeper (maybe also Kafka) from starting.

```
Error: A fatal exception has occurred. Program will exit.Unrecognized VM option 'UseParNewGC'
Error: Could not create the Java Virtual Machine.
```

jeffwidman:
> Test failure seems like there's a different java setup for this new test configuration and it is preventing Zookeeper (maybe also Kafka) from starting.

That's probably because of the Travis workaround: https://github.com/dpkp/kafka-python/pull/1614/files#diff-354f30a63fb0907d4ad57269548329e3R11

I'm glad it at least got that far, IIRC (it's been a little while) originally I kept seeing timeouts before even getting to that point.

It'd probably be better to switch to the `G1GC` for all the Java builds that support it. But this isn't something I am focused on right now... first I want to get #1193 and then fix some bugs with the new `KafkaAdmin`... hopefully have some PRs inbound there in the next few days.

cclauss:
Travis xenial is now official but not the default: https://blog.travis-ci.com/2018-11-08-xenial-release

That makes the workaround, less of a hack so why not try putting these lines at the top of the .travis.yml file:
```
dist: xenial    # required for Python >= 3.7 (travis-ci/travis-ci#9069)
sudo: required  # required for Python >= 3.7 (travis-ci/travis-ci#9069)
```

dpkp:
I think we need to continue running tests w/ Java 8 (there are some strange failures on the older kafka versions when trying to run w/ Java 11). Unfortunately I haven't figured out how to get jdk8 installed with the new travis xenial image. I opened a forum post here: https://travis-ci.community/t/how-to-use-java8-in-a-python-non-java-project-on-xenial/1823

Issue #1555, opened 2018-07-20
Speed improvements for pack/unpack functions in protocol/types.py


billyevans:
I noticed that pack/unpack functions from https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/types.py might be slightly improved. I made pre-compilation for them. It gives about 10% better performance compared to the current implementation.
Consumption of 100msg:
```
239884    0.187    0.000    0.287    0.000 types.py:18(_unpack) # new version
239884    0.192    0.000    0.323    0.000 types.py:17(_unpack)
```
I also made some profiling for producers/consumers. It gives about 1-1.5% time savings.
My changes https://github.com/billyevans/kafka-python/commit/99bf02113bdbddcb377796f9e5dd470c4de5f194

Should I make a PR?

jeffwidman:
I am not super familiar with the `struct` interfaces, but at first glance, this looks like a straightforward change with an obvious performance win. 

So yes, I'd appreciate a PR @billyevans

tvoinarovskyi:
Seems nice to me

jeffwidman:
I created the PR and cherry-picked the commit in https://github.com/dpkp/kafka-python/pull/1619 so that we don't forget to add this.

PR #1619, opened 2018-10-29
Pre-compile pack/unpack function calls

jeffwidman:
PR that pulls in @billyevans suggested performance improvement from #1555:

I noticed that pack/unpack functions from
https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/types.py
might be slightly improved. I made pre-compilation for them. It gives
about 10% better performance compared to the current implementation.

Consumption of 100msg:
```
239884    0.187    0.000    0.287    0.000 types.py:18(_unpack) # new version
239884    0.192    0.000    0.323    0.000 types.py:17(_unpack)
```

I also made some profiling for producers/consumers. It gives about
1-1.5% time savings.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1619)

### DIFF
--- a/kafka/protocol/types.py
+++ b/kafka/protocol/types.py
@@ -1,13 +1,14 @@
 from __future__ import absolute_import
 
-from struct import pack, unpack, error
+import struct
+from struct import error
 
 from kafka.protocol.abstract import AbstractType
 
 
 def _pack(f, value):
     try:
-        return pack(f, value)
+        return f(value)
     except error as e:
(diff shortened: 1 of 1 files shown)


jeffwidman:
@billyevans can you confirm that you're okay with us pulling in this patch? I don't want to merge it without your permission.

billyevans:
@jeffwidman yes, I am totally fine.

PR #1634, opened 2018-11-13
Be explicit with tuples for %s formatting

### LINKED ISSUE #1633: `BrokerMetadata` throwing TypeError using old-style string conversion
I was accidentally passing a `BrokerMetada` object rather than a `node_id`, but debugging was more cryptic than it needed to be:

```/Users/jeffwidman/Code/kafka-python/kafka/client_async.py in _maybe_connect(self, node_id)
    354
    355             if conn is None:
--> 356                 assert broker, 'Broker id %s not in current metadata' % node_id
    357
    358                 log.debug("Initiating connection to node %s at %s:%s",

TypeError: not all arguments converted during string formatting
```

Looks like the old-style `%s` is having problems but using `format()` doesn't... 

```ipdb> node_id
BrokerMetadata(nodeId=3, host=u'kafka03.stg.local', port=9092, rack=None)
ipdb> str(node_id)
"BrokerMetadata(nodeId=3, host=u'kafka03.stg.local', port=9092, rack=None)"
ipdb> assert False, "%s" % node_id
*** TypeError: not all arguments converted during string formatting
ipdb> assert False, "{}".format(node_id)
*** AssertionError: BrokerMetadata(nodeId=3, host=u'kafka03.stg.local', port=9092, rack=None)
```

jeffwidman:
Fix #1633

I think I got them all... I would appreciate another pair of eyes to review that I didn't accidentally make mistakes on this.

This is a bit hideous, and definitely overkill, but it's faster than trying to debug `TypeError`s in production...

Alternatively, could switch these to `.format()`. 

Unfortunate that the lovely new python 3.7 syntax f-strings are not backport-able.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1634)

### DIFF
--- a/kafka/admin/kafka.py
+++ b/kafka/admin/kafka.py
@@ -166,7 +166,7 @@ class KafkaAdmin(object):
         log.debug("Starting Kafka administration interface")
         extra_configs = set(configs).difference(self.DEFAULT_CONFIG)
         if extra_configs:
-            raise KafkaConfigurationError("Unrecognized configs: %s" % extra_configs)
+            raise KafkaConfigurationError("Unrecognized configs: %s" % (extra_configs,))
 
         self.config = copy.copy(self.DEFAULT_CONFIG)
         self.config.update(configs)
--- a/kafka/client.py
+++ b/kafka/client.py
@@ -174,7 +174,7 @@ class SimpleClient(object):
 
             return decoder_fn(future.value)
 
-        raise KafkaUnavailableError('All servers failed to process request: %s' % hosts)
+        raise KafkaUnavailableError('All servers failed to process request: %s' % (hosts,))
 
     def _payloads_by_broker(self, payloads):
         payloads_by_broker = collections.defaultdict(list)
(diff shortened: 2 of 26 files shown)


kafka/metrics/stats/sampled_stat.py
```diff
@@ -73,7 +73,7 @@ def purge_obsolete_samples(self, config, now):
                 sample.reset(now)
 
     def _advance(self, config, time_ms):
-        self._current = (self._current + 1) % config.samples
+        self._current = (self._current + 1) % (config.samples,)
```
tvoinarovskyi: This one's not formatting

tvoinarovskyi:
1 place failed eye test

dpkp:
Looks fine w/ the one modulo issue taras noted. I think `format` is a better long term solution though.

jeffwidman:
Good catch @tvoinarovskyi, I fixed, can you take one last look?

PR #1635, opened 2018-11-13
Fix list_consumer_groups() to query all brokers

jeffwidman:
Fix list_consumer_groups() to query all brokers

Previously, this only queried the controller. In actuality, the Kafka
protocol requires that the client query all brokers in order to get the
full list of consumer groups.

Note: The Java code (as best I can tell) doesn't allow limiting this to
specific brokers. And on the surface, this makes sense... you typically
don't care about specific brokers.

However, the inverse is true... consumer groups care about knowing their
group coordinator so they don't have to repeatedly query to find it.

In fact, a Kafka broker will only return the groups that it's a
coordinator for. While this is an implementation detail that is not
guaranteed by the upstream broker code, and technically should not be
relied upon, I think it very unlikely to change.

So monitoring scripts that fetch the offsets or describe the consumers
groups of all groups in the cluster can simply issue one call per broker
to identify all the coordinators, rather than having to issue one call
per consumer group. For an ad-hoc script this doesn't matter, but for
a monitoring script that runs every couple of minutes, this can be a big
deal. I know in the situations where I will use this, this matters more
to me than the risk of the interface unexpectedly breaking.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1635)

### DIFF
--- a/kafka/admin/kafka.py
+++ b/kafka/admin/kafka.py
@@ -575,20 +575,54 @@ class KafkaAdmin(object):
         # TODO this is completely broken, as it needs to send to the group coordinator
         # return self._send(request)
 
-    def list_consumer_groups(self):
+    def list_consumer_groups(self, broker_ids=None):
         """List all consumer groups known to the cluster.
 
-        :return: Appropriate version of ListGroupsResponse class
+        This returns a list of Consumer Group tuples. The tuples are
+        composed of the consumer group name and the consumer group protocol
+        type.
+
+        Only consumer groups that store their offsets in Kafka are returned.
+        The protocol type will be an empty string for groups created using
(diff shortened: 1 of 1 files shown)


jeffwidman:
Please don't merge, I want to finish https://github.com/dpkp/kafka-python/pull/1635/files#diff-df0a2cc359b6183ba34610d3de8dfbdbR534 and possibly add some more tests.

jeffwidman:
This should be merged _after_ https://github.com/dpkp/kafka-python/pull/1641/ and the other accompanying PRs... 

When I got further into it, I realized how it all needed to stack together...

jeffwidman:
travis tests passed before the rebase, so rebasing to fix the diffs and now merging w/o waiting for travis

PR #1640, opened 2018-11-17
Fix KafkaAdmin send to controller

jeffwidman:
The `KafkaAdmin` controller send error handling was completely broken.
It also pinned the metadata version unnecessarily.

Additionally, several of the methods were sending to the controller
but either that was unnecessary, or just plain wrong. So updated
following the pattern of the Java Admin client.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1640)

### DIFF
--- a/kafka/admin/kafka.py
+++ b/kafka/admin/kafka.py
@@ -4,9 +4,10 @@ import copy
 import logging
 import socket
 from kafka.client_async import KafkaClient, selectors
+import kafka.errors as Errors
 from kafka.errors import (
-    IncompatibleBrokerVersion, KafkaConfigurationError, KafkaConnectionError,
-    NodeNotReadyError, NotControllerError)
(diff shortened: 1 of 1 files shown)


kafka/admin/kafka.py
```diff
@@ -261,22 +267,34 @@ def _send_request_to_node(self, node, request):
...
+                    # No need to inspect the rest of the errors for
+                    # non-retriable errors because NotControllerError should
+                    # either be thrown for all errors or no errors.
+                    self._refresh_controller_id()
+                    break
+                elif error_type is not Errors.NoError:
+                    raise error_type(
+                        "Request '{}' failed with response '{}'."
+                        .format(request, response))
+            else:
```
dpkp: nice use of for / else !

PR #1652, opened 2018-11-20
Rename KafkaAdmin to KafkaAdminClient

### LINKED ISSUE #1621: Should `KafkaAdmin` be renamed to `KafkaAdminClient`?
Should `KafkaAdmin` be renamed to `KafkaAdminClient`?

The Java world typically calls it the `AdminClient`. And I can visualize a world where `Admin` could refer to a user, an ACL flag governing whether an application has admin permissions, or this client.

Thoughts @dpkp  / @tvoinarovskyi ?

jeffwidman:
Fix #1621 

This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1652)

### DIFF
--- a/docs/apidoc/KafkaAdmin.rst
+++ b/docs/apidoc/KafkaAdmin.rst
@@ -1,5 +1,5 @@
-KafkaAdmin
+KafkaAdminClient
 ===========
 
-.. autoclass:: kafka.admin.KafkaAdmin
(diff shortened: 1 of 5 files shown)


jeffwidman:
Merging this now so that it gets in before the release is cut... that way we never have a public release where `KafkaAdmin` was exposed.

PR #1655, opened 2018-11-20
Differentiate between client and broker version errors

jeffwidman:
do not merge yet--I plan to continue fleshing this out later tonight... it is blocking me using `KafkaAdmin.list_consumer_group_offsets()` against brokers < 0.10.0.0

While both indicate a mismatch between the broker and client, we want to
be able to differentiate between whether the error is thrown by the
client vs the broker.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1655)


jeffwidman:
Closing, as it's low-priority and I will probably never get to this...

PR #1658, opened 2018-11-23
Fix response error checking in KafkaAdminClient send_to_controller

### LINKED ISSUE #1657: Error 'too many values to unpack' in KafkaAdminClient: _send_request_to_controller
Dear kafka-python team,

We start using the API KafkaAdminClient (1.4.4) to create a Topic on a kafka version 1.1 (0.11.0 protocol). We use the following code:
`kac_dst = KafkaAdminClient(bootstrap_servers=bootstrap_servers,client_id='data')`
`logger.info("Creating topic {}".format(topic_name))`
`topic_list = []`
`topic = NewTopic(name=topic_name, num_partitions=2, replication_factor=3)`
`topic_list.append(topic)`
`response = kac_dst.create_topics(new_topics=topic_list, validate_only=True)`

The following error occur while executing create_topics function:

`kafka/admin/client.py", line 398, in create_topics`
`    return self._send_request_to_controller(request)`
`  kafka/admin/client.py", line 335, in _send_request_to_controller`
`    for topic, error_code in getattr(response, "topic_errors", response.topic_error_codes):`
`ValueError: too many values to unpack (expected 2)`

If we set validate_only=False, using one element or few,the same error occur, but the topic/s are created .

Regards,
Pablo.

dpkp:
Fix #1657

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1658)

### DIFF
--- a/kafka/protocol/admin.py
+++ b/kafka/protocol/admin.py
@@ -51,7 +51,7 @@ class CreateTopicsResponse_v0(Response):
     API_KEY = 19
     API_VERSION = 0
     SCHEMA = Schema(
-        ('topic_error_codes', Array(
+        ('topic_errors', Array(
             ('topic', String('utf-8')),
             ('error_code', Int16)))
     )
@@ -61,7 +61,7 @@ class CreateTopicsResponse_v1(Response):
     API_KEY = 19
     API_VERSION = 1
     SCHEMA = Schema(
-        ('topic_error_codes', Array(
+        ('topic_errors', Array(
             ('topic', String('utf-8')),
             ('error_code', Int16),
             ('error_message', String('utf-8'))))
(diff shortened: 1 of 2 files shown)


kafka/protocol/admin.py
```diff
@@ -619,7 +619,7 @@ class CreatePartitionsResponse_v0(Response):
     API_VERSION = 0
     SCHEMA = Schema(
         ('throttle_time_ms', Int32),
-        ('topic_errors', Array(
```
jeffwidman: Java docs calls this field `topic_errors` for the `CreateTopics` and `CreatePartitions` calls. Only `DeleteTopics` returns `topic_error_codes` so that's why I'd handled the difference in the `_send_request_to_controller()`

jeffwidman: I cleaned up the `CreateTopicsResponse` classes to match Java and use `topic_errors`.

jeffwidman:
I amended this commit to clarify the inconsistency between `topic_errors` / `topic_error_codes`, and then force-pushed... @dpkp I hope you don't mind me editing your work, it just seemed simpler than having multiple commits in the repo history.

This should solve it, so I'm going to go ahead and merge.

jeffwidman:
This fix had a bug since `getattr()` always evaluated the fallback default. Fixed in https://github.com/dpkp/kafka-python/commit/1a31be52ec012dfa0ef5079ff9982e01408a8fe1

PR #1252, opened 2017-10-09
Refactor MessageSet and Message into LegacyRecordBatch 

tvoinarovskyi:
To later support v2 message format

PR #1677, opened 2018-12-13
Remove unused `skip_double_compressed_messages`

jeffwidman:
This `skip_double_compressed_messages` flag was added in https://github.com/dpkp/kafka-python/pull/755 in
order to fix https://github.com/dpkp/kafka-python/issues/718.

However, grep'ing through the code, it looks like it this is no longer
used anywhere and doesn't do anything.

So removing it.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1677)

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -55,7 +55,6 @@ class Fetcher(six.Iterator):
         'max_partition_fetch_bytes': 1048576,
         'max_poll_records': sys.maxsize,
         'check_crcs': True,
-        'skip_double_compressed_messages': False,
         'iterator_refetch_records': 1,  # undocumented -- interface may change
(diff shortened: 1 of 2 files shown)


jeffwidman:
Note to self: Lookup what happens if someone creates a consumer with this kwarg... does `KafkaConsumer` ignore unknown kwargs or raise an exception? The exception would make this  backwards incompatible...

dpkp:
I think it is fine to land now. If you want to leave in a default value (undocumented / unused) to avoid the configuration error, that's reasonable. The supporting code for this config was removed in #1252 / commit `fbea5f04bccd28f3aa15a1711548b131504591ac`, so the config is and has been a no-op for a while.

jeffwidman:
Sounds good, I'd rather land it now.

PR #1678, opened 2018-12-14
Bug fix in _send_request_to_controller

onehotvector:
_send_request_to_controller throws errors for different responses. In some case response contains topic_errors and in some topic_error_codes. Fixed as checking for both attributes if none of them is present then only throws an error.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1678)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -332,7 +332,7 @@ class KafkaAdminClient(object):
             tries -= 1
             response = self._send_request_to_node(self._controller_id, request)
             # DeleteTopicsResponse returns topic_error_codes rather than topic_errors
-            for topic, error_code in getattr(response, "topic_errors", response.topic_error_codes):
+            for topic, error_code in map(lambda x: x[:2], getattr(response, "topic_errors", getattr(response, "topic_error_codes", None))):
                 error_type = Errors.for_code(error_code)
                 if tries and error_type is NotControllerError:
                     # No need to inspect the rest of the errors for


Ormod:
Just wanted to chime in that this PR fixes the issue for me, any chance of getting this merged @dpkp

jeffwidman:
I think we fixed this in https://github.com/dpkp/kafka-python/pull/1658. Sorry it took a bit to get that merged in.

Ormod:
@jeffwidman unfortunately this wasn't fixed by #1658 , when creating a topic I now get:

    self.admin_client.create_topics([topic], timeout_ms=self.timeout_ms)
  File "/usr/local/lib/python3.6/site-packages/python3_kafka-1.4.5.dev0-py3.6.egg/kafka/admin/client.py", line 408, in create_topics
  File "/usr/local/lib/python3.6/site-packages/python3_kafka-1.4.5.dev0-py3.6.egg/kafka/admin/client.py", line 340, in _send_request_to_controller
    topic_error_tuples = getattr(response, "topic_errors", response.topic_error_codes)
AttributeError: 'CreateTopicsResponse_v2' object has no attribute 'topic_error_codes'

So the bit that is failing here is the getattr() default setting which actually was working with this PR.

jeffwidman:
@Ormod indeed, I didn't realize `getattr()` always evaluated the default value. That's what I get for throwing up a quick fix w/o testing it.

Fixed in https://github.com/dpkp/kafka-python/commit/1a31be52ec012dfa0ef5079ff9982e01408a8fe1

PR #1688, opened 2019-01-03
Error if connections_max_idle_ms not larger than request_timeout_ms

### LINKED ISSUE #1680: Verify these timeouts are in descending order on KafkaConsumer instantiation
When we create a `KafkaConsumer`, we should validate that the following timeouts are in descending order:
1. `connections_max_idle_ms`
2. `request_timeout_ms`
3. `session_timeout_ms` (or equivalent, depending on pre/post 0.10.1 broker)

If they are not, then raise an error... otherwise the consumer can end up in weird states.

jeffwidman:
`connections_max_idle_ms` must always be larger than `request_timeout_ms`
to avoid potentially unexpected behavior.

Fix #1680.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1688)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -313,11 +313,15 @@ class KafkaConsumer(six.Iterator):
                         new_config, self.config['auto_offset_reset'])
             self.config['auto_offset_reset'] = new_config
 
+        connections_max_idle_ms = self.config['connections_max_idle_ms']
         request_timeout_ms = self.config['request_timeout_ms']
         fetch_max_wait_ms = self.config['fetch_max_wait_ms']
-        if request_timeout_ms <= fetch_max_wait_ms:
-            raise KafkaConfigurationError("Request timeout (%s) must be larger than fetch-max-wait-ms (%s)" %
-                                          (request_timeout_ms, fetch_max_wait_ms))
+        if not (fetch_max_wait_ms < request_timeout_ms < connections_max_idle_ms):
+            raise KafkaConfigurationError(
+                "connections_max_idle_ms ({}) must be larger than "
+                "request_timeout_ms ({}) which must be larger than "
+                "fetch_max_wait_ms ({})."
+                .format(connections_max_idle_ms, request_timeout_ms, fetch_max_wait_ms))
 
         metrics_tags = {'client-id': self.config['client_id']}
         metric_config = MetricConfig(samples=self.config['metrics_num_samples'],
--- a/test/test_consumer.py
+++ b/test/test_consumer.py
@@ -15,11 +15,15 @@ from kafka.structs import (
 class TestKafkaConsumer:
     def test_session_timeout_larger_than_request_timeout_raises(self):
         with pytest.raises(KafkaConfigurationError):
-            KafkaConsumer(bootstrap_servers='localhost:9092', api_version=(0,9), group_id='foo', session_timeout_ms=60000, request_timeout_ms=40000)
+            KafkaConsumer(bootstrap_servers='localhost:9092', api_version=(0, 9), group_id='foo', session_timeout_ms=50000, request_timeout_ms=40000)
 
     def test_fetch_max_wait_larger_than_request_timeout_raises(self):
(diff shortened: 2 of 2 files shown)


dpkp:
Would a warning be more appropriate here? Is this configuration always doomed, or is it possible some power user would want do go against the grain in some circumstance?

kafka/consumer/group.py
```diff
@@ -313,11 +313,15 @@ def __init__(self, *topics, **configs):
                         new_config, self.config['auto_offset_reset'])
             self.config['auto_offset_reset'] = new_config
 
+        connections_max_idle_ms = self.config['connections_max_idle_ms']
         request_timeout_ms = self.config['request_timeout_ms']
         fetch_max_wait_ms = self.config['fetch_max_wait_ms']
-        if request_timeout_ms <= fetch_max_wait_ms:
-            raise KafkaConfigurationError("Request timeout (%s) must be larger than fetch-max-wait-ms (%s)" %
-                                          (request_timeout_ms, fetch_max_wait_ms))
+        if connections_max_idle_ms <= request_timeout_ms <= fetch_max_wait_ms:
```
dpkp: This is a logical AND, meaning that it will only raise the config error if connections_max_idle_ms is <= request_timeout_ms AND request_timeout_ms <= fetch_max_wait_ms. Perhaps:

```
if not (fetch_max_wait_ms < request_timeout_ms < connections_max_idle_ms):
```

jeffwidman: Oh of course, I was thinking it was implementing an `OR` but that makes a lot more sense

jeffwidman:
> Would a warning be more appropriate here? Is this configuration always doomed, or is it possible some power user would want do go against the grain in some circumstance?

It's a great question, and I've thought about it several times. But each time I cannot come up with a valid scenario... unlike Zookeeper, clients don't keep any session/request state that spans multiple connections.

test/test_consumer.py
```diff
@@ -15,11 +15,15 @@
 class TestKafkaConsumer:
     def test_session_timeout_larger_than_request_timeout_raises(self):
         with pytest.raises(KafkaConfigurationError):
-            KafkaConsumer(bootstrap_servers='localhost:9092', api_version=(0,9), group_id='foo', session_timeout_ms=60000, request_timeout_ms=40000)
+            KafkaConsumer(bootstrap_servers='localhost:9092', api_version=(0, 9), group_id='foo', session_timeout_ms=50000, request_timeout_ms=40000)
 
     def test_fetch_max_wait_larger_than_request_timeout_raises(self):
         with pytest.raises(KafkaConfigurationError):
-            KafkaConsumer(bootstrap_servers='localhost:9092', fetch_max_wait_ms=41000, request_timeout_ms=40000)
+            KafkaConsumer(bootstrap_servers='localhost:9092', fetch_max_wait_ms=50000, request_timeout_ms=40000)
```
jeffwidman: this was a simple change from `41000` to `50000` purely for improved readability/consistency with the other tests here, the test passes either way.

test/test_consumer.py
```diff
@@ -15,11 +15,15 @@
...
+
+    def test_request_timeout_larger_than_connections_max_idle_ms_raises(self):
+        with pytest.raises(KafkaConfigurationError):
+            KafkaConsumer(bootstrap_servers='localhost:9092', api_version=(0, 9), request_timeout_ms=50000, connections_max_idle_ms=40000)
```
jeffwidman: I do not understand why this test fails with `NoBrokerFound` if `api_version=(0, 9)` is omitted... or rather, I realize it's a problem with the version probing timing out, but why does `test_fetch_max_wait_larger_than_request_timeout_raises()` right above succeed without needing to specify the version?

dpkp: I think that may have been a side effect of the old AND check. If you drop the api_version here I would expect the test to still pass

jeffwidman: IIRC, I actually did try that exact thing this morning, and it still failed with the `NoBrokerFound` error. That's what confuses me.

PR #1669, opened 2018-12-06
Fix SSL connection testing in Python 3.7

silentben:
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1669)

PR #1661, opened 2018-11-27
Fix SSL connection testing in Python 3.7

seanthegeek:
Closes issue #1549

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1661)

PR #1694, opened 2019-01-08
Catch thrown OSError by python 3.7 when creating a connection

danjo133:
This is needed to use kafka-python with python 3.7.

It builds on #1669 and the solution is a slightly modified version of a comment on #1661

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1694)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -467,6 +467,9 @@ class BrokerConnection(object):
         # old ssl in python2.6 will swallow all SSLErrors here...
         except (SSLWantReadError, SSLWantWriteError):
             pass
+        # python 3.7 throws OSError
+        except OSError:
+            pass
         except (SSLZeroReturnError, ConnectionError, SSLEOFError):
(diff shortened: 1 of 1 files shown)


dpkp:
python bug report here: https://bugs.python.org/issue31122

It seems like `wrap_socket()` could also raise OSError?

danjo133:
So, that gets raised when a socket disconnects? I would say that is another error. This PR is for OSError that is raised when trying to connect normally. I can add a catch for 31122 as well if you want though. How do you want it handled in that case? should it raise SSLEOFError or set errno to something like ENOTCONN ? If I understand the bug report, then if you get hit by that, then your connection has been closed and you need to try to reconnect in some way, which I guess should be up to the user.

danjo133:
@dpkp Do you want me to do anything more/different with this PR?

dpkp:
We've finally got python 3.7 automated testing working on travis, so I think we're good to go. Would love more test coverage wrt SSL connections though.

PR #1712, opened 2019-02-11
Ignore lookup_coordinator result in commit_offsets_async

Faqa:
Fixes an issue where if commit_offsets_async has to lookup the coordinator before actually sending the offset_commit_request, the request will fail with the following error:
```
Error processing callback
  File "/home/runner/pipeline.runfiles/kafka_python_git/kafka/future.py", line 79, in _call_backs
    f(value)
TypeError: _do_commit_offsets_async() takes from 2 to 3 positional arguments but 4 were given
```

This is because the current logic directly adds the invocation of _do_commit_offsets_async as a callback to the invocation of lookup_coordinator, but _do_commit_offsets_async has no use for that value, and has no parameter to receive it, resulting in the above error. This PR just passes a wrapper lambda as the actual callback to the invocation of lookup_coordinator.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1712)

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division
 
 import collections
 import copy
+import functools
 import logging
 import time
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
This makes sense to me.

@dpkp / @tvoinarovskyi your thoughts? If we are going to take this, it'd be nice to get it in before the next release.

@Faqa can you rebase/squash to a single commit?

tvoinarovskyi:
Kind of find the partial part a bit hard to read.

dpkp:
Thanks for the catch + bugfix!

PR #1715, opened 2019-02-17
Drop dependency on sphinxcontrib-napoleon

stanislavlevin:
Since 1.3b1 (released Oct 10, 2014) Sphinx has support for NumPy and
Google style docstring support via sphinx.ext.napoleon extension.
The latter is already used, but sphinxcontrib-napoleon requirement
still presents.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1715)


jeffwidman:
Thanks!

Issue #1717, opened 2019-02-19
Python2 "test_legacy_correct_metadata_response" fails on x86 arch


stanislavlevin:
I run tests on several architectures. One of them is x86.
Under Python2 "test_legacy_correct_metadata_response" fails with:
```
========================================== FAILURES ===========================================
__________________________ test_legacy_correct_metadata_response[0] ___________________________
                                                                                
magic = 0                                                                       
                                                                                
    @pytest.mark.parametrize("magic", [0, 1])                                   
    def test_legacy_correct_metadata_response(magic):                           
        builder = LegacyRecordBatchBuilder(                                     
            magic=magic, compression_type=0, batch_size=1024 * 1024)            
        meta = builder.append(                                                  
            0, timestamp=9999999, key=b"test", value=b"Super")                  
                                                                                
        assert meta.offset == 0                                                 
        assert meta.timestamp == (9999999 if magic else -1)                     
        assert meta.crc == (-2095076219 if magic else 278251978) & 0xffffffff   
>       assert repr(meta) == (                                                  
            "LegacyRecordMetadata(offset=0, crc={}, size={}, "                  
            "timestamp={})".format(meta.crc, meta.size, meta.timestamp)         
        )                                                                       
E       AssertionError: assert 'LegacyRecord...timestamp=-1)' == 'LegacyRecordM...timestamp=-1)'
E         Skipping 34 identical leading characters in diff, use -v to show      
E         - =278251978L, size=35, timestamp=-1)                                 
E         ?           -                                                         
E         + =278251978, size=35, timestamp=-1)
```

PR #1718, opened 2019-02-19
Fix test_legacy_correct_metadata_response on x86 arch

stanislavlevin:
The problem is that the type of required operation result is
"long".

```
>>> type(278251978 & 0xffffffff)
<type 'long'>
```

However, by default "format" method uses __format__():

```
>>> (278251978 & 0xffffffff).__format__('')
'278251978'
```

So, let's compare things using the same engine:

```
>>> "{!r}".format(278251978 & 0xffffffff)
'278251978L'
```

Fixes: https://github.com/dpkp/kafka-python/issues/1717

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1718)


tvoinarovskyi:
Cool, thanks

PR #1720, opened 2019-02-26
Update error code 13 as per spec

### LINKED ISSUE #1719: Getting StaleLeaderEpochCodeError while committing the offset
While committing the offsets, we are sometimes getting the following exception:
```
  File "/usr/local/lib/python2.7/dist-packages/mdl_kafka_shared/core/core_kafka_consumer.py", line 246, in commit_offsets_on_batch_completion
    self.commit()
  File "/usr/local/lib/python2.7/dist-packages/kafka/consumer/group.py", line 515, in commit
    self._coordinator.commit_offsets_sync(offsets)
  File "/usr/local/lib/python2.7/dist-packages/kafka/coordinator/consumer.py", line 513, in commit_offsets_sync
    raise future.exception # pylint: disable-msg=raising-bad-type
kafka.errors.StaleLeaderEpochCodeError: [Error 13] StaleLeaderEpochCodeError
```

The code here mentions a TODO asking if this particular exception is deprecated:
https://github.com/dpkp/kafka-python/blob/f3105a434f3bd2fb3f8899e4861e187e786b03da/kafka/errors.py#L213

**Question:**
- How do we handle this exception? 
In case of `CommitFailedError`,  we know that it is ok to drop the message because this message will be picked up again by the consumer which now owns the partition to which this message originally belongs. But could you please explain what is the reason for getting `StaleLeaderEpochCodeError` and how we should handle it? We could not find any documentation or explanation about this exception.

- Is this exception `StaleLeaderEpochCodeError` deprecated?
If yes, then how do you think we are hitting it? We are using kafka-python 1.4.3

arerlend:
https://kafka.apache.org/protocol.html#protocol_error_codes

Fixes #1719 

@jamoham

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1720)

### DIFF
--- a/kafka/errors.py
+++ b/kafka/errors.py
@@ -209,11 +209,12 @@ class OffsetMetadataTooLargeError(BrokerResponseError):
     description = ('If you specify a string larger than configured maximum for'
                    ' offset metadata.')
 
-
-# TODO is this deprecated? https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes
-class StaleLeaderEpochCodeError(BrokerResponseError):
+    
+class NetworkException(BrokerResponseError):
     errno = 13
-    message = 'STALE_LEADER_EPOCH_CODE'
+    message = 'NETWORK_EXCEPTION'
+    description = 'The server disconnected before a response was received.'
+    retriable = True
 
 
 class GroupLoadInProgressError(BrokerResponseError):


dpkp:
I don't believe this is correct. If you check the kafka source you'll find that error code 13 / NETWORK_EXCEPTION is only ever used in the java client code. It is never referenced in the kafka broker (scala) code. Afaik, the latest kafka broker will never return error code 13.

arerlend:
I hear you.  But is it wrong to update the errors to match the protocol?

dpkp:
I don't believe this is a protocol-level error. I.e., we will never see an error-code 13 on the wire that means NetworkException. And so in that context, yes, I think it is wrong to make this change.

Further, we have an issue reported that appears to have an actual error code 13 decoded from the wire protocol. We do need to figure out what that is / what it means. And I'm fairly confident that it is not NetworkException, given that there does not seem to be any code that will return 13 over the wire from the broker.

dpkp:
See: https://github.com/apache/kafka/blob/1.1.1/core/src/main/scala/kafka/common/ErrorMapping.scala#L47

Issue #1580, opened 2018-08-23
NotEnoughReplicasError and NotEnoughReplicasAfterAppendError should be retriable.


varunl:
According to the kafka protocol doc, the errors `NOT_ENOUGH_REPLICAS` and `NOT_ENOUGH_REPLICAS_AFTER_APPEND`, error codes 19 and 20 respectively, should be retriable.

I don't see them as retriable in the code. Is there a reason for that?

varunl:
@dpkp Any insight into this?

dpkp:
likely an oversight. PRs welcome!

PR #1722, opened 2019-02-28
Make NotEnoughReplicasError/NotEnoughReplicasAfterAppendError retriable

le-linh:
According to [kafka docs](https://kafka.apache.org/protocol.html#protocol_error_codes), both are supposed to be retriable errors, so this fix is to resolve that

See issue #1580

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1722)


tvoinarovskyi:
Yea, makes sense.

dpkp:
LGTM -- thanks!

Issue #1701, opened 2019-01-17
Consumer getting stuck with compacted topic


jgardner3:
We're seeing the consumer hang before reaching the end of all of the topic partitions. It varies which partitions it hangs on. After writing to the partitions, some will work and different ones will hang.
We saw the following in the debug logs which might be relevant:

```
16:18:27 | DEBUG Adding fetch request for partition TopicPartition(topic='...', partition=38) at offset 706
16:18:27 | DEBUG Sending FetchRequest to node 132
16:18:27 | DEBUG Sending request FetchRequest_v4(replica_id=-1, max_wait_time=500, min_bytes=1, max_bytes=52428800, isolation_level=0, topics=[(topic='...', partitions=[(partition=38, offset=706, max_bytes=1048576)])])
16:18:27 | DEBUG <BrokerConnection node_id=132 host=...:9092 <connected> [IPv4 ('...', 9092)]> Request 370: FetchRequest_v4(replica_id=-1, max_wait_time=500, min_bytes=1, max_bytes=52428800, isolation_level=0, topics=[(topic='...', partitions=[(partition=38, offset=706, max_bytes=1048576)])])
16:18:27 | DEBUG Received correlation id: 370
16:18:27 | DEBUG Processing response FetchResponse_v4
16:18:27 | DEBUG <BrokerConnection node_id=132 host=...:9092 <connected> [IPv4 ('...', 9092)]> Response 370 (150.505065918 ms): FetchResponse_v4(throttle_time_ms=0, topics=[(topics=u'...', partitions=[(partition=38, error_code=0, highwater_offset=1171, last_stable_offset=-1, aborted_transactions=NULL, message_set='\x00\x00\x00\x00\x00\x00\x02\xc1\x00\x00\x11Z\
16:18:27 | DEBUG Adding fetched record for partition TopicPartition(topic=u'...', partition=38) with offset 706 to buffered record list
16:18:27 | DEBUG Skipping message offset: 705 (expecting 706)
```

Our kafka-python version is 1.4.4 and our broker is version 1.0.1. The topic is configured with:

```
partitions: 124
replication-factor: 6
segment.bytes=1048576
min.cleanable.dirty.ratio=.01
delete.retention.ms=5000
retention.ms=86400000
cleanup.policy=compact
```

We've experimented with setting various config options for the consumer as well, particularly `max_partition_fetch_bytes` setting it up to `1024*1024*1024`, however, it still would get stuck in the same offset for the same partitions.

[...]

PR #1724, opened 2019-03-01
1701 use last offset from fetch v4 if available

keithks:
We can reproduce #1701 locally 

This PR fixes the issue where sometimes a consumer is stuck reading past certain offsets from a compacted topic if the last few messages from a partition have been removed by compaction and the server supports fetch V4 or above.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1724)


dpkp:
@tvoinarovskyi thoughts on this?  (lgtm)

tvoinarovskyi:
@dpkp we can merge to mitigate the problem. I wanted to make it same as Java's approuch, but I can do that for the release after.

dpkp:
Sounds good to me -- thanks for the PR!

keithks:
Thanks all.  Wasn't too familiar with the code so happy to have it refactored later -- just needed something to work for right now.

Issue #1728, opened 2019-03-05
Consumer intermittently hangs when committing offsets


jeffwidman:
There is some sort of deadlock that we are intermittently hitting within `kafka-python` when our applications are calling `commit()`. The consumer will drop out of the group w/o the process actually dying and the only fix is to restart the process.

This is hurting us badly, we are having to restart multiple applications every six hours. However, I have not yet been able to reliably reproduce what's happening. Here is what I have so far.

The code is wrapped in a thin-layer, so the pseudo code/configs may be slightly named differently, but it should be relatively clear what's happening.

Kafka Broker Version: `1.0.1`

`kafka-python` Version: commit eed59ba3b3c8800859572db046f36b5d8bd66487 (this is newer than 1.4.4, as I wanted to be sure we picked up the latest deadlock fix from a4f0cb881e8cb71f285ae802aecbf716bfe71d01)

`KafkaConsumer` config:

    {
        "max_poll_records": 5,
        "connections_max_idle_ms":540000,
        "request_timeout_ms":305000,
        "session_timeout_ms":10000,
        "heartbeat_interval_ms":3000,
        "auto_offset_reset": "earliest"
    }

Application pseudo-code:

    while True:
        logging.info("About to call consumer.poll for new messages")
        messages = consumer.poll()
        logging.info("Finished consumer.poll, now process_messages")
        process_messages(messages)
        logging.info("Finished process_messages, now committing new offsets")
        consumer.commit()

I enabled debug logging for `kafka-python`, and noticed a difference between healthy and unhealthy logs.

Here are healthy logs:

    2019-03-05 00:55:35,422     INFO __main__                 ns_consumer.py:0213 NSLog: Finished process_messages, now committing new offsets
    # here the application calls consumer.commit()
    2019-03-05 00:55:35,422    DEBUG    kafka.coordinator.consumer      consumer.py:0616 NSLog: Sending offset-commit request with {TopicPartition(topic=u'secret_topic', partition=54): OffsetAndMetadata(offset=11822203, metadata='')} for group secret_group.ns_consumer to 5
    2019-03-05 00:55:35,422    DEBUG    kafka.protocol.parser           parser.py:0053 NSLog: Sending request OffsetCommitRequest_v2(consumer_group='secret_group.ns_consumer', consumer_group_generation_id=3238, consumer_id=u'secret_id', retention_time=-1, topics=[(topic=u'secret_topic', partitions=[(partition=54, offset=11822203, metadata='')])])
[...]

PR #1733, opened 2019-03-12
Synchronize puts to KafkaConsumer protocol buffer during async sends

dpkp:
If we attempt to maintain a FIFO queue of in-flight requests in the BrokerConnection, we will have to be careful that once a request is added to the _protocol buffer that we also make sure to add the response future to the in-flight-requests queue. Furthermore, we need to make sure that the order of requests added to the _protocol buffer is the same order as the order of the IFR queue. Otherwise, if we do not, then we end up with a mismatch between actual response and the expected response as determined by the IFR queue. Currently we do not synchronize the code that does this, so  two requests could come in at the same time and get placed (1) R1 -> _protocol buffer, (2) R2 -> _procotol buffer, (3) R2 -> IFR list, (4) R1 -> IFR list.

To address this issue I've modified BrokerConnection to use the correlation_id that is already tracked by the _protocol buffer, and switched the in-flight request tracking from a FIFO queue to a simple dict of correlation id => tracking data. We've already pushed the low-level request/response correlation to the _protocol buffer, so I don't think there is any need to maintain a second FIFO queue at the BrokerConnection level. In addition, I added a small lock to BrokerConnection to prevent more than one thread from adding a new request to the _protocol buffer concurrently. This should address the async send synchronization issue and allow the KafkaClient to use BrokerConnection.send() without acquiring an external lock (for all other network access we are still relying on the external KafkaClient lock to synchronize access).

I believe this should fix the out-of-sync errors reported in #1728 

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1733)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -17,6 +17,7 @@ except ImportError:
 import socket
 import struct
 import sys
+import threading
 import time
 
 from kafka.vendor import six
@@ -220,7 +221,6 @@ class BrokerConnection(object):
         self.afi = afi
         self._sock_afi = afi
         self._sock_addr = None
-        self.in_flight_requests = collections.deque()
         self._api_versions = None
 
         self.config = copy.copy(self.DEFAULT_CONFIG)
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Lgtm

kafka/conn.py
```diff
@@ -870,20 +887,18 @@ def _recv(self):
...
             return []
         else:
-            return [resp for (_, resp) in responses]  # drop correlation id
+            return responses
 
     def requests_timed_out(self):
         if self.in_flight_requests:
-            (_, _, oldest_at) = self.in_flight_requests[0]
+            get_timestamp = lambda v: v[1]
+            oldest_at = min(map(get_timestamp,
```
tvoinarovskyi: Hope this will not introduce any speed problems.

jeffwidman:
I've seen no further errors after letting this run overnight/this morning on a fairly busy production host where I was previously consistently observing the #1728 problems.

PR #1739, opened 2019-03-14
Mock dns lookups in test_conn

dpkp:
Small change to avoid doing dns resolution when running local connection tests.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1739)

### DIFF
--- a/test/test_conn.py
+++ b/test/test_conn.py
@@ -17,6 +17,13 @@ import kafka.errors as Errors
 
 
 @pytest.fixture
+def dns_lookup(mocker):
+    return mocker.patch('kafka.conn.dns_lookup',
+                        return_value=[(socket.AF_INET,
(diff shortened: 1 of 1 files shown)


test/test_conn.py
```diff
@@ -25,7 +32,7 @@ def _socket(mocker):
 
 
 @pytest.fixture
-def conn(_socket):
+def conn(_socket, dns_lookup):
```
jeffwidman: Took me a moment to realize this fixture _always_ provides a localhost... it'd be nice if there was a docstring explaining what this fixture is intended for and what it provides.

That's a more general comment about our fixtures though, as I've slowly waded through #1193 it's taken longer than it should have to follow the fixture paths because they aren't documented.

PR #1745, opened 2019-03-19
Fix _generate_bootstrap_brokers, skip dns_lookup

gingercookiemage:
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1745)


dpkp:
Thanks for investigating. Unfortunately this is not a good solution because
it breaks multi-interface hostnames.

But I'm confident I understand the underlying issue and once we have SSL
test fixtures running I will gladly put up a fix.

Issue #1710, opened 2019-02-06
Support OAuthBearer Authentication


psycotica0-shopify:
With [KIP-255 OAuth Authentication via SASL/OAUTHBEARER](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876), Kafka now supports OAuth tokens as a method of authentication.

For us to use it on the brokers requires the library to support it, so I'm creating this issue here in case someone beats us to it (or it's already done somewhere I haven't found)

jeffwidman:
A PR would be very welcome!

shawnsarwar:
Is anyone actively working on this? I'd hate to duplicate effort. 

If not, there's some helpful prior art in [this article](https://medium.com/@jairsjunior/how-to-setup-oauth2-mechanism-to-a-kafka-broker-e42e72839fe), which describes a java implementation. The repo for the example is [available here](https://github.com/jairsjunior/kafka-playground) and would be a useful reference to work against if you don't already have a broker running `OAUTHBEARER`.

psycotica0-shopify:
We aren't _yet_, but we will be shortly, probably, if no one beats us to it (which would be great, TBH)
There's also https://github.com/zendesk/ruby-kafka/pull/710 and https://github.com/Shopify/sarama/pull/1240 if you're looking for other references.

pt2pham:
@shawnsarwar just wondering if you're currently doing work on this or if it would be okay for me to do it!

pt2pham:
I am now working on this!

pt2pham:
This issue can be closed now

dpkp:
Woot!

PR #1750, opened 2019-03-20
Support SASL OAuthBearer Authentication

pt2pham:
This PR adds support for OAuthBearer as a method of SASL Authentication as described in [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876) and https://github.com/dpkp/kafka-python/issues/1710.

When initializing the client, the user must pass in an instance of a class that implements a **token** method, which returns an ID or Access Token. The interface is described in the documentation for every class where `sasl_oauth_token_provider` should be configured. I am open to better ideas of where to store the required interface details! 

**Example:**
```python
class TokenProvider(object):
    def token(self):
        return "some_token_id"

producer = KafkaProducer(
    bootstrap_servers='127.0.0.1:9094',
    ...
    security_protocol='SASL_PLAINTEXT',
    sasl_oauth_token_provider=TokenProvider(),
    sasl_mechanism='OAUTHBEARER'
)
```

Tested with a local Kafka cluster and ID Token, and successfully authenticated with the example client initialization above.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1750)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -133,20 +133,8 @@ class KafkaAdminClient(object):
             Default: None
         sasl_kerberos_service_name (str): Service name to include in GSSAPI
             sasl mechanism handshake. Default: 'kafka'
-        sasl_oauth_token_provider (Object): OAuthBearer token provider instance.
-            THE FOLLOWING INTERFACE MUST BE FULFILLED:
-            (required) #token(): Returns an ID/Access Token to be sent to the Kafka
-                client. The implementation should ensure token reuse so that multiple
-                calls at connect time do not create multiple tokens. The implementation
-                should also periodically refresh the token in order to guarantee
-                that each call returns an unexpired token. A timeout error should
-                be returned after a short period of inactivity so that the
-                broker can log debugging info and retry.
-            (OPTIONAL) #extensions() - Returns a map of key-value pairs that can
-                be sent with the SASL/OAUTHBEARER initial client request. If
-                not provided, the values are ignored. This feature is only available
-                in Kafka >= 2.1.0.
-            Default: None
+        sasl_oauth_token_provider (AbstractTokenProvider): OAuthBearer token provider
+            instance. (See kafka.oauth.abstract). Default: None
 
     """
     DEFAULT_CONFIG = {
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -133,6 +133,20 @@ class KafkaAdminClient(object):
             Default: None
         sasl_kerberos_service_name (str): Service name to include in GSSAPI
             sasl mechanism handshake. Default: 'kafka'
+        sasl_oauth_token_provider (Object): OAuthBearer token provider instance.
+            THE FOLLOWING INTERFACE MUST BE FULFILLED:
+            (required) #token(): Returns an ID/Access Token to be sent to the Kafka
+                client. The implementation should ensure token reuse so that multiple
+                calls at connect time do not create multiple tokens. The implementation
+                should also periodically refresh the token in order to guarantee
+                that each call returns an unexpired token. A timeout error should
+                be returned after a short period of inactivity so that the
+                broker can log debugging info and retry.
+            (OPTIONAL) #extensions() - Returns a map of key-value pairs that can
+                be sent with the SASL/OAUTHBEARER initial client request. If
+                not provided, the values are ignored. This feature is only available
+                in Kafka >= 2.1.0.
+            Default: None
 
     """
     DEFAULT_CONFIG = {
@@ -166,6 +180,7 @@ class KafkaAdminClient(object):
         'sasl_plain_username': None,
         'sasl_plain_password': None,
         'sasl_kerberos_service_name': 'kafka',
+        'sasl_oauth_token_provider': None,
 
         # metrics configs
         'metric_reporters': [],
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -264,7 +264,7 @@ class BrokerConnection(object):
             if self.config['sasl_mechanism'] == 'OAUTHBEARER':
                 token_provider = self.config['sasl_oauth_token_provider']
                 assert token_provider is not None, 'sasl_oauth_token_provider required for OAUTHBEARER sasl'
-                assert isinstance(token_provider, AbstractTokenProvider), 'sasl_oauth_token_provider must implement AbstractTokenProvider'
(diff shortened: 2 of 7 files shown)


kafka/conn.py
```diff
@@ -659,6 +680,51 @@ def _try_authenticate_gssapi(self, future):
...
+        if data != b'\x00\x00\x00\x00':
+            error = Errors.AuthenticationFailedError('Unrecognized response during authentication')
+            return future.failure(error)
+
+        log.info('%s: Authenticated via OAuth', self)
+        return future.success(True)
+
+    def _build_oauth_client_request(self):
+        token_provider = self.config['sasl_oauth_token_provider']
+        return "n,,\x01auth=Bearer {}{}\x01\x01".format(token_provider.token(), self._token_extensions())
```
dpkp: is string `format()` going to guarantee correct encoding here? I might suggest operating on and returning `bytes` instead. E.g., `b''.join([b'n,,\x01auth=Bearer ', token, extensions, b'\x01\x01'])`

kafka/admin/client.py
```diff
@@ -133,6 +133,20 @@ class KafkaAdminClient(object):
             Default: None
         sasl_kerberos_service_name (str): Service name to include in GSSAPI
             sasl mechanism handshake. Default: 'kafka'
+        sasl_oauth_token_provider (Object): OAuthBearer token provider instance.
```
dpkp: Rather than `object`, consider defining an abstract interface class w/ `abc`. See kafka/serializer/abstract.py as an example. Alternately, consider splitting into two separate configs, `sasl_oauth_token` and `sasl_oauth_extensions`, each being simple `callable` type

kafka/conn.py
pt2pham: I convert to bytes and encode the output of this function in `_try_authenticate_oauth` so I believe the encoding will be correct.

kafka/admin/client.py
pt2pham: Oh I didn't know we could specify abstract interfaces! I will do that then in a `kafka/oauth` directory

kafka/conn.py
```diff
@@ -659,6 +668,51 @@ def _try_authenticate_gssapi(self, future):
...
+        Return a string representation of the OPTIONAL key-value pairs that can be sent with an OAUTHBEARER
+        initial request.
+        """
+        token_provider = self.config['sasl_oauth_token_provider']
+
+        # Only run if the #extensions() method is implemented by the clients Token Provider class
+        # Builds up a string separated by \x01 via a dict of key value pairs
+        if callable(getattr(token_provider, "extensions", None)):
+            msg = "\x01".join(["{}={}".format(k, v) for k, v in token_provider.extensions().items()])
+            return "\x01" + msg
```
dpkp: Does this work if the map is empty? It generates a bare `\x01`, but maybe it should be an empty string?

kafka/oauth/abstract.py
```diff
@@ -0,0 +1,38 @@
...
+    calls at connect time do not create multiple tokens. The implementation
+    should also periodically refresh the token in order to guarantee
+    that each call returns an unexpired token. A timeout error should
+    be returned after a short period of inactivity so that the
+    broker can log debugging info and retry.
+
+    Token Providers MUST be implemented from this ABC.
+
+    An optional method that may be implemented if the user chooses is:
+    #extensions() - Returns a map of key-value pairs that can
```
dpkp: Consider including a stub method definition below. Don't decorate with `@abc.abstractmethod` so that it isn't required. But I think it adds more clarity to the class overall.

kafka/oauth/abstract.py
```diff
@@ -0,0 +1,38 @@
...
+    A Token Provider must be used for the SASL OAuthBearer protocol.
+
+    The implementation shsould ensure token reuse so that multiple
+    calls at connect time do not create multiple tokens. The implementation
+    should also periodically refresh the token in order to guarantee
+    that each call returns an unexpired token. A timeout error should
+    be returned after a short period of inactivity so that the
+    broker can log debugging info and retry.
+
+    Token Providers MUST be implemented from this ABC.
```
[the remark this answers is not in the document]
pt2pham: I updated the authenticator configuration to `assert isinstance(token_provider, AbstractTokenProvider), 'sasl_oauth_token_provider must implement AbstractTokenProvider'` instead of just checking for the presence of a `token()` method. I figured I would enforce the usage of this abstract class but both methods would technically work.

kafka/oauth/abstract.py
```diff
@@ -0,0 +1,38 @@
...
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+class AbstractTokenProvider(ABC):
+    """
+    A Token Provider must be used for the SASL OAuthBearer protocol.
+
+    The implementation shsould ensure token reuse so that multiple
+    calls at connect time do not create multiple tokens. The implementation
+    should also periodically refresh the token in order to guarantee
+    that each call returns an unexpired token. A timeout error should
```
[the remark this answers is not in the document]
pt2pham: It's up to the implementer but ideally if the Token Provider pings some OAuth server for a token, if it takes too long it shouldn't go on forever. This was something that I've read/done in other OAuth implementations for different Kafka client libraries.

kafka/conn.py
pt2pham: I can check that the extensions isn't an empty map in the `if` so that it goes to the `else` statement and returns empty string!

kafka/conn.py
```diff
@@ -257,7 +261,10 @@ def __init__(self, host, port, afi, **configs):
             if self.config['sasl_mechanism'] == 'GSSAPI':
                 assert gssapi is not None, 'GSSAPI lib not available'
                 assert self.config['sasl_kerberos_service_name'] is not None, 'sasl_kerberos_service_name required for GSSAPI sasl'
-
+            if self.config['sasl_mechanism'] == 'OAUTHBEARER':
+                token_provider = self.config['sasl_oauth_token_provider']
+                assert token_provider is not None, 'sasl_oauth_token_provider required for OAUTHBEARER sasl'
+                assert isinstance(token_provider, AbstractTokenProvider), 'sasl_oauth_token_provider must implement AbstractTokenProvider'
```
dpkp: I'd actually prefer we drop this -- duck typing should be fine. This is python, after all ;)

dpkp:
Thanks for this!

PR #1752, opened 2019-03-21
kafka-python #1748 Fix race condition in protocol.send_bytes

### LINKED ISSUE #1748: Correlation IDs do not match [1.4.5]
I am trying 1.4.5 consumers with autocommit=off`.
I occassionally see `Correlation IDs do not match` error in the logs, in different scenarios. Sometimes it's OffsetCommitRequest, sometimes HeartbeatRequest.

```
2019-03-20 16:20:18,083 WARNING [kafka.client] Node 2 connection failed -- refreshing metadata
2019-03-20 16:20:18,084 ERROR [kafka.coordinator] Error sending OffsetCommitRequest_v2 to node 2 [CorrelationIdError: Correlation IDs do not match: sent 266569, recv 266570]
2019-03-20 16:20:18,084 WARNING [kafka.coordinator] Marking the coordinator dead (node 2) for group <redacted-group>: CorrelationIdError: Correlation IDs do not match: sent 266569, recv 266570.
2019-03-20 16:20:18,085 ERROR [kafka.coordinator] Error sending HeartbeatRequest_v1 to node 2 [CorrelationIdError: Correlation IDs do not match: sent 266569, recv 266570]
2019-03-20 16:20:18,189 INFO [kafka.cluster] Group coordinator for <redacted-group> is BrokerMetadata(nodeId=2, host='<redacted-node-2>', port=9092, rack=None)
```

#1733 uses map to pair responses with correlate_ids, but only does so in `conn.py`. `protocol/parser.py` still uses deque and this is where the error occurs.

Related to my investigation of #1744, where I am playing with more locking on `in_flight_requests` manipulation and closing in `conn.py`. This causes this race condition to occur even more frequently in `parser.py` (which is not that surprising).
Also related to #1529

isamaru:
Resolves #1748 based on suggestion by @dpkp

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1752)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -790,7 +790,8 @@ class BrokerConnection(object):
         if self.state not in (ConnectionStates.AUTHENTICATING,
                               ConnectionStates.CONNECTED):
             return Errors.NodeNotReadyError(str(self))
-        data = self._protocol.send_bytes()
+        with self._lock:
+            data = self._protocol.send_bytes()
         try:
             # In the future we might manage an internal write buffer
             # and send bytes asynchronously. For now, just block


dpkp:
Thanks!!

Issue #1741, opened 2019-03-15
Cannot bootstrap with SSL using 1.4.5 on python 3.7


gingercookiemage:
After switching to new kafka-python 1.4.5 and upgrading to python 3.7 I keep getting these logs in infinite loop:
> Give up sending metadata request since no node is available

I have noticed that broker list is always empty and node-id=bootstrap . Running the same code with 1.4.4 and python 3.6 worked perfectly.
`KafkaProducer(
        bootstrap_servers='{}:{}'.format(
            config.get('tracing_host'),
            config.get('tracing_port')
        ),
        api_version=(0, 10, 1),
        security_protocol='SSL',
        ssl_check_hostname=True,
        ssl_cafile=config.get('ssl_ca_file'),
        ssl_certfile=config.get('ssl_cert_file'),
        ssl_keyfile=config.get('ssl_key_file')
    )`
Any ideas what to try next?

jeffwidman:
Can you post DEBUG-level logs?

I suspect this is a small bug that crept in somewhere through this change: https://github.com/dpkp/kafka-python/pull/1736

dpkp:
Would also love to get integration tests for ssl brokers... If you have
time / motivation, maybe give a shot at configuring a ssl broker fixture
for our test suite? As of now we do not have automated testing for SSL, so
regressions / bugs do happen from time to time.  (I personally do not use
ssl brokers)

dpkp:
Can you try 1.4.5 w/ python 3.6 and see if you get the same results? My hunch is this is related to ssl changes in python 3.7.

gingercookiemage:
This happens immediately after KafkaProducer tries to initialize. I haven't tried this with the Consumer as I only need to post some messages.

`DEBUG       - Starting the Kafka producer [kafka.producer.kafka:336]
DEBUG       - Added sensor with name connections-closed [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name connections-created [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name select-time [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name io-time [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name bufferpool-wait-time [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name batch-size [kafka.metrics.metrics:156]
DEBUG       - Added sensor with name compression-rate [kafka.metrics.metrics:156]
[...]

PR #1754, opened 2019-03-21
Wrap SSL sockets after connecting

dpkp:
Fix for #1741 . Previously we have been wrapping sockets with an SSLContext before connection, and calling `connect_ex` to do a non-blocking connect, then calling `do_handshake`. But the TCP connect itself does not require the SSLContext, and it appears that in general, and definitely on python3.7, non-blocking connect_ex w/ ssl sockets is not fully supported.

So this PR defers wrap_ssl until after the TCP socket is connected, which should side-step all of the ValueError / OSError issues raised by python3.7 due to poor support for connect_ex. Also it is simpler, which I like.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1754)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -356,14 +356,9 @@ class BrokerConnection(object):
 
             self._sock.setblocking(False)
             self.state = ConnectionStates.CONNECTING
-            if self.config['security_protocol'] in ('SSL', 'SASL_SSL'):
-                self._wrap_ssl()
-            # _wrap_ssl can alter the connection state -- disconnects on failure
-            # so we need to double check that we are still connecting before
-            if self.connecting():
-                self.config['state_change_callback'](self)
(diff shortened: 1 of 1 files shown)


kafka/conn.py
```diff
@@ -486,9 +481,6 @@ def _try_handshake(self):
         # old ssl in python2.6 will swallow all SSLErrors here...
         except (SSLWantReadError, SSLWantWriteError):
```
jeffwidman: Should this also check whether python < 2.7 before `pass`'ing? Just in the off-chance these exceptions get-used later?

dpkp: no, see the import hackery above. this is an old problem. we could try to check exception error codes etc, but unless there's a real user asking for help w/ old python installs I'm not that interested in adding more compatibility layers here.

Plus python 2.7 is EOL very soon...

PR #1746, opened 2019-03-19
DP-238 Test a fix for race condition with closed connection

isamaru:
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1746)

PR #1757, opened 2019-03-21
Fix race conditions with conn.in_flight_requests

### LINKED ISSUE #1744: Worker stuck after "Protocol out of sync"
After switching to 1.4.5 from an older 1.3 version, we see some workers getting stuck with this pattern in the logs:

1. "`Protocol out of sync`" with no previous errors
2. "`Closing connection. KafkaConnectionError: Socket EVENT_READ without in-flight-requests`"
3. "`[kafka.client] Node 2 connection failed -- refreshing metadata`"
4. 5 minutes later, worker starts looping out infinite "`Duplicate close() with error: [Error 7] RequestTimedOutError: Request timed out after 305000 ms`", but with no apparent attempts to reconnect

More detailed (INFO) logs here:
[kafka-python-logs.txt](https://github.com/dpkp/kafka-python/files/2979127/kafka-python-logs.txt)

We found a worker which was stuck like this for 2 days, processing no messages but not failing directly or even rebalancing the group, causing lag on its partition. The broker is running and other workers can connect to that broker in that period.
Note that Node 2 is leader for the partition which the worker is assigned to. Group coordinator is Node 1 which is why heartbeat keeps beating.

This seems to be the same thing as #1728 which wasn't completely fixed.

#1733 was about fixing one possible cause for that error (= avoid it). I believe that in our case the error is legitimate (temporary connection problems to broker).
The real issue is that **the worker is not able to restore itself after this error happens and becomes stuck instead of either giving up and dying or reconnecting**.
I've tried to find code responsible for reconnecting (which doesn't seem to fire) but I don't understand your codebase that well. I will continue investigating, this is important for us.

(We have deployed this on several environments and see this on both 1.0.1 and 2.1 brokers, identified by client as 1.0.0)

isamaru:
Resolves #1744.

Based on previous version and investigations in #1746 which can be discarded.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1757)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -612,7 +612,7 @@ class KafkaClient(object):
             conn = key.data
             processed.add(conn)
 
-            if not conn.in_flight_requests:
+            if not conn.has_in_flight_requests():
                 # if we got an EVENT_READ but there were no in-flight requests, one of
                 # two things has happened:
                 #
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -852,7 +852,8 @@ class BrokerConnection(object):
         # augment respones w/ correlation_id, future, and timestamp
         for i, (correlation_id, response) in enumerate(responses):
             try:
-                (future, timestamp) = self.in_flight_requests.pop(correlation_id)
+                with self._ifr_lock:
+                    (future, timestamp) = self.in_flight_requests.pop(correlation_id)
             except KeyError:
                 if self.disconnected():
                     log.warning('%s: Received response %s after the connection had been closed.',
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -776,18 +776,18 @@ class BrokerConnection(object):
         with self._lock:
             correlation_id = self._protocol.send_request(request)
 
-        log.debug('%s Request %d: %s', self, correlation_id, request)
-        if request.expect_response():
-            with self._ifr_lock:
-                if self.disconnected():
-                    log.debug("%s: Connection already closed.", self)
-                    future.failure(Errors.Cancelled(str(self)))
-                else:
-                    sent_time = time.time()
-                    assert correlation_id not in self.in_flight_requests, 'Correlation ID already in-flight!'
-                    self.in_flight_requests[correlation_id] = (future, sent_time)
-        else:
-            future.success(None)
+            log.debug('%s Request %d: %s', self, correlation_id, request)
+            if request.expect_response():
+                with self._ifr_lock:
+                    if self.disconnected():
+                        log.debug("%s: Connection already closed.", self)
+                        future.failure(Errors.Cancelled(str(self)))
+                    else:
+                        sent_time = time.time()
+                        assert correlation_id not in self.in_flight_requests, 'Correlation ID already in-flight!'
+                        self.in_flight_requests[correlation_id] = (future, sent_time)
+            else:
+                future.success(None)
 
         # Attempt to replicate behavior from prior to introduction of
         # send_pending_requests() / async sends
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -828,17 +828,20 @@ class BrokerConnection(object):
         Return list of (response, future) tuples
         """
         with self._ifr_lock:
-            if not self.connected() and not self.state is ConnectionStates.AUTHENTICATING:
-                log.warning('%s cannot recv: socket not connected', self)
-                # If requests are pending, we should close the socket and
-                # fail all the pending request futures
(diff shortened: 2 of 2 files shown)


kafka/client_async.py
```diff
@@ -612,7 +612,7 @@ def _poll(self, timeout):
             conn = key.data
             processed.add(conn)
 
-            if not conn.in_flight_requests:
+            if not conn.has_in_flight_requests():
```
isamaru: Race condition which causes the "Protocol out of sync" (the root cause) happens on this condition.

dpkp:
I wonder if we should just bite the bullet and make the entire class
threadsafe? It seems likely that we will eventually hit other race
conditions wrt _sock, connect, send, recv, close, etc.

isamaru:
> I wonder if we should just bite the bullet and make the entire class threadsafe? 
That is an end goal, but also looks like a big project. Just putting Rlocks everywhere is probably not going to cut it :laughing:.
In particular, there's quite a lot of future resolving. I'd say those callbacks would need to be put outside locked sections.

isamaru:
I missed one more race condition around IFR which just popped up in logs, easy addition.

dpkp:
I believe there is still a race here that could lead to protocol out of sync:

Thread A (with conn._lock) -> _protocol.send_request(request)
Thread B (with client._lock) -> conn.send_pending_requests(), including network I/O write
Broker -> processes request, sends response
Thread B (with client._lock) -> receives selector.EVENT_READ
Thread B (with conn._ifr_lock) -> checks conn.has_in_flight_requests()

Notice that Thread A has yet placed an entry on its in_flight_requests dict, so Thread B will think this is a protocol out-of-sync error and close the socket.

Granted, this timing is extreme and seems extremely unlikely. And furthermore, your changes I believe would fix the "hanging" problem caused by the race between disconnect + ifr queue.

Nonetheless, I wonder if we should be reusing the existing conn._lock here? I think that would synchronize the protocol buffer, network I/O, and ifr tracking.

isamaru:
@dpkp 
Good catch, I am actually reproducing this one too.
Yes, the solution is to extend `self._lock` until `in_flight_requests` is processed. 

I don't think it would work to use `self._lock` everywhere, particularly in `close`, since it looks like it can get called from inside sections which are already locked (and I'd like to avoid a more expensive reentrant lock).

kafka/conn.py
```diff
@@ -808,24 +819,26 @@ def send_pending_requests(self):
...
-            if self.in_flight_requests:
-                self.close(Errors.KafkaConnectionError('Socket not connected during recv with in-flight-requests'))
-            return ()
+        with self._ifr_lock:
+            if not self.connected() and not self.state is ConnectionStates.AUTHENTICATING:
+                log.warning('%s cannot recv: socket not connected', self)
+                # If requests are pending, we should close the socket and
+                # fail all the pending request futures
+                if self.in_flight_requests:
+                    self.close(Errors.KafkaConnectionError('Socket not connected during recv with in-flight-requests'))
```
dpkp: `close` also acquires the ifr_lock, so this is going to fail unless the lock is reentrant

isamaru: Ah, not another one :(
Thanks!

isamaru:
I hear you, but I am worried about performance if we use RLocks. 

In older Pythons it would be significantly slower, and we have components still stuck to 2.7
https://stackoverflow.com/a/5441992

I'll try from scratch and see if I manage to rewrite this using just a non-reentrant `_lock`, but it will have to include some more substantial changes.

dpkp:
This was great work! But I'm going to close in favor of the other PRs because I think we have a good path forward there.

jeffwidman:
> I am worried about performance if we use RLocks. In older Pythons it would be significantly slower, and we have components still stuck to 2.7

Given how late we are in the python 2.7 lifecycle, I personally think we should be more concerned about getting the semantics correct and not worry too much about the python 2.7 performance.

The momentuum to EOL python 2.7 seems to have really picked up over the last 18 months... We've been experiencing this at my dayjob, seems like every other week a new 3p open source library announces they will quit supporting python 2.7 in 2020.

So I suspect what will happen is that companies/application owners will be faced with a choice (beyond just `kafka-python`) to either migrate to python 3 or simply stop upgrading their external libraries. And if they choose to stop upgrading, then whether we are using normal locks or RLocks won't matter because they'll still be on an old version.

Again, this isn't quite where we are at today, but it's coming very very quickly, so if we go for what's semantically correct, it will make long-term maintenance much easier.

PR #1759, opened 2019-03-22
Improve SSLWantReadError handling when ssl not available

dpkp:
When ssl is not available we should be careful not to accidentally catch all network exceptions. I did not apply the same technique to the `_try_handshake` except clauses because that method is only used when ssl is available + configured.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1759)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -861,9 +861,6 @@ class BrokerConnection(object):
                     return []
                 else:
                     recvd.append(data)
-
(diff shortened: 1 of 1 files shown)


jeffwidman:
Curious why you closed w/o merging? Maybe it was mentioned in another issue or PR, but there's not a paper trail here and I'm not remembering...

Issue #1760, opened 2019-03-22
[1.4.5] KafkaProducer raises KafkaTimeoutError when attempting wakeup()


braedon:
I tried upgrading one of our projects to 1.4.5, hoping to use the fix for consuming compacted topics in https://github.com/dpkp/kafka-python/issues/1701. However, it constantly results in KafkaTimeoutErrors almost immediately.

This only seems to happen when consuming (a compacted section, e.g. from the start of) a compacted topic. Consuming from the non-compacted end of a compacted topic seems to be fine. Similarly, consuming from (the start or end of) a non-compacted topic seems to be fine, though at a much slower speed - dropping from ~6k/s on 1.4.4 to ~2k/s on 1.4.5 in one test.

It seems to be inconsistent about exactly when it breaks - sometimes it doesn't produce anything before failing, sometimes it produces a small amount then fails, sometimes it fails while trying to commit the first time, sometimes it commits the first time and then fails - etc.

I tried upgrading to the current master, hoping the extra lock that was added since 1.4.5 would fix the issue, but it doesn't seem to have made any difference.

I also tried setting the `api_version` to 1.1.0 manually, since it was incorrectly auto-detecting 1.0.0, but that also hasn't helped.

The code I'm running is a little more complicated than a standard consumer/producer, but worked fine on 1.4.4.

The consumer uses a consumer group, but has auto-commit disabled. The producer uses `snappy` compression, and `acks='all'`. Messages are consumed via the `KafkaConsumer` iterator interface. Consumed messages are produced with `KafkaProducer.send()`, without waiting on the futures - they're kept in a buffer. When the buffer is full, the thread will wait on one or more of the produce futures to clear space in the buffer. Periodically, the thread will call `KafkaConsumer.commit(offsets=x)`, where `x` is the appropriate offsets based on the produces that have been confirmed to be completed (their futures cleared from the buffer).

Here's the logs from an example, where it produced a small amount, but failed before committing. While it produced some messages, based on the logs it didn't complete flushing any produce futures from the buffer, nor did it commit any offsets.
```
[2019-03-22 03:46:46,861] kafka.consumer.subscription_state.INFO MainThread Updating subscribed topics to: ['topic.0']
[...]

PR #1765, opened 2019-03-27
Avoid more cases of client wakeup when sending from sender / same thread

braedon:
Extra fixes for https://github.com/dpkp/kafka-python/issues/1760.

Note that I don't pretend to fully understand the threading/socket setup here - just extending the fix from https://github.com/dpkp/kafka-python/pull/1761 to cover more cases, as my code seems to be hitting the `_maybe_refresh_metadata()` path, and I spotted a few others that looked similar.

I've done a couple of long running tests with this fix, and haven't run into any more `KafkaTimeoutError`s on `wakeup()`.

This still only covers producer calls. Is this issue/fix relevant to the consumer as well, or any other users of `KafkaClient`?

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1765)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -517,7 +517,7 @@ class KafkaClient(object):
             Future: resolves to Response struct or Error
         """
         if not self._can_send_request(node_id):
-            self.maybe_connect(node_id)
+            self.maybe_connect(node_id, wakeup=wakeup)
             return Future().failure(Errors.NodeNotReadyError(node_id))
 
         # conn.send will queue the request internally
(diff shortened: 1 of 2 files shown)


jeffwidman-ns:
Thanks. Looks like tests are failing?

braedon:
Some are just due to the extra parameter on calls, but it looks like a lot of the producer tests are (ironically?) failing with `KafkaTimeoutError`s.

As noted above, I'm not familiar enough with the threading and networking stuff going on here to really know how to debug those failures...

jeffwidman:
Oops, accidentally wrote my earlier comment from my work account, so deleted that.

Yeah, the threading/networking stuff here is tricky. @dpkp is the one who best understands, by far.

dpkp:
I think #1767 is a simpler fix for the root issue here. Want to give it a try?

braedon:
Thanks @dpkp, I'll take a look.

dpkp:
Thanks for the PR! I've updated / simplified a bit and fixed tests in #1769 , so I'm going to close this one.

PR #1766, opened 2019-03-27
kafka-python #1744 Attempt 2 at race conditions with IFR

isamaru:
Attempt 2 (after #1757) at fixing #1744 with more aggresive (and reentrant) locking.
I expect this to be less performant, especially on Python 2.x, but it's safer.

@dpkp I tried to avoid using RLock but then just gave up as due to avoiding problems with callbacks (particularly during authorization), the size and complexity of it was quickly getting out of hand.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1766)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -273,7 +273,7 @@ class BrokerConnection(object):
         # per-connection locks to the upstream client, we will use this lock to
         # make sure that access to the protocol buffer is synchronized
         # when sends happen on multiple threads
-        self._lock = threading.Lock()
+        self._lock = threading.RLock()
 
         # the protocol parser instance manages actual tracking of the
         # sequence of in-flight requests to responses, which should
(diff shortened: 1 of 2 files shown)


kafka/conn.py
```diff
@@ -784,43 +789,48 @@ def close(self, error=None):
...
-        self.state = ConnectionStates.DISCONNECTED
-        self._sasl_auth_future = None
-        self._protocol = KafkaProtocol(
-            client_id=self.config['client_id'],
-            api_version=self.config['api_version'])
+        with self._lock:
+            if self.state is ConnectionStates.DISCONNECTED:
+                if error is not None:
+                    log.warning('%s: Duplicate close() with error: %s', self, error)
+                self._fail_ifrs(error)
```
dpkp: What is your thinking on this one? If we take care to always drain here and never add more ifrs while the state is disconnected, can't we assume that ifrs will also always be empty here?

kafka/conn.py
```diff
@@ -784,43 +789,48 @@ def close(self, error=None):
...
+                return
+            log.info('%s: Closing connection. %s', self, error or '')
+            self.state = ConnectionStates.DISCONNECTING
+            self.config['state_change_callback'](self)
```
dpkp: I think there may be some deadlock issues wrt this callback because it requires the client lock:

thread A acquires conn._lock
thread B acquires client._lock
thread A calls conn_state_change, blocks waiting for client._lock
thread B calls conn.close() or conn.send() etc and blocks waiting for conn._lock

=> deadlock

kafka/conn.py
```diff
@@ -784,43 +789,48 @@ def close(self, error=None):
...
+            self._update_reconnect_backoff()
+            self._close_socket()
+            self.state = ConnectionStates.DISCONNECTED
+            self._sasl_auth_future = None
+            self._protocol = KafkaProtocol(
+                client_id=self.config['client_id'],
+                api_version=self.config['api_version'])
+            self._fail_ifrs(error)
```
dpkp: I think we want to release the conn lock before processing callbacks here

dpkp:
I merged #1768, which uses a non-reentrant lock. I'm planning to merge #1775 as well and I think that should cover all of concurrency issues I found on review.  If you're able to test these changes in your setup, I would love that. We're planning to push out a patch release quickly.

Thanks again for all your excellent work on this issue. Fantastic bug report, debugging, and PRs. We really appreciate it, and I hope 1.4.6 will not fail you!

PR #1768, opened 2019-03-29
Additional BrokerConnection locks to synchronize protocol/IFR state

dpkp:
@isamaru I threw this together as perhaps a middle-ground between the attempts to fix protocol-out-of-sync issues discussed in #1744 .

One small detail re: python atomic operations / locking: my understanding is that a simple boolean check on a python dict is atomic and would not require additional locks. Note the single CALL_FUNCTION opcode:
```
>>> import dis
>>> n = {'a': 1, 'b': 2}
>>> def foo():
...     bool(n)
...
>>> dis.dis(foo)
  2           0 LOAD_GLOBAL              0 (bool)
              2 LOAD_GLOBAL              1 (n)
              4 CALL_FUNCTION            1
              6 POP_TOP
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE
```

Of course, the value could still change between when the dict is loaded and when it is returned, which means the return value could be stale. But the same is true if we grab the lock, check the value, and then release the lock. As soon as we've released the lock, the value may become stale before we make our next decision. Which means that for the most part, if the value is important for decisions about state, we should check the value directly while holding the lock at the decision point. Wrapping an accessor method in a lock doesn't go far enough.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1768)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -586,11 +586,12 @@ class BrokerConnection(object):
                                self.config['sasl_plain_password']]).encode('utf-8'))
         size = Int32.encode(len(msg))
         try:
-            self._send_bytes_blocking(size + msg)
+            with self._lock:
+                self._send_bytes_blocking(size + msg)
 
-            # The server will send a zero sized message (that is Int32(0)) on success.
-            # The connection is closed on failure
-            data = self._recv_bytes_blocking(4)
+                # The server will send a zero sized message (that is Int32(0)) on success.
+                # The connection is closed on failure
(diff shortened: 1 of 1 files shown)


dpkp:
There are still some issues related to the state_change_callback that need to be cleaned up, but I think they can be dealt with in separate PRs.

isamaru:
@dpkp Thanks a lot, this looks better than my attempts. Will try 1.4.6.!

PR #1772, opened 2019-03-31
Avoid race condition on client._conns in send()

dpkp:
There was a very small possibility that between checking `self._can_send_request(node_id)` and grabbing the connection object via `self._conns[node_id]` that the connection could get closed / recycled / removed from _conns and cause a KeyError. This PR should prevent such a KeyError. In the case where the connection is disconnected by the time we call send(), we should expect conn.send() simply to fail the request.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1772)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -516,14 +516,15 @@ class KafkaClient(object):
         Returns:
             Future: resolves to Response struct or Error
         """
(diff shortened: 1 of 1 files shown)


jeffwidman:
Makes sense and looks very straightforward, merging

PR #1773, opened 2019-03-31
Dont treat popped conn.close() as failure in state change callback

dpkp:
Small change to KafkaClient's state_change_callback to attempt to avoid re-fetching Metadata or increasing bootstrap failure count after an intentional disconnect / conn.close()

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1773)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -314,7 +314,12 @@ class KafkaClient(object):
                     idle_disconnect = True
                 self._idle_expiry_manager.remove(node_id)
 
-                if self.cluster.is_bootstrap(node_id):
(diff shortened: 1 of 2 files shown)


kafka/client_async.py
```diff
@@ -419,10 +424,12 @@ def close(self, node_id=None):
         with self._lock:
             if node_id is None:
```
jeffwidman: while the code is relatively clear, it'd be nice if the docstring stated that `node_id=None` results in closing all broker connections. The docstring implies it, but doesn't say that explicitly.

Issue #1774, opened 2019-04-01
Consumer.partitions_for_topic requires metadata has already been fetched


ershashu:
Consumer 'partitions_for_topic' method is not returning the proper value. 

We have used the method : partitions_for_topic(topic) in our code, it was working fine with version 1.4.4 but getting an error in latest one (1.4.5)

dpkp:
`partitions_for_topic()` is currently non-blocking and will return empty if the client has not yet fetched metadata. It is likely that in the 1.4.4 you had already fetched metadata in the cache. I would suggest calling `consumer.topics()` to make a blocking metadata request prior to calling `partitions_for_topic()`

Baisang:
hey @dpkp, I recently ran into this issue as well in some scripts using version 1.4.2. The fix you described (fetching metadata with `consumer.topics()` works, but we never ran into the issue before (and it seems like the issue is more common now?). Not sure if something else changed (feels like it shouldn't have?) to explain this behavior.

dpkp:
I believe the java client has been updated to make a blocking request for metadata in `partitionsFor` -- we probably should do the same.

jeffwidman:
Fixed by #1781

PR #1781, opened 2019-04-07
Update group.partitions_for_topic to fetch topic metadata if it does …

Baisang:
…not have it

As mentioned in https://github.com/dpkp/kafka-python/issues/1774, this attempts to simulate the Java client, which does the following

```
public java.util.List<PartitionInfo> partitionsFor(java.lang.String topic)
Get metadata about the partitions for a given topic. This method will issue a remote call to the server if it does not already have any metadata about the given topic.
```
https://kafka.apache.org/10/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html#partitionsFor

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1781)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -552,11 +552,9 @@ class KafkaConsumer(six.Iterator):
                 committed = None
         return committed
 
-    def topics(self):
-        """Get all topics the user is authorized to view.
-
-        Returns:
-            set: topics
+    def _fetch_all_topit_metadata(self):
+        """A blocking call that fetches topic metadata for all topics in the
+        cluster that the user is authorized to view.
         """
         cluster = self._client.cluster
         if self._client._metadata_refresh_in_progress and self._client._topics:
@@ -567,10 +565,24 @@ class KafkaConsumer(six.Iterator):
         future = cluster.request_update()
         self._client.poll(future=future)
         cluster.need_all_topic_metadata = stash
-        return cluster.topics()
+
+    def topics(self):
+        """Get all topics the user is authorized to view.
+        This will always issue a remote call to the cluster to fetch the latest
+        information.
+
+        Returns:
+            set: topics
+        """
+        self._fetch_all_topic_metadata()
+        return self._client.cluster.topics()
 
     def partitions_for_topic(self, topic):
-        """Get metadata about the partitions for a given topic.
+        """This method first checks the local metadata cache for information
+        about the topic. If the topic is not found (either because the topic
+        does not exist, the user is not authorized to view the topic, or the
+        metadata cache is not populated), then it will issue a metadata update
+        call to the cluster.
 
         Arguments:
             topic (str): Topic to check.
@@ -578,7 +590,12 @@ class KafkaConsumer(six.Iterator):
         Returns:
             set: Partition ids
         """
-        return self._client.cluster.partitions_for_topic(topic)
+        cluster = self._client.cluster
+        partitions = cluster.partitions_for_topic(topic)
+        if partitions is None:
+            self._fetch_all_topic_metadata()
+            partitions = cluster.partitions_for_topic(topic)
+        return partitions
 
     def poll(self, timeout_ms=0, max_records=None):
         """Fetch data from assigned topics / partitions.


kafka/consumer/group.py
```diff
@@ -567,18 +564,34 @@ def topics(self):
...
         Arguments:
             topic (str): Topic to check.
 
         Returns:
             set: Partition ids
         """
-        return self._client.cluster.partitions_for_topic(topic)
+        cluster = self._client.cluster
+        partitions = cluster.partitions_for_topic(topic)
+        if not partitions:
```
jeffwidman: The return value of `partitions` is `None`, so let's explicitly check for that: `if partitions is None`

Additionally, I noticed that a non-existent topic will also return `None`, so this will issue a remote call if the topic simply does not exist... I can't think of any perf problems this will cause and it's  already mentioned in my suggested docstring so I think it's fine.

```
In [1]: from kafka import KafkaConsumer

In [2]: kc = KafkaConsumer('test_topic', group_id='testings', bootstrap_servers=['172.28.0.3'])

In [3]: cluster = kc._client.cluster

In [4]: cluster.partitions_for_topic('test_poll_group_by_topic_true')

In [5]: cluster.partitions_for_topic('non_existent_topic')

In [6]: kc.topics()
Out[6]: {u'test_poll_group_by_topic_true'}

In [7]: cluster.partitions_for_topic('non_existent_topic')

In [8]: cluster.partitions_for_topic('test_poll_group_by_topic_true')
Out[8]: {0}
```

kafka/consumer/group.py
```diff
@@ -567,18 +564,34 @@ def topics(self):
         future = cluster.request_update()
         self._client.poll(future=future)
         cluster.need_all_topic_metadata = stash
+
+    def topics(self):
+        """Get all topics the user is authorized to view.
```
jeffwidman: Can you add:
`This will always issue a remote call to the cluster to fetch the latest information`.

kafka/consumer/group.py
```diff
@@ -567,18 +564,34 @@ def topics(self):
...
+            set: topics
+        """
+        cluster = self._client.cluster
+        self._fetch_all_topic_metadata()
         return cluster.topics()
 
     def partitions_for_topic(self, topic):
         """Get metadata about the partitions for a given topic.
+        This method will issue a remote call to the server if it does not already
+        have any metadata about the given topic.
```
jeffwidman: Let's make it clear in the docstring that there will be times when this method could return values that do not match the state of the cluster due to an outdated cache entry... 

Also, while I have never played with Kafka ACL's, I'm guessing since `_fetch_all_topic_metadata()` only returns info about topics the user is authorized to view that trying to fetch info about a topic that exists but the user can't view will also trigger a remote call / returning `None`? Normally I'd expect a call about a specific topic to return some kind of error from the cluster about the user not authorized to view but here the cluster never sees a topic-scoped call so that's probably not the case.

maybe reword to:

`This method first checks the local metadata cache for information about the topic. If the topic is not found (either because the topic does not exist, the user is not authorized to view the topic, or the metadata cache is not populated), then it will issue a metadata update call to the cluster.`

kafka/consumer/group.py
```diff
@@ -552,11 +552,8 @@ def committed(self, partition):
                 committed = None
         return committed
 
-    def topics(self):
-        """Get all topics the user is authorized to view.
-
-        Returns:
-            set: topics
+    def _fetch_all_topic_metadata(self):
```
jeffwidman: Is the Java client structured this way with this private method?

It's a nice clean way to do it this way, but just curious how they handle it...

kafka/consumer/group.py
```diff
@@ -552,11 +552,8 @@ def committed(self, partition):
...
+        """A blocking call that fetches topic metadata for all topics in the cluster.
```
jeffwidman: I'm guessing the statement `...that the user is authorized to view.` also applies here?

kafka/consumer/group.py
```diff
@@ -567,18 +564,34 @@ def topics(self):
...
         self._client.poll(future=future)
         cluster.need_all_topic_metadata = stash
+
+    def topics(self):
+        """Get all topics the user is authorized to view.
+
+        Returns:
+            set: topics
+        """
+        cluster = self._client.cluster
```
jeffwidman: Why is this needed?

Why not just change it to `return self._client.cluster.topics()`?

kafka/consumer/group.py
Baisang: I also have not played with Kafka's ACL's, but that would be my assumption as well.

kafka/consumer/group.py
Baisang: Yep! I think that's what I based it on https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1892

Baisang:
@jeffwidman Thanks for the feedback -- I addressed your comments, hopefully travis is kind to me.

jeffwidman:
🎆 

Thanks again for this!

Issue #1780, opened 2019-04-06
Consumer can block indefinitely when resetting offsets for a deleted topic


commanderdishwasher:
Hi, bumped into the following edge case which results in the fetcher getting stuck indefinitely
inside `fetcher._retrieve_offsets()`:

- Consumer is running in a stable group, subscribed to multiple topics
- A topic that is part of the subscription is deleted (no commits for these TPs)
- A fetch request will get `error_code=3` (Unknown) and a `MetadataRequest` will be made
- Subscription state is being updated due after the `MetadataResponse`
- Rebalance is triggered due to subscription change
- Revoke, group leader election, sync assignment, and assignment are taking place
- During that time, another topic that was part of the subscription is deleted (no commits for these TPs either)
- `KafkaConsumer._coordinator.poll()` did not trigger a rebalance
- `KafkaConsumer._update_fetch_positions(self._subscription.missing_fetch_positions())` is causing the client to get the reset positions for the assignment (which still holds a topic that no longer exists due to deletion)
- `Errors.UnknownTopicOrPartitionError` is thrown from `fetcher._handle_offset_response()` since the partition does no exist
- A metadata refresh is taking place and the subscription is updated, but we are caught in an infinite loop inside `fetcher._retrieve_offsets()` since the `timestamps` are not updated and timeout is infinity and we now get an early exit in `fetcher._send_offset_requests()` that forces metadata update over and over again 

```    def _send_offset_requests(self, timestamps):
        timestamps_by_node = collections.defaultdict(dict)
        for partition, timestamp in six.iteritems(timestamps):
            node_id = self._client.cluster.leader_for_partition(partition)
            if node_id is None:
                self._client.add_topic(partition.topic)
                log.debug("Partition %s is unknown for fetching offset,"
                          " wait for metadata refresh", partition)
                return Future().failure(Errors.StaleMetadata(partition))
...
...
```

If you guide me to what you think would be a good way to tackle this, i'll work on a PR.

Cheers.

commanderdishwasher:
The first thing I wanted to do is to remove the unknown partition from the timestamps list since the metadata and subscription are updated.

jeffwidman:
What version of Kafka broker / `kafka-python` are you using

commanderdishwasher:
1.4.4, 1.4.5, 1.4.6, master

Kafka 0.10.2.1 / 2.0.0

commanderdishwasher:
[...]

PR #1782, opened 2019-04-08
Issue #1780 - Consumer hang indefinitely in fetcher._retrieve_offsets() due to topic deletion while rebalancing 

commanderdishwasher:
Detailed scenario in #1780 .

The scenario can also trigger another edge case I bumped into while testing which is
`ValueError` exception from `subscription_state.assign_from_subscribed()`

This PR also handles this case by catching the exception and requesting a rejoin.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1782)

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -235,14 +235,16 @@ class Fetcher(six.Iterator):
         log.debug("Resetting offset for partition %s to %s offset.",
                   partition, strategy)
         offsets = self._retrieve_offsets({partition: timestamp})
-        if partition not in offsets:
-            raise NoOffsetForPartitionError(partition)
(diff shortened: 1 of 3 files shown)


jeffwidman:
nudge @dpkp can you take a look?

dpkp:
Apologies for not merging this sooner. I'm going to include this PR in 1.4.7.

Please note that w/o test coverage, we may end up breaking this functionality in a later release! This section of the codebase (retrieving offsets) has been refactored upstream, and I suspect we may migrate in that direction at some point.

royantman:
thanks @dpkp, as mentioned in the original issue, I will get into looking at the refactor to match Java client 2.x eventually. Just got my hands full atm.

Cheers.

PR #1792, opened 2019-04-24
Minor housekeeping--fixed typos: iff -> if

rustyrothwurt:
Was reading and re-reading the docs while working/musing on a problem and kept running into "iff". I thought it was perhaps a choice of style, but I am pretty sure the author meant if and could have been a simple copy and paste error.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1792)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -757,7 +757,7 @@ class BrokerConnection(object):
             return float('inf')
 
     def connected(self):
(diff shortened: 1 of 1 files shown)


dpkp:
iff typically means "if and only if"

rustyrothwurt:
Ah -- gotcha. That makes sense

PR #1807, opened 2019-05-18
Use futures to parallelize calls to _send_request_to_node()

Lou-Cipher:
Changes made as per the discussion in #1801 
Uses futures to parallelize broker calls in the admin client.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1807)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -251,7 +251,7 @@ class KafkaAdminClient(object):
             request = MetadataRequest[version]()
             future = self._send_request_to_node(self._client.least_loaded_node(), request)
 
-            wait_for_futures([future])
+            self.wait_for_futures([future])
 
             response = future.value
             controller_id = response.controller_id
@@ -287,7 +287,7 @@ class KafkaAdminClient(object):
         gc_request = GroupCoordinatorRequest[0](group_id)
         future = self._send_request_to_node(self._client.least_loaded_node(), gc_request)
 
-        wait_for_futures([future])
+        self.wait_for_futures([future])
 
         gc_response = future.value
         # use the extra error checking in add_group_coordinator() rather than
@@ -339,7 +339,7 @@ class KafkaAdminClient(object):
             tries -= 1
             future = self._send_request_to_node(self._controller_id, request)
 
-            wait_for_futures([future])
+            self.wait_for_futures([future])
 
             response = future.value
             # In Java, the error fieldname is inconsistent:
@@ -500,7 +500,7 @@ class KafkaAdminClient(object):
                 .format(version))
         future = self._send_request_to_node(self._client.least_loaded_node(), request)
 
-        wait_for_futures([future])
+        self.wait_for_futures([future])
 
         return future.value
 
@@ -543,7 +543,7 @@ class KafkaAdminClient(object):
         # So this is currently broken as it always sends to the least_loaded_node()
         future = self._send_request_to_node(self._client.least_loaded_node(), request)
 
-        wait_for_futures([future])
+        self.wait_for_futures([future])
 
         return future.value
 
@@ -641,7 +641,7 @@ class KafkaAdminClient(object):
                     "Support for DescribeGroups v{} has not yet been added to KafkaAdminClient."
                     .format(version))
 
-        wait_for_futures(futures)
+        self.wait_for_futures(futures)
 
         for future in futures:
             response = future.value
@@ -705,7 +705,7 @@ class KafkaAdminClient(object):
             for broker_id in broker_ids:
                 futures.append(self._send_request_to_node(broker_id, request))
 
-            wait_for_futures(futures)
+            self.wait_for_futures(futures)
 
             for future in futures:
                 response = future.value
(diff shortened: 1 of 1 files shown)


kafka/admin/client.py
```diff
@@ -605,42 +621,49 @@ def describe_consumer_groups(self, group_ids, group_coordinator_id=None):
...
         for group_id in group_ids:
-            if group_coordinator_id is not None:
-                this_groups_coordinator_id = group_coordinator_id
-            else:
-                this_groups_coordinator_id = self._find_group_coordinator_id(group_id)
             if version <= 1:
+                if group_coordinator_id is not None:
+                    this_groups_coordinator_id = group_coordinator_id
+                else:
+                    this_groups_coordinator_id = self._find_group_coordinator_id(group_id)
```
jeffwidman: Is there a reason you moved this under the version check? This seems generic code that's currently needed regardless of the version.

Side note: `_find_group_coordinator_id()` is a blocking call so this could be further optimized into firing off a bunch of futures for all the groups and then handling the responses but that would require breaking `_find_group_coordinator_id()` into two methods to generate request/parse the response... that optimization is probably best handled in different PR. Especially because that wouldn't even be used when fetching offsets for all consumer groups since the group coordinator is explicitly passed in already... example: https://github.com/DataDog/integrations-core/pull/2730/files#diff-deed983f73a04522ac621d7f6d0c8403R244

kafka/admin/client.py
```diff
@@ -738,7 +767,12 @@ def list_consumer_group_offsets(self, group_id, group_coordinator_id=None,
                     topics_partitions_dict[topic].add(partition)
                 topics_partitions = list(six.iteritems(topics_partitions_dict))
             request = OffsetFetchRequest[version](group_id, topics_partitions)
-            response = self._send_request_to_node(group_coordinator_id, request)
+            future = self._send_request_to_node(group_coordinator_id, request)
+
+            self.wait_for_futures([future])
+
+            response = future.value
+
```
jeffwidman: I would probably delete a lot of these blank lines, they make the code more difficult to read since IMHO they don't really separate out distinct sections of code... so it makes the code slightly less skimmable...

Lou-Cipher:
Thanks for the feedback and tips, I'll update the MR this evening and report back here.

kafka/admin/client.py
Lou-Cipher: Figured it could be a minor optimization.  We don't actually use this_groups_coordinator_id unless version <= 1 evaluates true, so why initialize it or call _find_group_coordinator_id() if we don't have to?

I'll peek at _find_group_coordinator_id() when I have a spare minute and see if I can optimize it.  Agreed, something for another PR, but shouldn't be too bad.

kafka/admin/client.py
```diff
@@ -764,3 +798,13 @@ def list_consumer_group_offsets(self, group_id, group_coordinator_id=None,
...
+    def wait_for_futures(self, futures):
+        while not all(future.succeeded() for future in futures):
+            for future in futures:
+                self._client.poll(future=future)
+
+                if future.failed():
+                    raise future.exception  # pylint: disable-msg=raising-bad-type
+
+        return True
```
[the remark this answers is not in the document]
Lou-Cipher: Eh, I grew up on C, I likes my return values to be explicit.  But agreed, I'll remove it.

Lou-Cipher:
OK, I had a quick moment, so I made the changes.  The one thing I left for the moment is the code for this_groups_coordinator_id.  Since it's only currently used in the one conditional, I figured for the moment it's more optimized to move it under that conditional.  I can move it back if you want, just figured we could debate the merits a bit.

jeffwidman:
> The one thing I left for the moment is the code for this_groups_coordinator_id. Since it's only currently used in the one conditional, I figured for the moment it's more optimized to move it under that conditional.

Gotcha. From this section of code, that reasoning makes sense. 

However, under the covers https://github.com/dpkp/kafka-python/pull/1807/files#diff-c893e6ff8a43607853864c15224c1802R627 will only blow up if we're talking to such a new broker that it drops support for the `DescribeGroups_v1` request, which I think extremely unlikely to happen for years. And in that case, we should be fixing by adding support for `DescribeGroups_v2` to support the new request style. 

The other case is if we are communicating with an ancient broker that doesn't support the `DescribeGroups` request at all. However the `KafkaAdminClient` only guarantees support for clusters > `0.10`. Additionally `DescribeGroups` has been around long enough that other code will blow up if we're talking to such an old cluster.

So while you're totally right that it's a nifty little micro-optimization (and I'd overlooked that earlier), in the broader picture the optimization is actually extremely unlikely to ever be hit.

> I can move it back if you want, just figured we could debate the merits a bit.

Absolutely, happy to discuss these things anytime.

> I'm assuming you want it back out in the event a future version conditional is added which will also need to reference this_groups_coordinator_id?

I think that quite likely. So having it DRY'd up is more likely to be of benefit than the optimization of putting it under the conditional.

At the end of the day, this one isn't a big deal to me--up to you. I am happy to merge either way, I just think it's cleaner how it already is.

Lou-Cipher:
OK, I've moved it back to the original location outside the conditional.  Agreed that it's a very minor optimization, costs outweighing the benefits and whatnot.  I think once the tests pass, we should be good!

jeffwidman:
Merged, thank you again!

Lou-Cipher:
Woot, and thank you!  Pretty excited, my first commit to an OSS project that has nothing to do with either media centers or emulators.  I'll peek at _find_group_coordinator_id() and see if I can't optimize that, too.

PR #1809, opened 2019-05-23
Fix typo in _fetch_all_topic_metadata function

Baisang:
@jeffwidman wow, this is really embarrassing. I just noticed there's a typo in https://github.com/dpkp/kafka-python/pull/1781, so I corrected it.

It's weird it wasn't caught by any tests -- do we not do any testing for these methods?

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1809)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -552,7 +552,7 @@ class KafkaConsumer(six.Iterator):
                 committed = None
         return committed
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
Oops, thanks for catching that.

Hmm, not sure about test coverage... clearly not! If you have time to add one, we'd certainly appreciate it. Even a basic one would have caught this.

PR #1814, opened 2019-05-24
Remove string support in `api_version`

jeffwidman:
This breaks backwards compatibility, probably best not to merge 'til `2.0` release
--
A long time ago, `api_version` supported strings. That has been
deprecated for years in favor of tuples. Time to remove support for the
strings.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1814)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -340,16 +340,6 @@ class KafkaConsumer(six.Iterator):
         self._metrics = Metrics(metric_config, reporters)
         # TODO _metrics likely needs to be passed to KafkaClient, etc.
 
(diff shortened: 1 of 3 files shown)


dpkp:
What do we gain by removing this? I understand the desire to keep code clean, but is there a specific feature that is blocked by keeping this in ?

jeffwidman:
Agreed. And personally I think the string value is a sensible enough choice, even though it is error prone.

I also noticed that `confluent-kafka-python` uses this config value as a string, so I'd rather keep us supporting a similar interface for folks switching back and forth.

PR #1815, opened 2019-05-24
Remove deprecated `'smallest'`/`'largest'`

jeffwidman:
This breaks backwards compatibility, probably best not to merge 'til `2.0` release
--
These have been deprecated in favor of `'earliest'` and `'latest'` for a
while.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1815)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -315,13 +315,6 @@ class KafkaConsumer(six.Iterator):
         self.config = copy.copy(self.DEFAULT_CONFIG)
         self.config.update(configs)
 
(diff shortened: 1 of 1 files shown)


dpkp:
Again, I understand the desire to remove old code, but if this doesn't unlock some new feature I'm not convinced we should do it. I don't really want to treat a 2.0 release as license to break interfaces -- would prefer that we try to avoid breaking things unless needed.

jeffwidman:
Agreed.

PR #1821, opened 2019-05-29
Change coordinator lock acquisition order

dpkp:
Attempt to address heartbeat lock contention by changing the lock acquisition order from client->coordinator to coordinator->client. This aligns with the java client and (I hope) should help improve concurrency coordination between the heartbeat thread loop and the main consumer loop by allowing us to drop the client lock during future processing.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1821)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -595,7 +595,9 @@ class KafkaClient(object):
 
                 self._poll(timeout / 1000)
 
-                responses.extend(self._fire_pending_completed_requests())
+            # called without the lock to avoid deadlock potential
+            # if handlers need to acquire locks
+            responses.extend(self._fire_pending_completed_requests())
 
             # If all we had was a timeout (future is None) - only do one poll
             # If we do have a future, we keep looping until it is done
--- a/kafka/coordinator/base.py
+++ b/kafka/coordinator/base.py
@@ -243,7 +243,7 @@ class BaseCoordinator(object):
         """Block until the coordinator for this group is known
         (and we have an active connection -- java client uses unsent queue).
         """
-        with self._client._lock, self._lock:
+        with self._lock:
             while self.coordinator_unknown():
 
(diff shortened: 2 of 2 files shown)


dpkp:
The thing to be concerned about here is whether it is possible to find a code path that attempts to acquire the coordinator lock while holding the client lock. Unfortunately, I think this is super tricky because even though we have deferred processing of futures that handle connection responses into an unlocked pending_completions list, there are loads of other futures that are not in this list. Now, most of these are not going to get generated while the client lock is acquired (see `git grep future.success`). But a very real risk is future.failure(). An obvious place where this might happen is during `KafkaClient._maybe_connect`, which acquires the client lock and then calls into the `BrokerConnection.connect` function. If something goes wrong during connect, we will call `self.close()` and that will cause all in-flight-request futures to be immediately failed. That will cause errbacks to run with the client lock acquired, and if one of those errbacks ends up attempting to acquire the coordinator lock, for example in `_handle_join_failure` or `_handle_heartbeat_failure`, then we could get a deadlock.

The java client uses its `RequestFutureCompletionHandler` to shim these connection failures onto the pending completions queue, in order to avoid the problem of errbacks running while the client lock is acquired. I expect that we will need to implement something similar to really get this under control.

dpkp:
I haven't found any specific paths that would trigger a deadlock with this change. The example I mentioned above would require that somehow we get an in-flight-request before the connection is completed. That shouldn't happen, so I'm going to merge and keep an eye out for any other issues.

huangcuiyang:
I also think that the repair method you suggested is more thorough and better, and my method is relatively simple, in order to quickly repair production problems.
When the heartbeat thread uses 2 keys to call client.poll to read data, the performance of the main consumer thread is less affected. I have tested the performance.
I hope you can use a better way to solve the design problem of multi-threaded shared client.poll

PR #1822, opened 2019-05-29
Use dedicated connection for group coordinator

dpkp:
This changes the coordinator_id to be a unique string, e.g., `coordinator-1`, so that it will get a dedicated connection. This won't eliminate lock contention because the client lock applies to all connections, but it should improve in-flight-request contention.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1822)

### DIFF
--- a/kafka/cluster.py
+++ b/kafka/cluster.py
@@ -65,6 +65,7 @@ class ClusterMetadata(object):
                 self.config[key] = configs[key]
 
         self._bootstrap_brokers = self._generate_bootstrap_brokers()
+        self._coordinator_brokers = {}
 
     def _generate_bootstrap_brokers(self):
         # collect_hosts does not perform DNS, so we should be fine to re-use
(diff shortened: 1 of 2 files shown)


kafka/cluster.py
```diff
@@ -341,41 +346,28 @@ def add_group_coordinator(self, group, response):
...
         log.debug("Updating coordinator for %s: %s", group, response)
         error_type = Errors.for_code(response.error_code)
         if error_type is not Errors.NoError:
             log.error("GroupCoordinatorResponse error: %s", error_type)
             self._groups[group] = -1
-            return False
+            return
 
-        node_id = response.coordinator_id
+        # Use a coordinator-specific node id so that group requuests
```
jeffwidman: typo: `requuests`

dpkp: uu are the best

PR #1823, opened 2019-05-29
Reduce client poll timeout when no IFRs

dpkp:
When there are no IFRs, client.poll can potentially block up to the full request timeout ms. For consumers, this would also prevent the heartbeat thread from acquiring the client lock and therefore prevent the heartbeat loop from proceeding. By reducing the poll timeout to the retry timeout (typically 100ms), we should help prevent lock contention and enable the heartbeat thread to loop more freely. Note that when there is an IFR, we may still block for the full request timeout, so this isn't a general solution.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1823)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -588,6 +588,9 @@ class KafkaClient(object):
                         metadata_timeout_ms,
                         idle_connection_timeout_ms,
                         self.config['request_timeout_ms'])
+                    # if there are no requests in flight, do not block longer than the retry backoff
+                    if self.in_flight_request_count() == 0:
+                        timeout = min(timeout, self.config['retry_backoff_ms'])
                     timeout = max(0, timeout / 1000)  # avoid negative timeouts
 
                 self._poll(timeout)
--- a/test/test_client_async.py
+++ b/test/test_client_async.py
@@ -233,6 +233,8 @@ def test_send(cli, conn):
 def test_poll(mocker):
     metadata = mocker.patch.object(KafkaClient, '_maybe_refresh_metadata')
     _poll = mocker.patch.object(KafkaClient, '_poll')
+    ifrs = mocker.patch.object(KafkaClient, 'in_flight_request_count')
+    ifrs.return_value = 1
     cli = KafkaClient(api_version=(0, 9))
 
     # metadata timeout wins
@@ -248,6 +250,11 @@ def test_poll(mocker):
     metadata.return_value = 1000000
     cli.poll()
     _poll.assert_called_with(cli.config['request_timeout_ms'] / 1000.0)
+
+    # If no in-flight-requests, drop timeout to retry_backoff_ms
+    ifrs.return_value = 0
+    cli.poll()
+    _poll.assert_called_with(cli.config['retry_backoff_ms'] / 1000.0)
 
 
 def test__poll():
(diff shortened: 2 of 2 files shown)


kafka/client_async.py
```diff
@@ -588,6 +588,9 @@ def poll(self, timeout_ms=None, future=None):
                         metadata_timeout_ms,
                         idle_connection_timeout_ms,
                         self.config['request_timeout_ms'])
+                    # if there are no requests in flight, do not block longer than the retry backoff
+                    if self.in_flight_request_count() == 0:
+                        timeout = min(timeout, self.config['retry_backoff_ms'])
                     timeout = max(0, timeout / 1000)  # avoid negative timeouts
```
jeffwidman: It feels a little weird that this conflates both avoiding negative timeouts and changing the units from ms to seconds. The code works fine as-is, it just required me to re-read it twice to understand what was happening, especially given the inline comment. I would probably delay the conversion into seconds until `self._poll(timeout / 1000)` here: https://github.com/dpkp/kafka-python/pull/1823/files#diff-f161ac626efaecd3719ea5d72129604cR596 That way you keep it consistent as milliseconds until the last possible moment.

dpkp: makes sense -- worth doing that in a separate PR

jeffwidman: Done in https://github.com/dpkp/kafka-python/pull/1825, will merge once tests pass (to make sure I'm not overlooking something)

jeffwidman:
WOW! 

I have a script that fetches all consumer offsets and calculates consumer lag for all consumer groups on cluster. On one cluster that was consistently taking ~242 seconds so the metrics agent was constantly timing out the script.

Part of the problem was that my I/O code was two nested for loops:

    for broker in kafka_admin_client._client.cluster.brokers():
        for consumer_group, group_type in kafka_admin_client.list_consumer_groups(broker_ids=[broker.nodeId]):
            # consumer groups from Kafka < 0.9 that store their offset
            # in Kafka don't use Kafka for group-coordination so
            # group_type is empty
            if group_type in ('consumer', ''):
                # Typically the consumer group offset fetch sequence is:
                # 1. For each broker in the cluster, send a ListGroupsRequest
                # 2. For each consumer group, send a FindGroupCoordinatorRequest
                # 3. Query the group coordinator for the consumer's offsets.
                # However, since Kafka brokers only include consumer
                # groups in their ListGroupsResponse when they are the
                # coordinator for that group, we can skip the
                # FindGroupCoordinatorRequest.
                this_group_offsets = kafka_admin_client.list_consumer_group_offsets(
                    group_id=consumer_group, group_coordinator_id=broker.nodeId)

In other words, I was making each request sequentially and waiting for the response before continuing. So I spent some time converting that to async I/O using callbacks (see #1845). 

I immediately saw the fetch time drop from 242 seconds to 2 seconds, and considered myself a minor genius.

However, I was curious so also ran the old code and saw that it took ~3 seconds when run using the code from `master`. I eventually tracked the change down to this PR... 

When I use the callbacks approach without the change in this PR, I see collection time taking either 31 or 61 seconds, just depends. So it did make a difference. But as soon as I added this change, callbacks only saves ~0.5-1 seconds.

@dpkp You mentioned:

> Note that when there is an IFR, we may still block for the full request timeout, so this isn't a general solution.

Given the dramatic improvement I saw from this PR, I'm wondering whether we might see similar results from this possible "more general solution"??

dpkp:
I think the more general solution here is to make the list of connections tracked by the selector more dynamic so that we only "select" against connections that have pending sends or recvs.

PR #1824, opened 2019-05-29
Reduce internal client poll timeout for consumer iterator interface

dpkp:
Another PR attempting to address heartbeat timing issues in consumers, especially with the iterator interface. Here we can reduce the client.poll timeout to at most the retry backoff (typically 100ms) so that the consumer iterator interface doesn't block for longer than the heartbeat timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1824)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -1070,8 +1070,7 @@ class KafkaConsumer(six.Iterator):
                 self._update_fetch_positions(partitions)
 
             poll_ms = 1000 * (self._consumer_timeout - time.time())
-            if not self._fetcher.in_flight_fetches():
-                poll_ms = min(poll_ms, self.config['reconnect_backoff_ms'])
+            poll_ms = min(poll_ms, self.config['retry_backoff_ms'])
             self._client.poll(timeout_ms=poll_ms)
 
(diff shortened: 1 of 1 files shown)


kafka/consumer/group.py
```diff
@@ -1070,8 +1070,7 @@ def _message_generator(self):
                 self._update_fetch_positions(partitions)
 
             poll_ms = 1000 * (self._consumer_timeout - time.time())
-            if not self._fetcher.in_flight_fetches():
-                poll_ms = min(poll_ms, self.config['reconnect_backoff_ms'])
+            poll_ms = min(poll_ms, self.config['retry_backoff_ms'])
```
jeffwidman: I would probably do this as a one-liner combined with the line above (https://github.com/dpkp/kafka-python/pull/1824/files#diff-9da855eb60eed8ecc72ce217c25e576cR1072)... I think it's easier to mentally think through the calculated value that way

jeffwidman: I pushed this as an additional commit... feel free to either squash or if you really don't like it then just drop that particular commit.

Issue #1810, opened 2019-05-23
Add test for consumer.topics(), consumer.partitions_for_topic()


jeffwidman:
In https://github.com/dpkp/kafka-python/pull/1809 we discovered there are no tests on `consumer.topics()` and `consumer.partitions_for_topic()`.

This should be remedied. Even a basic test would have caught the typo.

jeffwidman:
Fixed by #1829

PR #1829, opened 2019-06-01
Sanity test for consumer.topics() and consumer.partitions_for_topic()

Baisang:
This should hopefully address https://github.com/dpkp/kafka-python/issues/1810. A really simple sanity check for the two methods (I guess it's also a sanity check for `subscribe()` as well)

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1829)

### DIFF
--- a/test/test_consumer_group.py
+++ b/test/test_consumer_group.py
@@ -29,6 +29,15 @@ def test_consumer(kafka_broker, topic, version):
     assert consumer._client._conns[node_id].state is ConnectionStates.CONNECTED
     consumer.close()
 
+@pytest.mark.skipif(not version(), reason="No KAFKA_VERSION set")
+def test_consumer_topics(kafka_broker, topic, version):
+    consumer = KafkaConsumer(bootstrap_servers=get_connect_str(kafka_broker))
+    # Necessary to drive the IO
+    consumer.poll(500)
+    consumer_topics = consumer.topics()
+    assert topic in consumer_topics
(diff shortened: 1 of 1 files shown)


test/test_consumer_group.py
```diff
@@ -29,6 +29,15 @@ def test_consumer(kafka_broker, topic, version):
     assert consumer._client._conns[node_id].state is ConnectionStates.CONNECTED
     consumer.close()
 
+@pytest.mark.skipif(not version(), reason="No KAFKA_VERSION set")
+def test_consumer_topics(kafka_broker, topic, version):
+    consumer = KafkaConsumer(bootstrap_servers=get_connect_str(kafka_broker))
+    consumer.subscribe([topic])
```
jeffwidman: Is subscribe() needed? Couldn't you just call topics() and partitions_for_topic() since they are both blocking for metadata? You may need to `poll()` at some point (even without an active subscription) in order to drive the I/O.

Baisang: Let me try that -- I do think it'd be good to add on other tests eventually that would check subscribe() and other methods

Baisang: @jeffwidman I changed the test to just have poll and removed subscribe

jeffwidman: > I do think it'd be good to add on other tests eventually that would check subscribe() and other methods

Absolutely agree. Happily welcome a PR. Just rather keep those in a separate test.

jeffwidman:
Thanks again for this!

Issue #1594, opened 2018-10-01
auto commit never executed with api_version 0.8.1


gaetano-guerriero:
Hello,
I'm using KafkaConsumer with `api_version=(0, 8, 1)` and `enable_auto_commit=True`.
I need to use api_version 0.8.1 since the consumer I'm migrating from uses `OffsetCommitRequest` version 0 and so stores offsets on zookeeper.

The offsets are not periodically committed as expected. From what I can understand this is due to early exit in `coordinator.ConsumerCoordinator.poll()`

jeffwidman:
Currently, `KafkaConsumer` doesn't support writing offsets to Zookeeper: https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer.commit

You'll need to use the old / deprecated `SimpleConsumer` to commit to zookeeper.

justecorruptio:
Hi hi,

We're doing this exact same thing ie. `api_version=(0, 8, 1)` and `enable_auto_commit=True`. This had worked in `kafka-python==1.3.5` but fails now in `kafka-python==1.4.6`.  

In `1.3.5` even though the documentation suggested that commits to ZooKeeper is not supported, we could see the commit reflected upon inspection with zkCli.

In `1.4.6` though, `KafkaConsumer` just doesn't function with the above configuration; because the early exit in `ConsumerCoordinator.poll()` prevents the coordinator from updating `next_auto_commit_deadline`. This stale value, in turn, causes `KafkaConsumer._message_generator()` to preempt the iteration of `self._fetcher` until `consumer_timeout` is reached.

Interestingly, if we simply allow for 0.8.1 in `ConsumerCoordinator.poll()` ie. `if self.group_id is None or self.config['api_version'] < (0, 8, 1):` then everything seems fine. Messages are yielded, and commits are written to ZK.

I must be missing something... Can anyone shed light to this?
Thanks thanks!

PR #1832, opened 2019-06-03
Allow the coordinator to auto-commit for api_version==0.8.1

justecorruptio:
Raised once in [https://github.com/dpkp/kafka-python/issues/1594](https://github.com/dpkp/kafka-python/issues/1594)

We're using `api_version=(0, 8, 1)` and `enable_auto_commit=True`. This had worked in `kafka-python==1.3.5` but fails now in `kafka-python==1.4.6`.

In 1.3.5 even though the documentation suggested that commits to ZooKeeper is not supported, we could see the commit reflected upon inspection with zkCli.

In 1.4.6 though, `KafkaConsumer` just doesn't function with the above configuration; because the early exit in `ConsumerCoordinator.poll()` prevents the coordinator from updating `next_auto_commit_deadline`. This stale value, in turn, causes `KafkaConsumer._message_generator()` to preempt the iteration of `self._fetcher` until `consumer_timeout` is reached.

If we allow for 0.8.1 in `ConsumerCoordinator.poll()`, then everything seems fine. Messages are yielded, and commits are written to ZK.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1832)

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -256,7 +256,7 @@ class ConsumerCoordinator(BaseCoordinator):
         ensures that the consumer has joined the group. This also handles
         periodic offset commits if they are enabled.
         """
-        if self.group_id is None or self.config['api_version'] < (0, 8, 2):
+        if self.group_id is None:
             return
 
         self._invoke_completed_offset_commit_callbacks()


jeffwidman:
Also, I don't remember this far back... in `0.8.1`, was the client issuing the commits to zookeeper, or was it still issuing the commits to the broker and then the broker piped them through to zookeeper? If the latter, that would explain why this works... and in fact we may want to just remove the requirement altogether (at least at this point in the code)

justecorruptio:
Mmm, I'm not familiar with anything past `0.8.1` actually. But I haven't seen any issues, and `1.3.5` was working with this setup. 

Yes, `0.8.1` is where the writes to zookeeper are from the broker instead of the client.

jeffwidman:
> Yes, 0.8.1 is where the writes to zookeeper are from the broker instead of the client.

If that's the case, I'd prefer to simply change this to `if self.group_id is None:` and drop the version check altogether.

If you do that, I'll happily merge... this is very low risk as it won't change behavior for anyone > `0.8.1`, and that's the vast majority of our users.

justecorruptio:
aye aye!

jeffwidman:
Thank you again!

PR #1845, opened 2019-06-19
Break consumer operations into request / response methods

jeffwidman:
This breaks some of the consumer operations into request generation /
response parsing methods.

The public API does not change. However, this allows power users who are
willing to deal with risk of private methods changing under their feet
to decouple generating the message futures from processing their
responses. In other words, you can use these to fire a bunch of requests
at once and delay processing the responses until all requests are fired.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1845)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -349,7 +349,7 @@ class KafkaAdminClient(object):
             # one of these attributes and that they always unpack into
             # (topic, error_code) tuples.
             topic_error_tuples = (response.topic_errors if hasattr(response, 'topic_errors')
-                else response.topic_error_codes)
+                    else response.topic_error_codes)
             # Also small py2/py3 compatibility -- py3 can ignore extra values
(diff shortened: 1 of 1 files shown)


jeffwidman:
Thanks to @Lou-Cipher for his excellent work in #1807 that enabled this. 

At some point this idea of decoupling message generation from response parsing could be further extended to `_find_group_coordinator_id()` and some of the other methods, but I thought this PR was already big enough and I don't personally need that yet either...

jeffwidman:
Using this as part of a callback-based-approach to fetching all consumer offsets from a cluster of 9 brokers with several hundred consumer groups dropped collection time. I went from ~242 seconds down to ~2 seconds, but further testing revealed most of the gain was from https://github.com/dpkp/kafka-python/pull/1823... details in https://github.com/dpkp/kafka-python/pull/1823#issuecomment-503834550.

PR #1851, opened 2019-06-28
conn: fix BrokerConnection unreleased lock issues

melor:
Under some error scenarios the internal self._lock would be left
unreleased, which could cause e.g. connections hanging forever in
close(). Use "with self._lock:" in all cases and switch to RLock so that
close() works when called from within the locked code block.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1851)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -273,8 +273,10 @@ class BrokerConnection(object):
         # However, to avoid pushing responsibility for maintaining
         # per-connection locks to the upstream client, we will use this lock to
         # make sure that access to the protocol buffer is synchronized
-        # when sends happen on multiple threads
-        self._lock = threading.Lock()
+        # when sends happen on multiple threads.
+        # RLock is used so that all code requiring a lock can be safely surrounded by a "with self._lock:", yet
+        # allowing calling self.close() inside those blocks
+        self._lock = threading.RLock()
 
         # the protocol parser instance manages actual tracking of the
         # sequence of in-flight requests to responses, which should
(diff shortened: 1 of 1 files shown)


jeffwidman:
Do you know which error scenarios were triggering this for you?

Any way to add a test to simulate those and make sure this doesn't creep back in in the future?

melor:
> Do you know which error scenarios were triggering this for you?
> 
> Any way to add a test to simulate those and make sure this doesn't creep back in in the future?

Not sure what exactly triggered the problem. The clients were consuming messages normally, we then had a connectivity problem with the Kafka cluster and a number of clients were blocked after that and not consuming any longer. Debugger showed the clients were blocked in a close() call waiting for the self._lock.

Some possibilities:

Missing release() https://github.com/dpkp/kafka-python/pull/1851/files#diff-fd7bc8dbfa82fc1f3b8c97cdc1ab0f49R629
Missing release() https://github.com/dpkp/kafka-python/pull/1851/files#diff-fd7bc8dbfa82fc1f3b8c97cdc1ab0f49R677
any unexpected exception in this block https://github.com/dpkp/kafka-python/pull/1851/files#diff-fd7bc8dbfa82fc1f3b8c97cdc1ab0f49L691
any unexpected exception in this block https://github.com/dpkp/kafka-python/pull/1851/files#diff-fd7bc8dbfa82fc1f3b8c97cdc1ab0f49L942

dpkp:
ugh -- great find. This should be fixed stat and is worth a new release. I do prefer the simple Lock here for performance reasons, but also because I think there are potentially some related lock contention issues (risk of deadlock) if we hold the lock when calling, in conn.close(), the client state_change_callback and failing all pending futures. Since these calls may require the client lock, I feel like we could end up in a deadlock where one thread holds the conn lock during close and is blocked waiting to acquire the client lock, while another thread holds the client lock and is blocked waiting for the conn lock....

So because of that I would prefer that we not resort to switching the an RLock.

jeffwidman:
We just got bit badly by this... had a few minutes of networking issues resulting in  all our consumers hung and not reconnecting

dpkp:
merged #1895

PR #1854, opened 2019-06-29
Fix conn lock release for gssapi/oauth

dpkp:
Simplified version of #1851 -- avoiding RLock due to concern of deadlock / client lock contention in close().

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1854)


melor:
This won't fix the problem we bumped into (#1851), which was a `TimeoutError(110, 'Connection timed out')` in consumption / recv code. It would probably be safest to surround all code following any lock acquire with a `try: ... except: self._lock.release(); raise` if the `with` / `RLock` based approach is not acceptable.

PR #1869, opened 2019-07-28
Fix describe config for multi-broker clusters

jlandersen:
Currently all describe config requests are sent to "least loaded node". Requests for broker configs must, however, be sent to the specific broker, otherwise an error is returned. Only topic requests can be handled by any node.

This changes the logic to send all describe config requests to the specific broker.

Same should be done for alter configs (there is already a comment stating this should be done - I'll be happy to send a PR for this soon'ish).

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1869)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -826,7 +826,7 @@ class KafkaAdminClient(object):
         self._wait_for_futures(futures)
 
         # Use one of the results as the general response and add all other resources to it
-        response = copy.copy(futures[0])
+        response = copy.copy(futures[0].value)
         response.resources = []
 
         for future in futures:
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -820,19 +820,10 @@ class KafkaAdminClient(object):
                 ))
         else:
             raise NotImplementedError(
-                "Support for DescribeConfigs v{} has not yet been added to KafkaAdminClient."
-                    .format(version))
+                "Support for DescribeConfigs v{} has not yet been added to KafkaAdminClient.".format(version))
 
         self._wait_for_futures(futures)
-
-        # Use one of the results as the general response and add all other resources to it
-        response = copy.copy(futures[0].value)
-        response.resources = []
-
-        for future in futures:
-            response.resources.extend(future.value.resources)
-
-        return response
+        return [f.value for f in futures]
 
     @staticmethod
     def _convert_alter_config_resource_request(config_resource):
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -5,6 +5,7 @@ import copy
 import logging
 import socket
 
+from . import ConfigResourceType
 from kafka.vendor import six
 
 from kafka.client_async import KafkaClient, selectors
@@ -763,28 +764,74 @@ class KafkaAdminClient(object):
             supported by all versions. Default: False.
         :return: Appropriate version of DescribeConfigsResponse class.
         """
+
+        # Break up requests by type - a broker config request must be sent to the specific broker.
+        # All other (currently just topic resources) can be sent to any broker.
+        broker_resources = []
+        topic_resources = []
+
+        for config_resource in config_resources:
+            if config_resource.resource_type == ConfigResourceType.BROKER:
+                broker_resources.append(self._convert_describe_config_resource_request(config_resource))
+            else:
+                topic_resources.append(self._convert_describe_config_resource_request(config_resource))
+
+        futures = []
         version = self._matching_api_version(DescribeConfigsRequest)
         if version == 0:
             if include_synonyms:
                 raise IncompatibleBrokerVersion(
                     "include_synonyms requires DescribeConfigsRequest >= v1, which is not supported by Kafka {}."
-                    .format(self.config['api_version']))
-            request = DescribeConfigsRequest[version](
-                resources=[self._convert_describe_config_resource_request(config_resource) for config_resource in config_resources]
-            )
+                        .format(self.config['api_version']))
+
+            if len(broker_resources) > 0:
+                for broker_resource in broker_resources:
+                    try:
+                        futures.append(self._send_request_to_node(
+                            int(broker_resource[1]),
+                            DescribeConfigsRequest[version](resources=[broker_resource])
+                        ))
+                    except ValueError:
+                        raise ValueError("Broker resource names must be an integer or a string represented integer")
+
+            if len(topic_resources) > 0:
+                futures.append(self._send_request_to_node(
+                    self._client.least_loaded_node(),
+                    DescribeConfigsRequest[version](resources=topic_resources)
+                ))
+
         elif version == 1:
-            request = DescribeConfigsRequest[version](
-                resources=[self._convert_describe_config_resource_request(config_resource) for config_resource in config_resources],
-                include_synonyms=include_synonyms
-            )
+            if len(broker_resources) > 0:
+                for broker_resource in broker_resources:
+                    try:
+                        futures.append(self._send_request_to_node(
+                            int(broker_resource[1]),
+                            DescribeConfigsRequest[version](
+                                resources=[broker_resource],
+                                include_synonyms=include_synonyms)
+                        ))
+                    except ValueError:
+                        raise ValueError("Broker resource names must be an integer or a string represented integer")
+
+            if len(topic_resources) > 0:
+                futures.append(self._send_request_to_node(
+                    self._client.least_loaded_node(),
+                    DescribeConfigsRequest[version](resources=topic_resources, include_synonyms=include_synonyms)
+                ))
         else:
             raise NotImplementedError(
                 "Support for DescribeConfigs v{} has not yet been added to KafkaAdminClient."
-                .format(version))
-        future = self._send_request_to_node(self._client.least_loaded_node(), request)
+                    .format(version))
 
-        self._wait_for_futures([future])
-        response = future.value
+        self._wait_for_futures(futures)
+
+        # Use one of the results as the general response and add all other resources to it
+        response = copy.copy(futures[0])
+        response.resources = []
+
+        for future in futures:
+            response.resources.extend(future.value.resources)
+
         return response
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


kafka/admin/client.py
```diff
@@ -484,28 +485,74 @@ def describe_configs(self, config_resources, include_synonyms=False):
...
             raise NotImplementedError(
                 "Support for DescribeConfigs v{} has not yet been added to KafkaAdminClient."
-                .format(version))
-        future = self._send_request_to_node(self._client.least_loaded_node(), request)
+                    .format(version))
+
+        self._wait_for_futures(futures)
+
+        # Use one of the results as the general response and add all other resources to it
+        response = copy.copy(futures[0])
```
dpkp: should this be `futures[0].value` ?

kafka/admin/client.py
```diff
@@ -484,28 +485,74 @@ def describe_configs(self, config_resources, include_synonyms=False):
...
+        response.resources = []
+
+        for future in futures:
+            response.resources.extend(future.value.resources)
```
dpkp: This seems a bit strange to me. If we want to combine responses into a singular return value, why keep the response structure? I would prefer to only return response data classes when they are unmodified.

kafka/admin/client.py
jlandersen: Yes it should :) fixed

kafka/admin/client.py
jlandersen: I agree, this was a bit odd. The reason for combining the responses into one was to keep the same response format. As it is now (with the above fix of copying the value attribute instead of the future object) An alternative would be to return a list of responses, but that would break contract. 

I am open for suggestions here :-)

jeffwidman: > An alternative would be to return a list of responses, but that would break contract.

Breaking changes are fine in `KafkaAdminClient`--I explicitly put a big warning in the docs about that. Not what we want long term, but right now there are a number of methods that are only partially stubbed out (like this one) and so I decided the simplest course was just say "you can use this, just beware it may change" as the other alternative was either not releasing it until fully complete, or tying our hands into awkward/broken interfaces.

So do whatever seems best, w/o regard to backward compatibility right now.

kafka/admin/client.py
```diff
@@ -484,28 +485,74 @@ def describe_configs(self, config_resources, include_synonyms=False):
...
+
+            if len(broker_resources) > 0:
+                for broker_resource in broker_resources:
+                    try:
+                        futures.append(self._send_request_to_node(
+                            int(broker_resource[1]),
+                            DescribeConfigsRequest[version](resources=[broker_resource])
+                        ))
+                    except ValueError:
+                        raise ValueError("Broker resource names must be an integer or a string represented integer")
```
jeffwidman: Hmm... By catching and then raising, you squash the original traceback... and there's enough code under your feet that the `ValueError` may come from somewhere unexpected. 

Would it make more sense to catch, modify the existing exception message string and then re-raise the original exception?

test/test_admin_integration.py
```diff
@@ -0,0 +1,66 @@
+import os
+
+from kafka import KafkaAdminClient
+from kafka.admin import ConfigResource, ConfigResourceType
+from test.fixtures import ZookeeperFixture, KafkaFixture
+from test.testutil import KafkaIntegrationTestCase, kafka_versions
+
+
+class TestAdminClientIntegration(KafkaIntegrationTestCase):
```
jeffwidman: Please do **not** use the `KafkaIntegrationTestCase` or anything that depends on `unittest`. 

It makes https://github.com/dpkp/kafka-python/pull/1196 much harder to merge.

There are quite a few `pytest`-based fixtures that I added recently, perhaps you can use one of those?

test/test_admin_integration.py
```diff
@@ -0,0 +1,66 @@
...
+        cls.server.close()
+        cls.zk.close()
+
+    def admin_client(self, **configs):
+        brokers = '%s:%d' % (self.server.host, self.server.port)
+        admin_client = KafkaAdminClient(
+            bootstrap_servers=brokers, **configs)
+        return admin_client
+
+    @kafka_versions('>=0.10.0')
```
jeffwidman: Use `pytest.mark.skipif()` instead of `@kafka_versions()`. Details in https://github.com/dpkp/kafka-python/pull/1887

jeffwidman:
nudge @jlandersen

jlandersen:
> 
> 
> nudge @jlandersen

apologies for the late reply, been a bit busy + traveling as well.
I'll try and get the changes in by the end of next week (still away at the moment). I pretty much agree with all of it (and not being concerned with backwards compat in the admin client makes it alot easier :) )

test/test_admin_integration.py
jlandersen: Makes sense - I can see the admin client integration test was added after my PR was started so i'll just move my tests into the new one

kafka/admin/client.py
jlandersen: I simplified it to just return a list of the describe config responses. For future work it could be converted into a more custom type similar to the ACL

jlandersen:
1) Tests are changed to follow the new structure
2) The response is now a list of each response received back rather than trying to merge it - should follow the general admin client interface more closely now. As described previously, a potential improvement for the future could be to convert it to a custom Config type that is more easily accessible.

One remaining comment is the handling of ValueError (see comment)

kafka/admin/client.py
jlandersen: I am inclined to keep it sort of as-is. If we are concerned about catching an underlying ValueError, we could do the parsing in isolation? Rewiring the error message isn't very pretty when maintaining both 2 and 3 compatibility (unless you know of a nice way :) )

jeffwidman: >  If we are concerned about catching an underlying ValueError, we could do the parsing in isolation?

Yes, let's do this. It's cleaner and only adds 3-4 lines of code.

The only py2/py3 way that I'm aware of is this:

```
except ValueError as exc:
    m = "Broker resource names must be an integer or a string represented integer"
    if six.PY2:
        exc.message += m
        raise
    else:
        raise ValueError(m) from exc
```

But I probably prefer parsing in isolation as that feels cleaner than munging them together.

To be clear, I don't see this as a major show stopper because it would be rarely a problem; however, I've been bitten before when debugging this library where we were accidentally hiding an exception under the covers so was unnecessarily painful to identify the actual root cause of the problem. This library already has enough complexity, let's not make debugging any more difficult!

test/test_admin_integration.py
```diff
@@ -80,3 +81,45 @@ def test_create_describe_delete_acls(kafka_admin_client):
 
     assert error is NoError
     assert len(acls) == 0
+
+
+@pytest.mark.skipif(env_kafka_version() < (0, 11), reason="Describe config features require broker >=0.11")
+def test_describe_configs_broker_resource_returns_configs(simple_client, kafka_admin_client):
```
jeffwidman: Oops, the `simple_client` fixture is going away in #1196... I'm working with @dpkp to figure out how we can get that merged sooner rather than later so we don't waste folks time.

In this case you can actually just get this info via `kafka_admin_client._client._cluster.brokers` so it should be an easy fixup and use the existing fixture.

test/test_admin_integration.py
```diff
@@ -80,3 +81,45 @@ def test_create_describe_delete_acls(kafka_admin_client):
...
+
+    assert len(configs) == 1
+    assert configs[0].resources[0][2] == ConfigResourceType.BROKER
+    assert configs[0].resources[0][3] == str(broker_id)
+    assert len(configs[0].resources[0][4]) > 1
+
+
+@pytest.mark.skipif(env_kafka_version() < (0, 11), reason="Describe config features require broker >=0.11")
+def test_describe_configs_topic_resource_returns_configs(simple_client, kafka_admin_client):
+    topic = simple_client.topics[0]
```
jeffwidman: I think this is available via `kafka_admin_client._client.topics()`?

test/test_admin_integration.py
```diff
@@ -80,3 +81,45 @@ def test_create_describe_delete_acls(kafka_admin_client):
...
+    assert len(configs) == 1
+    assert configs[0].resources[0][2] == ConfigResourceType.TOPIC
+    assert configs[0].resources[0][3] == topic
+    assert len(configs[0].resources[0][4]) > 1
+
+
+@pytest.mark.skipif(env_kafka_version() < (0, 11), reason="Describe config features require broker >=0.11")
+def test_describe_configs_mixed_resources_returns_configs(simple_client, kafka_admin_client):
+    topic = simple_client.topics[0]
+    broker_id = simple_client.brokers[0].nodeId
```
jeffwidman: see above comments re: dropping `simple_client` and how to get this info via `kafka_admin_client._client`

test/test_admin_integration.py
jlandersen: Close :) but by default it doesen't seem like the kafka admin fixture creates topics for tests, so added "topic" as parameter instead.

kafka/admin/client.py
jlandersen: Fixed :) Now extracts and validates the validate right before using it. Also added test for this scenario.

jlandersen:
@jeffwidman  changes should be in - let me know of any other things to change!

kafka/admin/client.py
jeffwidman: Thank you!

test/test_admin_integration.py
```diff
@@ -80,3 +81,57 @@ def test_create_describe_delete_acls(kafka_admin_client):
 
     assert error is NoError
     assert len(acls) == 0
+
+
+@pytest.mark.skipif(env_kafka_version() < (0, 11), reason="Describe config features require broker >=0.11")
```
jeffwidman: As a comment, since all these tests require a version, you could probably have organized them into a pytest class (which is primarily for organization, doesn't actually do most typical class/object style stuff) and then applied the decorator to the class and it would be applied to all the tests within that class. I'm not worried about it, more just an FYI if you ever work on stuff like this in the future...

jlandersen: Thanks for the heads up - i'll keep that in mind for future work!

PR #1876, opened 2019-08-03
Added a function to determine if bootstrap is successfully connected

Wayde2014:
Added a function to determine if bootstrap is successfully connected

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1876)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -377,6 +377,12 @@ class KafkaConsumer(six.Iterator):
             self._subscription.subscribe(topics=topics)
             self._client.set_topics(topics)
 
+    def bootstrap_connected(self):
+        """Return True if the bootstrap is connected."""
+        if self._client._bootstrap_fails > 0:
+            return False
+        return True
+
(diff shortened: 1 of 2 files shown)


dpkp:
I like the idea! Rather than adding more state to the KafkaProducer (self._client), maybe delegate this function to the Sender class, which owns the client instance?

Wayde2014:
I tested your optimized `bootstrap_connected` function, then found that it was not what I wanted.
However, when I changed the `bootstrap_connected` function as the following code, it work correctly.
```python
    def bootstrap_connected(self):
        """Return True if a bootstrap node is connected"""
        if self._bootstrap_fails > 0:
            return False
        return True
```

old code:
```python
    def bootstrap_connected(self):
        """Return True if a bootstrap node is connected"""
        for node_id in self._conns:
            if not self.cluster.is_bootstrap(node_id):
                continue
            if self._conns[node_id].connected():
                return True
        else:
            return False
```

PR #1879, opened 2019-08-12
tests: Use socket.SOCK_STREAM in assertions

iv-m:
socket.SOCK_STREAM is platform specific and on some
platforms (most notably on Linux on MIPS) does not
equal 1; so it's better to use the constant where
appropriate.

This change fixes the tests on my MIPS32 LE machine.

Signed-off-by: Ivan A. Melnikov <iv@altlinux.org>

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1879)


jeffwidman:
Thanks!

PR #1883, opened 2019-08-16
Load system default CAs when explicit CA not provided

### LINKED ISSUE #1795: Load default certs for SSL
Hi there,

I'm using a valid SSL certificate signed by Amazon for my Kafka connection and I made the following change to conn.py to pull in the default CA database if the user is validating the hostname but did not provide a CA cert.

```
[line 462 kafka/conn.py]
    if self.config['ssl_cafile']:
        log.info('%s: Loading SSL CA from %s', self, self.config['ssl_cafile'])
        self._ssl_context.load_verify_locations(self.config['ssl_cafile'])
        self._ssl_context.verify_mode = ssl.CERT_REQUIRED
+   if (not self.config['ssl_cafile']) and self.config['ssl_check_hostname']:
+      self._ssl_context.load_default_certs()
```
I would like to propose this change to allow users to use valid certs. Thanks!

iAnomaly:
Resolves #1795.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1883)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -464,6 +464,9 @@ class BrokerConnection(object):
                 log.info('%s: Loading SSL CA from %s', self, self.config['ssl_cafile'])
                 self._ssl_context.load_verify_locations(self.config['ssl_cafile'])
                 self._ssl_context.verify_mode = ssl.CERT_REQUIRED
+            else:
+                log.info('%s: Loading system default SSL CAs from %s', self, ssl.get_default_verify_paths())
+                self._ssl_context.load_default_certs()
             if self.config['ssl_certfile'] and self.config['ssl_keyfile']:
                 log.info('%s: Loading SSL Cert from %s', self, self.config['ssl_certfile'])
                 log.info('%s: Loading SSL Key from %s', self, self.config['ssl_keyfile'])


jeffwidman:
Thanks!

I noticed that #1795 limited loading the default to only when `ssl_check_hostname` was `True`.  

It makes sense to me to fall back to defaults, but I have not worked a lot with SSL, so did you already consider this? Just want to make sure it was an intentional choice and not an oversight.

iAnomaly:
@jeffwidman Thanks for confirming, but yes; it was an intentional choice: 

Even if the user decides to ignore the hostname check against the certificate, the handshake will still fail without a valid chain of trust if the root certificate authority is unknown (i.e. by loading the system default CAs).

iAnomaly:
@jeffwidman I need this fix released to resolve a downstream consumer. Is it possible for you or another maintainer to cut a patch release?

iAnomaly:
@dpkp Can you help here? Is it possible to cut a patch release with this fix? Thanks.

dpkp:
Included in 1.4.7, which is now released + live on pypi

PR #1886, opened 2019-08-22
Convert remaining `KafkaConsumer` tests to `pytest`

jeffwidman:
This makes it so the only remaining use of `unittest` is in the old
tests of the deprecated `Simple*` clients. All `KafkaConsumer` tests are
migrated to `pytest`.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1886)

### DIFF
--- a/test/test_consumer_integration.py
+++ b/test/test_consumer_integration.py
@@ -631,7 +631,7 @@ def test_kafka_consumer_max_bytes_simple(kafka_consumer_factory, topic, send_mes
     consumer = kafka_consumer_factory(
         auto_offset_reset='earliest', fetch_max_bytes=300)
     seen_partitions = set()
-    for i in range(10):
+    for i in range(90):
         poll_res = consumer.poll(timeout_ms=100)
         for partition, msgs in poll_res.items():
             for msg in msgs:
--- a/test/test_consumer_integration.py
+++ b/test/test_consumer_integration.py
@@ -4,7 +4,6 @@ import time
 
 from mock import patch
 import pytest
-from kafka.vendor import six
 from kafka.vendor.six.moves import range
 
 from . import unittest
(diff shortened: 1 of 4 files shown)


jeffwidman:
This is in preparation for #1193.

test/test_consumer_integration.py
```diff
@@ -547,242 +538,240 @@ def test_fetch_buffer_size(self):
...
+        auto_offset_reset='earliest', fetch_max_bytes=300)
+    seen_partitions = set()
+    for i in range(10):
+        poll_res = consumer.poll(timeout_ms=100)
+        for partition, msgs in poll_res.items():
+            for msg in msgs:
+                seen_partitions.add(partition)
+
+    # Check that we fetched at least 1 message from both partitions
+    assert seen_partitions == {TopicPartition(topic, 0), TopicPartition(topic, 1)}
```
jeffwidman: I'm seeing occasional test failures here.

I actually don't see the point of this test--when @tvoinarovskyi originally wrote it, it also tested that we were getting 3K bytes back, but that is no longer true in Kafka 0.11, so @dpkp removed that part of the test.

It appears to be verifying no partition starvation, and sometimes it works and sometimes it doesn't, but 10 poll cycles feels a little short to guarantee that we receive from both partitions... maybe increase the cycle count to 100 and not worry about the very infrequent failure?

Personally, I'd rather merge with `test_kafka_consumer_max_bytes_one_msg()` and verify that even if we are sending larger messages to two partitions that we can consume from both partitions and don't get starvation...

Perhaps this change should be saved for a separate PR and in this one just get the tests to pass by re-running the occasional failure and then merge as-is.

tvoinarovskyi: Yea, I don't remember the context under which I did this test. Feels like it doesn't test much to begin with. I think I wanted to verify that large messages will pass through even with a low max_bytes.
Feel free to merge it as you see fit. Sorry for the trouble again with this test.

jeffwidman: Looking at your original code, the intent was good, just Kafka 0.11 changed behavior a bit so this test got munged around. I'll try to merge it with the following test in a followup PR.

jeffwidman: https://github.com/dpkp/kafka-python/pull/1891

PR #1887, opened 2019-08-22
Cleanup handling of KAFKA_VERSION env var in tests

jeffwidman:
Now that we are using `pytest`, there is no need for a custom decorator
because we can use `pytest.mark.skipif()`.

This makes the code significantly simpler. In particular, dropping the
custom `@kafka_versions()` decorator is necessary because it uses
`func.wraps()` which doesn't play nice with `pytest` fixtures:
- https://github.com/pytest-dev/pytest/issues/677
- https://stackoverflow.com/a/19614807/770425

So this is a pre-requisite to migrating some of those tests to using `pytest` fixtures.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1887)

### DIFF
--- a/test/fixtures.py
+++ b/test/fixtures.py
@@ -4,9 +4,7 @@ import atexit
 import logging
 import os
 import os.path
-import random
 import socket
-import string
 import subprocess
 import time
 import uuid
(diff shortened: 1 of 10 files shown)


test/fixtures.py
```diff
@@ -19,29 +17,12 @@
...
 from kafka.protocol.admin import CreateTopicsRequest
 from kafka.protocol.metadata import MetadataRequest
+from test.testutil import env_kafka_version, random_string
 from test.service import ExternalService, SpawnedService
 
 log = logging.getLogger(__name__)
 
 
-def random_string(length):
-    return "".join(random.choice(string.ascii_letters) for i in range(length))
```
jeffwidman: I had to move this out of this file, at least for now, to solve a circular import problem. That wasn't present in the original version of working on this, likely because I'd already deleted a ton of code related to `Simple*` code paths... so it can be moved back later if semantically better here. But for now simplest to move it out to solve the import problem.

PR #1891, opened 2019-08-23
Merge two tests that are very similar

jeffwidman:
Previously the `test_kafka_consumer_max_bytes_simple()` was seeing
occasional test failures because it was doing only 10 iterations. And
much of the purpose of it was gutted when Kafka 0.11 came out and
changed the behavior. So this merges the two tests into one which should
be relatively straightforward.

Further discussion in https://github.com/dpkp/kafka-python/pull/1886/files#r316860737

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1891)

### DIFF
--- a/test/test_consumer_integration.py
+++ b/test/test_consumer_integration.py
@@ -623,48 +623,28 @@ def test_kafka_consumer__offset_commit_resume(kafka_consumer_factory, send_messa
 
 
 @pytest.mark.skipif(env_kafka_version() < (0, 10, 1), reason="Requires KAFKA_VERSION >= 0.10.1")
-def test_kafka_consumer_max_bytes_simple(kafka_consumer_factory, topic, send_messages):
-    send_messages(range(100, 200), partition=0)
-    send_messages(range(200, 300), partition=1)
+def test_kafka_consumer_max_bytes_one_msg(kafka_consumer_factory, topic, send_messages):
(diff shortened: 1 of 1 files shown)


tvoinarovskyi:
Hmm, seems to me like the test actually fails for a valid reason here. We don't have round robin partition rotation in fetch requests like Java does. We will probably have starvation. It's strange thou, as I clearly remember we added a random shuffle to partition order, maybe I've mistaken something.

tvoinarovskyi:
I may look into it later if you don't want to waste time on it. Just mark expected fail for now.

jeffwidman:
K, for now I kept at 10 iterations, just like originally. And I also remember seeing the comment in the code about random shuffle for partition. That probably explains why sometimes it succeeds, and other times it fails... it seems to be somewhat non-deterministic, although I haven't dug into it deeply.

Regardless, I did keep them merged, as I think the resulting single test easily covers both cases. The only thing we lost by merging was one of the tests had used the `poll()` interface and the other used the iterator (`next()`) interface. I don't think it's worrying about though, as the underlying bit this is trying to test AFAIK should be the same with both code paths.

I'll see what happens on the tests, and if necessary can mark it as `xfail`

jeffwidman:
Closing as I won't have time to work on this anytime soon. Feel free to reopen/pick it up if you have time/interest.

Issue #1888, opened 2019-08-22
KafkaConsumer runs 8x slower since v1.4.5


jutley:
My org uses the [prometheus-kafka-consumer-group-exporter](https://github.com/braedon/prometheus-kafka-consumer-group-exporter) project, which depends on this library. All this project does is read through the `__consumer_offsets` topic and generate metrics about the consumer groups. The latest version fails to read through the topic quickly enough to keep up. 

With some digging, I found that this performance change started at this commit: 8c0792581d8a38822c01b40f5d3926c659b0c439. I verified this with an experiment where I ran the following script against this commit, the commit preceding it (7a99013668b798aaa0acffcf382a7e48e7bd41c1), and master:
```
from kafka import KafkaConsumer
import schedule
import datetime

consumer_config = {
    'bootstrap_servers': <redacted>,
    'auto_offset_reset': 'earliest',
    'group_id': None
}

consumer = KafkaConsumer(
    '__consumer_offsets',
    **consumer_config
)

iterations=0

def print_status():
    print(datetime.datetime.now(), iterations)

schedule.every(5).seconds.do(print_status)

while True:
    for message in consumer:
        iterations = iterations + 1
        schedule.run_pending()
```

I ran three 2 minute trials against each commit to test the throughput of the consumer. Here are the results:
![image](https://user-images.githubusercontent.com/5660346/63535281-f33d6800-c4c5-11e9-8237-e70f954f69b1.png)

As you can see, this commit cause the consumer to run 8 times slower (2,426,802 messages vs. 300,187 messages). This has not improved since then.

I do not understand the details around this commit, but it has rendered this project unusuable on the latest version of kafka-python.

dpkp:
Hey, thanks for taking the time to investigate and run performance tests! We have some rudimentary benchmarking scripts in `benchmarks/` that confirm your results. I'll spend some time investigating and see if we can get this improved. I expect some performance regression, but not 8x. My hunch is that the move towards more decoupled send/receive is pushing into the legacy blocking-send code and causing us to be IO-bound where we should be (and have traditionally been) CPU bound. Indeed, a very quick flamegraph comparison suggests that the current master branch is IO bound and spends most of its time blocked in `_send_bytes_blocking` , while 1.4.4 is CPU bound on decoding + processing messages.

[...]

Issue #1672, opened 2018-12-10
Interval for background heartbeats unreliable


berupp:
Hi, 

This applies to 1.4.1, 1.4.3 and 1.4.4:

One of our python applications has issues with the consumer disconnecting from the broker: 
```
[W 181210 02:02:13 base:964] Heartbeat session expired, marking coordinator dead
[W 181210 02:02:13 base:698] Marking the coordinator dead (node 1002) for group cg_0: Heartbeat session expired.
```

After some investigation, I ran a packet capture to verify that the heartbeat is actually sent at the configured interval (10s in our case)

The results were quite surprising. Here is the timeline for Heartbeats:
16.63
26.72
36.79
51.91
72.02
112.20
132.33
142.43
177.62
187.69
197.77
222.94
233.01
243.10
283.40

Not only are the heartbeats completely inconsistent, there are gaps of > 40 seconds.
Our session_timeout was 30 seconds, which explained the consumer disconnects.

I raised the session_timeout to 3 minutes, but still eventually missed heartbeats during a soak, leading to consumer disconnect.

I ran a simultaneous capture of two other apps written in Golang and Java. Those have 3 second heartbeats configured and where consistently spot on.

Is this something that could be fixed, or simply a limitation due to GIL?

berupp:
Python 2.7.5

retry_backoff_ms is default (100)

```
consumer = KafkaConsumer(bootstrap_servers=self.kafka_url, group_id='cg_0',
                                     auto_offset_reset='earliest', session_timeout_ms=299999,
                                     heartbeat_interval_ms=10000, reconnect_backoff_ms=5000,
                                     reconnect_backoff_max_ms=30000)
```

xiumu-beta:
I have the same problem, I am so sad

berupp:
No, we simply increased the session_timeout to work around the issue for now.

I assume that the unreliable background 'thread' execution is due to python 2.7, and that upgrading the application to python 3 will fix or at least improve the condition.

I haven't had time to verify the approach. Maybe The maintainer could give his opinion on that theory.

dpkp:
This may be related to the lock contention between senders (here, heartbeat thread) and pollers (consumer iterator perhaps). See #981 for some more details. We've landed a few PRs to master to address this. If you are able to run/test with master, give it a try. Otherwise I believe this should be improved substantially in the next release.

jeffwidman:
@berupp was this solved by the latest `1.4.6` release?

[...]

PR #1902, opened 2019-09-16
Wrap consumer.poll() for KafkaConsumer iteration

dpkp:
Now that heartbeats have been moved to a background thread, the pythonic iterator can just wrap `consumer.poll()` and no longer needs to worry about breaking from iteration to support sending background requests. With this approach, some consumers may need to tweak `max_poll_records` and/or `max_poll_interval_ms`, such that consumer's average per-record processing time * max_poll_records < max_poll_interval_ms. Defaults remain at 500 max_poll_records and 5mins for max_poll_interval. This should be sufficient for the vast majority of consumers.

My local testing suggests that this change fixes consumer performance issues reported in #1888 .

Given the reduction in overall complexity, this change may also improve #1672 (though won't completely fix).

Because this is a relatively significant change to internals, I have added a configuration option to allow reverting back to the <= 1.4.6 iterator logic:
```
consumer = KafkaConsumer(bootstrap_servers=[...], legacy_iterator=True, ...)
```

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1902)

### DIFF
--- a/kafka/consumer/fetcher.py
+++ b/kafka/consumer/fetcher.py
@@ -292,7 +292,7 @@ class Fetcher(six.Iterator):
         raise Errors.KafkaTimeoutError(
             "Failed to get offsets by timestamps in %s ms" % (timeout_ms,))
 
-    def fetched_records(self, max_records=None):
+    def fetched_records(self, max_records=None, update_offsets=True):
         """Returns previously fetched records and updates consumed offsets.
 
         Arguments:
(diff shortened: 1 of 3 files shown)


dpkp:
> I question whether adding a feature flag is worth the trouble unless you truly think there's a significant chance it will need to be reverted. Folks can always revert to `1.4.6` if needed, although I suppose `1.4.7` will include a number of other changes so this makes it simpler to isolate the cause of problems.

Ok, I'm convinced -- I updated to make the new poll()-based iterator the default. If there is a problem, users can revert to prior implementation by setting `legacy_iterator=True` in KafkaConsumer.

> Looks like Travis is failing, may be related to #1891 and https://github.com/dpkp/kafka-python/pull/1886/files#r316860737.

I think this was related to deciding whether to update offsets in the fetcher (we need to do this for standard poll(), but not when wrapping with the iterator). I've pushed a fix and will see whether the next test run is clean.

Thanks for the review!

PR #1904, opened 2019-09-19
kafka-python: Fixed crc32c availability on non-intel architectures.

ossdev07:
requirements-dev.txt: The crc32c version mentioned in this file is 1.5, which fails to compile on non-intel archs. Refer https://github.com/ICRAR/crc32c/issues/6 . It has been fixed since version 1.6.

tox.ini: As mentioned in crc32c github page https://github.com/ICRAR/crc32c , all non-intel architectures have to use the software implementation of it since the hardware its specific to isn't available on non-intel archs. So, to successfully run the test cases, we need to set the environment variable `CRC32C_SW_MODE = auto` .

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1904)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,5 +12,5 @@ pylint==1.9.3
 pytest-pylint==0.12.3
 pytest-mock==1.10.0
 sphinx-rtd-theme==0.2.4
-crc32c==1.5
+crc32c==1.7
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Nice, thank you!

Issue #1907, opened 2019-09-25
Sender may busy loop when connection is in CONNECTING state


rikonen:
The following may happen:

1. kafka-python successfully fetches metadata for cluster indicating that leader for topic partition `X` is node `Y`
2. Client application starts producing data, including some that goes to topic partition `X`
3. kafka-python successfully connects to `Y` and starts sending batches to node `Y` as soon as they're completed
4. kafka-python loses connection to `Y`. Client still continues producing data
5. kafka-python tries establishing connection to `Y` but the connection doesn't proceed beyond `CONNECTING` state in a while
6. New batch going to node `Y` completes.
7. `Sender.run_once()` gets node Y from `RecordAccumulator.ready()` but `self._client.is_ready()` returns `False` because there's no working connection. `run_once` discards node `Y` from the list of nodes to process and calls `self._client.connection_delay()` to determine how long to block in `self._client.poll()`. Because the connection to Y is in state `CONNECTING`, `self._client.connection_delay()` returns 0.
8. `run_once()` calls `self._client.poll()` with timeout zero, causing `poll()` to exit practically immediately.
9. `run()` calls `run_once()` again with no delay and sequence continues from step 7: kafka-python keeps on busy looping until connection to Y is successfully established, which may take very long time if the broker has some problem that prevents new connections from being established.

Log extract showing this behavior below:

```
2019-09-23T15:54:50.542310  hostname  kafka.conn            kafka-python-producer-2-network-thread INFO      <BrokerConnection node_id=71 host=10.0.1.100:13737 <connecting> [IPv4 ('10.0.1.100', 13737)]>: connecting to 10.0.1.100:13737 [('10.0.1.100', 13737) IPv4]
2019-09-23T15:54:59.819240  hostname  kafka.producer.sender  kafka-python-producer-2-network-thread DEBUG     Node 71 not ready; delaying produce of accumulated batch
2019-09-23T15:54:59.849336  hostname  kafka.producer.sender  kafka-python-producer-1-network-thread DEBUG     Node 71 not ready; delaying produce of accumulated batch
2019-09-23T15:54:59.865445  hostname  kafka.producer.sender  kafka-python-producer-1-network-thread DEBUG     Node 71 not ready; delaying produce of accumulated batch
2019-09-23T15:54:59.868538  hostname  kafka.producer.sender  kafka-python-producer-2-network-thread DEBUG     Node 71 not ready; delaying produce of accumulated batch
[...]

PR #1908, opened 2019-09-25
conn: Return non-zero delay when in CONNECTING state

rikonen:
If connection is in state CONNECTING and there is data ready to be sent
using value 0 for the delay results in busy loop because Sender calls
select with the value returned from here and then just keeps on calling
it again and again until connection completes or fails. In some cases
this can take a long time and if the connection failed it will just be
retried and the busy loop continues.

This return value was changed from 999999999 to 0 in commit 16c13f91 and
it doesn't look like it needs to be exactly zero for other logic to work
correctly.

Fixes https://github.com/dpkp/kafka-python/issues/1907

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1908)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -107,6 +107,10 @@ class BrokerConnection(object):
             server-side log entries that correspond to this client. Also
             submitted to GroupCoordinator for logging with respect to
             consumer group administration. Default: 'kafka-python-{version}'
+        connecting_delay_ms (int): Time to wait before sending data when a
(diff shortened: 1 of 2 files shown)


dpkp:
Going to close in favor of #1909 -- I think we can just rely on the selector to wake on the socket buffer becoming writeable, and not introduce a new configuration option.

dpkp:
Thanks for the issue report + PR though -- this is a great catch!

PR #1912, opened 2019-09-29
Send socket data via non-blocking IO with send buffer

dpkp:
This PR completes migration of network IO writes to the asynchronous event loop. Each BrokerConnection object maintains a `send_buffer` to capture any bytes returned from KafkaProtocol but not yet placed on the socket buffer via `sock.send` .

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1912)


jeffwidman:
🍰

PR #1918, opened 2019-10-02
Enable SCRAM-SHA-256 and SCRAM-SHA-512 for sasl

swenzel:
closes #1882

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1918)

PR #1920, opened 2019-10-02
SASL SCRAM support

wsmith-twc:
It looks like someone beat me to a pull request by a few hours, but I thought I'd submit this in case this code is useful.

I've used against Kafka 2.2.1 and 2.3.0 with Python 2.7.15 and 3.6.8. I've only used security mechanism of SCRAM-SHA-512 (not SCRAM-SHA-256). I've tried it with security protocols of SASL_PLAINTEXT and SASL_SSL.

The code re-uses sasl_plain_username and sasl_plain_password. That makes these names a little inaccurate, but having separate sasl_scram_username and sasl_scram_password didn’t seem like a great option, either. Another option would be to introduce sasl_username and sasl_password that both mechanisms use and deprecate sasl_plain_*

I didn’t follow the exact style of conn._try_authentiate_plain and conn._try_authenticate_gssapi because there are more errors that can occur. I relied on exceptions more than in those functions.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1920)

### DIFF
--- a/kafka/conn.py
+++ b/kafka/conn.py
@@ -1,11 +1,15 @@
 from __future__ import absolute_import, division
 
+import base64
 import collections
 import copy
 import errno
+import hashlib
+import hmac
 import io
 import logging
-from random import shuffle, uniform
+from random import shuffle, uniform, SystemRandom
+import re
 
 # selectors in stdlib as of py3.4
 try:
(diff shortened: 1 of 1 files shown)


dpkp:
Thanks!! I'm going to go with #1918 primarily because it includes sasl fixture support + tests, which also benefit existing sasl plain mechanisms. I agree with your thinking on `sasl_plain_*` configs. Thanks again for working on this.

PR #1921, opened 2019-10-03
Update docstring to match conn.py's

dabcoder:
i.e: https://github.com/dpkp/kafka-python/blob/1.4.7/kafka/conn.py#L135.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1921)


jeffwidman:
Thanks!

PR #1925, opened 2019-10-11
Remove some unused code

jeffwidman:
Currently these are `TODO`'s, but hopefully @dpkp can provide backstory on whether to remove these or leave them in...

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1925)


kafka/producer/buffer.py
```diff
@@ -114,6 +114,7 @@ def queued(self):
         with self._lock:
             return len(self._waiters)
 
+# TODO should this be removed or finished?
```
dpkp: This is fine to remove. Memory allocation is not a bottleneck as far as I can tell.

kafka/protocol/struct.py
```diff
@@ -65,6 +65,7 @@ def __eq__(self, other):
                 return False
         return True
 
+# TODO should this be removed or finished?
```
dpkp: Yep -- remove

dpkp: Yep -- remove

kafka/util.py
```diff
@@ -67,6 +67,7 @@ class Dict(dict):
     pass
 
 
+# TODO what is the purpose of this? It has been unused since it was originally added
```
dpkp: I was thinking about using this to help prevent circular references when calling instance methods with `atexit` . Currently we only do that in the producer, and we use weakproxy directly. If we ever needed to do a similar cleanup for other classes then this method might be helpful. But I don't mind if you delete it.

PR #1940, opened 2019-10-30
[kafka]:crc32c software fallback for non-intel machines.

ossdev07:
A default software fallback is added for non Intel machines, which was previously giving ImportError.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1940)

### DIFF
--- a/kafka/record/util.py
+++ b/kafka/record/util.py
@@ -2,6 +2,12 @@ import binascii
 
 from kafka.record._crc32c import crc as crc32c_py
 try:
+    """ Environment variable CRC32C_SW_MODE needs to be declared to auto 
+    in order to use the software implementation of crc32c in case of
+    non-intel architectures, which otherwise would fail as ImportError.
+    """
+    import os
+    os.environ['CRC32C_SW_MODE'] = 'auto'
     from crc32c import crc32 as crc32c_c
 except ImportError:
     crc32c_c = None


ossdev07:
Kindly look into this PR.

ossdev07:
Kindly look into this PR.

ossdev07:
Kindly look into this PR.

jeffwidman:
Related `crc32c` discussion: 
- https://github.com/ICRAR/crc32c/issues/11
- https://github.com/ICRAR/crc32c/issues/12
- https://github.com/ICRAR/crc32c/pull/13

To clarify, since it took me a little bit to understand what this does: This makes it so that if the `crc32c` python package is present, it will always be successfully imported. If this happens on an intel architecture, it will use the chip's HW instruction. If not, it will use a C-based software implementation.

A few thoughts:
1. It appears the `crc32c` package is going to change the import logic which would make this OS environment value unnecessary. It may take a little bit, so as a temporary workaround this is probably fine for now until https://github.com/ICRAR/crc32c/issues/12 is fixed.
2. It'd be nice to see benchmarks of their pure C software implementation versus the vendored python implementation we currently ship in `kafka-python/kafka/record/_crc32c.py`. I suspect the C version faster, but like to see actual data.
3. If it is indeed faster, than ideally we could vendor that, but because it's C rather than python, I think that would force compilation upon install, which we probably want to avoid... but I haven't investigated this deeply.

Do I understand corrctly @ossdev07 ?

ossdev07:
> To clarify, since it took me a little bit to understand what this does: This makes it so that if the crc32c python package is present, it will always be successfully imported. If this happens on an intel architecture, it will use the chip's HW instruction. If not, it will use a C-based software implementation. 

> 

Correct !

> It appears the crc32c package is going to change the import logic which would make this OS environment value unnecessary. It may take a little bit, so as a temporary workaround this is probably fine for now until ICRAR/crc32c#12 is fixed. 

> 
I don't believe the env variable is going to be unnecessary since it also provides the flexibility of using the software implementation even if you have h/w capabilities ; some users might need that in order to review the performances of their code on both intel and non-intel architectures. 
Also maintainer has proposed a timeline which isn't in line with our timelines and hence we're forced to use the workaround.

> It'd be nice to see benchmarks of their pure C software implementation versus the vendored python implementation we currently ship in kafka-python/kafka/record/_crc32c.py. I suspect the C version faster, but like to see actual data. 

> 
I suspect the same, but in either case we wouldnt be able to know if the `import crc32c` returned us the c-software based or the hardware based solution, and the actual problem of test case failing wouldn't even happen if the import is successful.

> If it is indeed faster, than ideally we could vendor that, but because it's C rather than python, I think that would force compilation upon install, which we probably want to avoid... but I haven't investigated this deeply. 

> 
I haven't investigated it as well.

But we really need this PR to be pulled in asap in order to make kafka package 100% usable on our arm machine, without any external env variables dependency.

rtobar:
Regarding benchmarks, please see the logs from the Travis and AppVeyor builds, they always include a test run that measures the performance of the package with hw and sw. Don't have a laptop now to paste proper links, but should be easy to reach the builds from the badges on the README. From memory (but might be wrong), the sw impl runs at ~1.5, ~2 GB/s.

ossdev07:
How soon can we expect this PR to be pulled and merged ?

ossdev07:
Hi Maintainers,
Can you please review this PR ?

ossdev07:
This PR is now superseded by the latest changes in the master branch of https://github.com/ICRAR/crc32c/commits/master dated 25th November 2019, which implement a more thorough change in the import logic. I'm thus closing this PR.

Issue #1929, opened 2019-10-12
Intermittent test failure in test_describe_configs_topic_resource_returns_configs


jeffwidman:
Noticed that two test runs failed on one of the travis builds:
- https://travis-ci.org/dpkp/kafka-python/jobs/596753723
- https://travis-ci.org/dpkp/kafka-python/jobs/596753725

```_____________ test_describe_configs_topic_resource_returns_configs _____________
topic = 'test_describe_configs_topic_resource_returns_configs_nXpHTvKOpm'
kafka_admin_client = <kafka.admin.client.KafkaAdminClient object at 0x7f47e3367250>
    @pytest.mark.skipif(env_kafka_version() < (0, 11), reason="Describe config features require broker >=0.11")
    def test_describe_configs_topic_resource_returns_configs(topic, kafka_admin_client):
        """Tests that describe config returns configs for topic
        """
        configs = kafka_admin_client.describe_configs([ConfigResource(ConfigResourceType.TOPIC, topic)])
    
        assert len(configs) == 1
        assert configs[0].resources[0][2] == ConfigResourceType.TOPIC
        assert configs[0].resources[0][3] == topic
>       assert len(configs[0].resources[0][4]) > 1
E       assert 0 > 1
E        +  where 0 = len([])
test/test_admin_integration.py:108: AssertionError
```

I suspect (but have not verified) that this is a race condition where the topic or config resource is in the midst of being created, but simply hasn't completed by the time it's checked.

dpkp:
Let's skip this test until we have a solution. I've also seen several of these.

PR #1948, opened 2019-11-08
Admin protocol updates

TylerLubeck:
I've been digging in to the admin client and found myself adding extra api version support

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1948)

### DIFF
--- a/kafka/protocol/admin.py
+++ b/kafka/protocol/admin.py
@@ -319,6 +319,33 @@ class DescribeGroupsResponse_v1(Response):
     )
 
 
+class DescribeGroupsResponse_v2(Response):
+    API_KEY = 15
+    API_VERSION = 2
+    SCHEMA = DescribeGroupsResponse_v1.SCHEMA
+
+
+class DescribeGroupsResponse_v3(Response):
+    API_KEY = 15
+    API_VERSION = 3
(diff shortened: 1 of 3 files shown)


kafka/protocol/admin.py
```diff
@@ -565,6 +708,28 @@ class DescribeConfigsResponse_v1(Response):
...
+        ('resources', Array(
+            ('error_code', Int16),
+            ('error_message', String('utf-8')),
+            ('resource_type', Int8),
+            ('resource_name', String('utf-8')),
+            ('config_entries', Array(
+                ('config_names', String('utf-8')),
+                ('config_value', String('utf-8')),
+                ('read_only', Boolean),
+                ('config_source', Int8),
```
TylerLubeck: It looks like this was renamed from `is_default` to `config_source`, and the type changed from `Boolean` to `Int8` in v1 as well, but I didn't include that fix here. Should I?

TylerLubeck:
I know how this sounds, but I can't for the life of me reproduce the test failure locally. Any suggestions?

dpkp:
Test failure is known to be flakey and has been marked as xfail in master (see #1929)

PR #1951, opened 2019-11-13
Implement methods to convert a Struct object to a pythonic object

TylerLubeck:
There are a handful of places that call out for converting response tuples to more pythonic objects. This is a stab at it - I'm happy to make changes if this isn't what you were thinking

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1951)


kafka/protocol/api.py
```diff
@@ -65,3 +68,30 @@ def API_VERSION(self):
...
         pass
+
+    def to_object(self):
+        return _to_object(self.SCHEMA, self.fields)
+
+
+def _to_object(schema, data):
+    obj = {}
+    for idx, (name, _type) in enumerate(zip(schema.names, schema.fields)):
+        if isinstance(data, dict):
```
dpkp: why not check for `Struct` and use the existing attributes set in `__dict__` , as opposed to creating a duplicate dictionary for the same data (struct.fields) ?

kafka/protocol/struct.py
```diff
@@ -12,14 +12,17 @@ class Struct(AbstractType):
     SCHEMA = Schema()
 
     def __init__(self, *args, **kwargs):
+        self.fields = {}
```
dpkp: I don't think we need this -- the same data is in `__dict__` and could be accessed via `getattr`. Or you might add a helper, like `get_field(name)` that returns `__dict__[name]`

dpkp:
I like it! Just a few minor comments re: how to handle Struct fields

jeffwidman:
nudge @TylerLubeck

TylerLubeck:
Hey sorry, I've been away from Kafka for a while now. I think I've got this set up to not require double storing all the attributes. One of the reasons I had gone in that direction before was to avoid accidental use of other, non-schema items in `__dict__`. You'll see that still with the `KeyError`, but I think that's a more reasonable guard than what I had before.

PR #1987, opened 2020-01-17
Fix slots usage and use more slots

carsonip:
Use empty slots for ABC classes, otherwise classes which inherit from them will still have `__dict__`. Also use `__slots__` for more classes.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1987)

### DIFF
--- a/kafka/record/abc.py
+++ b/kafka/record/abc.py
@@ -4,6 +4,7 @@ import abc
 
 class ABCRecord(object):
     __metaclass__ = abc.ABCMeta
+    __slots__ = ()
 
     @abc.abstractproperty
     def offset(self):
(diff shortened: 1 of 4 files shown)


carsonip:
This saves ~200 bytes per object, because the object does not contain a `__dict__`, and there are also a marginal performance gain:
```
python benchmarks/record_batch_read.py
```
before:
```
.....................
batch_read_v0: Mean +- std dev: 1.23 ms +- 0.06 ms
.....................
batch_read_v1: Mean +- std dev: 1.28 ms +- 0.10 ms
.....................
batch_read_v2: Mean +- std dev: 961 us +- 82 us
```
after:
```
.....................
batch_read_v0: Mean +- std dev: 1.18 ms +- 0.05 ms
.....................
batch_read_v1: Mean +- std dev: 1.20 ms +- 0.06 ms
.....................
batch_read_v2: Mean +- std dev: 950 us +- 55 us
```

PR #1995, opened 2020-02-06
Use the controller for topic metadata requests

### LINKED ISSUE #1994: describe topics needs to speak to the controller
https://github.com/dpkp/kafka-python/pull/1993#discussion_r376085225

@TylerLubeck is putting together PR but opening this ticket so we don't lose track of this.

TylerLubeck:
Closes #1994

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/1995)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -473,7 +473,7 @@ class KafkaAdminClient(object):
         return response
 
 
-    def _get_cluster_metadata(self, topics=None, auto_topic_creation=False):
+    def _get_cluster_metadata(self, topics=None, auto_topic_creation=False, use_controller=False):
         """
         topics == None means "get all topics"
         """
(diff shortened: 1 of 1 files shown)


kafka/admin/client.py
```diff
@@ -473,7 +473,7 @@ def delete_topics(self, topics, timeout_ms=None):
         return response
 
 
-    def _get_cluster_metadata(self, topics=None, auto_topic_creation=False):
```
jeffwidman: it'd be cleaner if we could specify the broker ID here, and then the caller can pass in a function that identifies the controller... but looks like that's not how `self._send_request_to_controller()` is wired up. And it'd be hard to change because of needing to handle retries. So let's just go with this as-is (especially since it's non-public).

jeffwidman:
I'll wait for travis and then will merge.

Be nice to have tests for this describe/list topics stuff, but I'm going to merge this one regardless since it's a clear bug.

PR #2000, opened 2020-02-16
KAFKA-8962: Use least_loaded_node() for describe_topics()

jeffwidman:
In KAFKA-8962 the `AdminClient.describe_topics()` call was changed from
using the controller to using the `least_loaded_node()`:
https://github.com/apache/kafka/commit/317089663cc7ff4fdfcba6ee434f455e8ae13acd#diff-6869b8fccf6b098cbcb0676e8ceb26a7R1540

As a result, no metadata request/response processing needs to happen
through the controller, so it's safe to remove the custom
error-checking. Besides, I don't think this error-checking even added
any value because AFAIK no metadata response would return a
`NotControllerError` because the recipient broker wouldn't realize the
metadata request was intended for only the controller.

Originally our admin client was implemented using the least-loaded-node,
then later updated to the controller. So updating it back to
least-loaded node is a simple case of reverting the associated commits.

This reverts commit 7195f0369c7dbe25aea2c3fed78d2b4f772d775b.

This reverts commit 6e2978edee9a06e9dbe60afcac226b27b83cbc74.

This reverts commit f92889af79db08ef26d89cb18bd48c7dd5080010.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2000)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -376,16 +376,11 @@ class KafkaAdminClient(object):
             # In Java, the error field name is inconsistent:
             #  - CreateTopicsResponse / CreatePartitionsResponse uses topic_errors
             #  - DeleteTopicsResponse uses topic_error_codes
-            #  - MetadataResponse uses topics[].error_code
-            topic_error_tuples = []
-            if hasattr(response, 'topic_errors'):
-                topic_error_tuples.extend(response.topic_errors)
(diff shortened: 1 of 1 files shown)


jeffwidman:
cc @TylerLubeck @jtribble for FYI.

PR #2007, opened 2020-02-28
Add Envy functionality

Tim-Willard:
Hello!
​
We (@magmastonealex, @omstrumpf, @gbateman, @Tim-Willard) have been working on a tool called ENVy.
​
ENVy is an environment manager to make working on disparate projects easier. You don't need to worry about what a system package is called on your Linux distribution, or figure out how to build for Mac from scratch.
​
Setup with ENVy is as simple as typing `envy up` for any project you want to work on. Running `envy nuke` removes all traces of that environment.
With ENVy, you don't need to hunt down a lint command or how you run the tests. `envy -h` shows you all the commands you can run - or just jump into a shell with `envy shell`.
​
You can read more about ENVy [here](http://envy-project.github.io/), but the short version is that we think it'll make it easier for contributors to help out. 
​
This PR adds an `ENVyfile`, the file that describes a development environment and contains the commands to run. This should help manage the two python versions that kafka-python supports, without requiring the contributor to install both on their host machine. We used kafka-python while developing in order to test ENVy, and we hoped you'd consider committing this file to help out casual contributors.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2007)

### DIFF
--- a/.envyfile
+++ b/.envyfile
@@ -1,0 +1,30 @@
+environment:
+  base:
+    image: python:3.7.3
+  system-packages:
+    - recipe: python-dev
+    - recipe: libsnappy-dev
+  setup-steps:
+    - name: python-deps
+      label: "Installing python dependencies"
+      type: script
+      as_user: false
+      run:
+        - "pip install -r requirements-dev.txt"
+        - "pip install -Ue ."
+      triggers:
+        files:
+          - requirements-dev.txt
+actions:
+  - name: test37
+    script: 'make test37'
+    help: 'Lint and run tests with Python 3.7'
+  - name: test27
+    script: 'make test27'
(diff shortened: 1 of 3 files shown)


jeffwidman:
Thanks for the hard work.

I'm personally a bit hesitant to add this, as it tries to make things more convenient but at the same time is yet another tool... I find that developing with `tox` is generally good enough for my needs, and I don't worry about installing all versions of python locally, rather I just push up and let Travis worry about it if I missed something.

But I'm not strongly opposed if another maintainer feels differently. And if the project becomes popular, than I'd be open to adding it so that others can use it for developing... but until it becomes popular or solves a specific need of mine, then I'd rather hold off...

PR #2020, opened 2020-03-15
Add `log_start_offset` to message protocol parsing

gabriel-tincu:
As mentioned in the commit, these need to come all together, as the future code also implements logic that will unpack depending on version. The other PR would then be the ZSTD code itself

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2020)

### DIFF
--- a/kafka/producer/sender.py
+++ b/kafka/producer/sender.py
@@ -195,13 +195,18 @@ class Sender(threading.Thread):
 
             for topic, partitions in response.topics:
                 for partition_info in partitions:
+                    error_message = None
                     if response.API_VERSION < 2:
                         partition, error_code, offset = partition_info
                         ts = None
-                    else:
+                    elif 2 <= response.API_VERSION <= 4:
                         partition, error_code, offset, ts = partition_info
+                    elif 5 <= response.API_VERSION <= 7:
+                        partition, error_code, offset, ts, log_start_offset = partition_info
+                    else:
+                        partition, error_code, offset, ts, log_start_offset, _, error_message = partition_info
                     tp = TopicPartition(topic, partition)
-                    error = Errors.for_code(error_code)
+                    error = error_message or Errors.for_code(error_code)
                     batch = batches_by_partition[tp]
                     self._complete_batch(batch, error, offset, ts)
 
@@ -213,7 +218,7 @@ class Sender(threading.Thread):
             for batch in batches:
                 self._complete_batch(batch, None, -1, None)
 
-    def _complete_batch(self, batch, error, base_offset, timestamp_ms=None):
+    def _complete_batch(self, batch, error, base_offset, timestamp_ms=None, log_start_offset=None):
         """Complete or retry the given batch of records.
 
         Arguments:
@@ -221,6 +226,7 @@ class Sender(threading.Thread):
             error (Exception): The error (or None if none)
             base_offset (int): The base offset assigned to the records if successful
             timestamp_ms (int, optional): The timestamp returned by the broker for this batch
+            log_start_offset (int): The start offset of the log at the time this produce response was created
         """
         # Standardize no-error to None
         if error is Errors.NoError:
@@ -240,7 +246,7 @@ class Sender(threading.Thread):
                 error = error(batch.topic_partition.topic)
 
             # tell the user the result of their request
-            batch.done(base_offset, timestamp_ms, error)
+            batch.done(base_offset, timestamp_ms, error, log_start_offset)
             self._accumulator.deallocate(batch)
             if error is not None:
                 self._sensors.record_errors(batch.topic_partition.topic, batch.record_count)
@@ -293,7 +299,15 @@ class Sender(threading.Thread):
             produce_records_by_partition[topic][partition] = buf
 
         kwargs = {}
-        if self.config['api_version'] >= (0, 11):
+        if self.config['api_version'] >= (2, 1):
+            version = 7
+        elif self.config['api_version'] >= (2, 0):
+            version = 6
+        elif self.config['api_version'] >= (1, 1):
+            version = 5
+        elif self.config['api_version'] >= (1, 0):
+            version = 4
+        elif self.config['api_version'] >= (0, 11):
             version = 3
             kwargs = dict(transactional_id=None)
         elif self.config['api_version'] >= (0, 10):
--- a/kafka/producer/sender.py
+++ b/kafka/producer/sender.py
@@ -228,7 +228,7 @@ class Sender(threading.Thread):
             base_offset (int): The base offset assigned to the records if successful
             timestamp_ms (int, optional): The timestamp returned by the broker for this batch
             log_start_offset (int): The start offset of the log at the time this produce response was created
-            global_error (Exception): The summarising error message
+            global_error (str): The summarising error message
         """
         # Standardize no-error to None
         if error is Errors.NoError:
--- a/kafka/producer/sender.py
+++ b/kafka/producer/sender.py
@@ -196,6 +196,7 @@ class Sender(threading.Thread):
             for topic, partitions in response.topics:
                 for partition_info in partitions:
                     global_error = None
+                    log_start_offset = None
                     if response.API_VERSION < 2:
                         partition, error_code, offset = partition_info
                         ts = None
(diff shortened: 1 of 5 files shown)


kafka/producer/sender.py
```diff
@@ -195,13 +195,18 @@ def _handle_produce_response(self, node_id, send_time, batches, response):
...
-                    else:
+                    elif 2 <= response.API_VERSION <= 4:
                         partition, error_code, offset, ts = partition_info
+                    elif 5 <= response.API_VERSION <= 7:
+                        partition, error_code, offset, ts, log_start_offset = partition_info
+                    else:
+                        partition, error_code, offset, ts, log_start_offset, _, error_message = partition_info
                     tp = TopicPartition(topic, partition)
-                    error = Errors.for_code(error_code)
+                    error = error_message or Errors.for_code(error_code)
```
jeffwidman: so for `API_VERSION` > 7, this will be the raw `partition_info`... why is that? What is an example `partition_info`? Should it instead be called `partition_error_code`? And in that case, should it be:

```
error_code = partition_error_code or response_error_code
error = Errors.for_code(error_code)
```
?

kafka/protocol/produce.py
```diff
@@ -61,7 +61,6 @@ class ProduceResponse_v4(Response):
     API_VERSION = 4
     SCHEMA = ProduceResponse_v3.SCHEMA
 
-
```
jeffwidman: why is this dropped? Typically there are two blank lines between class definitions...

kafka/protocol/produce.py
```diff
@@ -78,6 +77,50 @@ class ProduceResponse_v5(Response):
...
+    """
+    API_KEY = 0
+    API_VERSION = 7
+    SCHEMA = ProduceResponse_v6.SCHEMA
+
+
+class ProduceResponse_v8(Response):
+    """
+    V8 bumped up to add two new fields record_errors offset list and error_message
+    (See KIP-467)
```
jeffwidman: These are much more helpful messages, thank you

requirements-dev.txt
```diff
@@ -15,3 +15,4 @@ pytest-mock==1.10.0
 sphinx-rtd-theme==0.2.4
 crc32c==1.7
 py==1.8.0
+zstandard==0.13.0
```
jeffwidman: This shouldn't be part of this PR since this PR doesn't add `zstd` support yet...

kafka/producer/sender.py
gabriel-tincu: @jeffwidman From http://kafka.apache.org/protocol.html#api_versions, `error_message` is 
```
The global error message summarizing the common root cause of the records that caused the batch to be dropped
```
So from my understanding, the broker will summarize the all the error messages into one, and provide them for you. The actual distinct messages plus codes are in the penultimate dropped / ignored unpack value, just because i found no need for it once the actual cumulated error messages were available all into one string

kafka/protocol/produce.py
gabriel-tincu: fixed

jeffwidman:
I'd like @dpkp and @tvoinarovskyi to take a look as well, particularly at the error handling bit.

kafka/producer/sender.py
tvoinarovskyi: @jeffwidman The structure seems to be as such:
```
Produce Response (Version: 8) => [responses] throttle_time_ms 
  responses => topic [partition_responses] 
    topic => STRING
    partition_responses => partition error_code base_offset log_append_time log_start_offset [record_errors] error_message 
      partition => INT32
      error_code => INT16
      base_offset => INT64
      log_append_time => INT64
      log_start_offset => INT64
      record_errors => batch_index batch_index_error_message 
        batch_index => INT32
        batch_index_error_message => NULLABLE_STRING
      error_message => NULLABLE_STRING
  throttle_time_ms => INT32
```
So the last part in partition_info is a UTF-8 string expressing the combined error that happened broker side.
But this PR will change the error passed to `_complete_batch `, which will result in callbacks called with an incorrect object. This is a regression and will also (!) break handling for code in `_complete_batch `, as it checks for exact equality to ``Errors.NoError`` there.

tvoinarovskyi: @gabriel-tincu Please pass the ``error_message`` explicitly using arguments up to the log in ``batch.done()``. There we can introduce it as part of the log statement, as you already did with ``log_start_offset``. I am not sure if we can safely change the error raised to the users (cc @dpkp ), but from how the code stands `Future.failure` has this magic line:
```python
    def failure(self, e):
        assert not self.is_done, 'Future is already complete'
        self.exception = e if type(e) is not type else e()        # <--------------
        assert isinstance(self.exception, BaseException), (
            'future failed without an exception')
        self.is_done = True
        self._call_backs('errback', self._errbacks, self.exception)
        return self
```
So we raise instances of exceptions to the users always. Thus if we change the
```python
self.produce_future.failure(exception)
```
line in ``batch.done()`` to something like:
```python
 if not error_message:
    exc = exception()
else:
    exc = exception(error_message)
self.produce_future.failure(exc)
```
We should be able to pass error message from broker to user without introducing any incompatible changes in code.

gabriel-tincu: @tvoinarovskyi ` Would keeping the old value be considered more stable then? I figured the cummulated strings would offer more information to the developer, but if one was to keep the old error implementation, there is no way to use that, since the api requires an error code as well, and there is no error code associated with the cumulated string

tvoinarovskyi:
@gabriel-tincu Great job on this, really appreciate your contribution. 
Pointed 1 minor point that needs addressing, as @jeffwidman had doubts about. Other parts lgtm!

kafka/producer/sender.py
jeffwidman: Thanks @tvoinarovskyi I was feeling a little queasy over the hand-wavy way this was dealing with errors, but wasn't sure how what a good way was. Thanks for digging in and stepping through it.

kafka/producer/sender.py
```diff
@@ -195,13 +195,18 @@ def _handle_produce_response(self, node_id, send_time, batches, response):
...
                     batch = batches_by_partition[tp]
                     self._complete_batch(batch, error, offset, ts)
```
tvoinarovskyi: I think you forgot to pass `log_start_offset` here

kafka/producer/sender.py
```diff
@@ -213,14 +219,16 @@ def _handle_produce_response(self, node_id, send_time, batches, response):
...
+    def _complete_batch(self, batch, error, base_offset, timestamp_ms=None, log_start_offset=None, global_error=None):
         """Complete or retry the given batch of records.
 
         Arguments:
             batch (RecordBatch): The record batch
             error (Exception): The error (or None if none)
             base_offset (int): The base offset assigned to the records if successful
             timestamp_ms (int, optional): The timestamp returned by the broker for this batch
+            log_start_offset (int): The start offset of the log at the time this produce response was created
+            global_error (Exception): The summarising error message
```
tvoinarovskyi: `global_error (str)`

tvoinarovskyi:
Nice! LGTM

tvoinarovskyi:
@jeffwidman Code looks good, but the current CI does not actually run 2.1 tests as API version checks are not up to date with the latest brokers. I will do a fix in separate PR, don't think it's a blocker.

kafka/producer/sender.py
gabriel-tincu: fixed

kafka/producer/sender.py
```diff
@@ -195,15 +195,21 @@ def _handle_produce_response(self, node_id, send_time, batches, response):
...
                     if response.API_VERSION < 2:
                         partition, error_code, offset = partition_info
                         ts = None
-                    else:
+                    elif 2 <= response.API_VERSION <= 4:
                         partition, error_code, offset, ts = partition_info
+                    elif 5 <= response.API_VERSION <= 7:
+                        partition, error_code, offset, ts, log_start_offset = partition_info
+                    else:
+                        partition, error_code, offset, ts, log_start_offset, _, global_error = partition_info
```
jeffwidman: It'd be helpful for future code readers if there is a code comment clarifying what the `_` is and why it's ignored.

jeffwidman:
Thanks!

PR #2038, opened 2020-04-21
Add logic for inferring newer broker versions

gabriel-tincu:
- New Fetch / ListOffsets request / response objects
- Add new test cases to inferr code based on mentioned objects
- Add unit test to check inferred version against whatever resides in KAFKA_VERSION
- Add new kafka broker versions to integration list
- Add more kafka broker versions to travis task list

As a side note, is there a ticket open or are there plans to use the same method the java client uses for keeping the Request/response classes up to date? It would seem much easier to import the folder with json api files and have our classes generated from that information, rather that doing a copy paste cycle multiple times, which can get very error prone

@jeffwidman

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2038)


.travis.yml
```diff
@@ -14,6 +14,10 @@ env:
     - KAFKA_VERSION=0.10.2.2
     - KAFKA_VERSION=0.11.0.3
     - KAFKA_VERSION=1.1.1
+    - KAFKA_VERSION=2.0.1
```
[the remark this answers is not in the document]
tvoinarovskyi: Let's do only 2.5.0 and 2.4.0 for now.

kafka/conn.py
```diff
@@ -1166,10 +1169,21 @@ def _infer_broker_version_from_api_versions(self, api_versions):
         # in reverse order. As soon as we find one that works, return it
         test_cases = [
             # format (<broker version>, <needed struct>)
+            ((2, 5, 0), DescribeAclsRequest[2]),
```
tvoinarovskyi, line 1172: Great job on those versions. I had a hard time finding them, did you copy it from somewhere?))

build_integration.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-: ${ALL_RELEASES:="0.8.2.2 0.9.0.1 0.10.1.1 0.10.2.2 0.11.0.3 1.0.2 1.1.1 2.0.1 2.1.1"}
+: ${ALL_RELEASES:="0.8.2.2 0.9.0.1 0.10.1.1 0.10.2.2 0.11.0.3 1.0.2 1.1.1 2.0.1 2.1.1 2.2.0 2.2.1 2.3.0 2.4.0 2.5.0"}
```
tvoinarovskyi, line 3: Let's stick to last minor versions, don't think there is a big difference between 2.2.0 and 2.2.1 in API...

tvoinarovskyi:
LGTM, but let's not add too much to Travis, it runs really long.
@jeffwidman Seems fine to me, could you have a look through, don't want to merge without someone else looking )

tvoinarovskyi:
@gabriel-tincu Great job!

jeffwidman:
Thanks for working on this, I don't have time today but will try to look at this over the weekend.

kafka/conn.py
gabriel-tincu: I ran the metadata request on each version of kafka, then plugged those in python and did an incremental diff. I was actually thinking of hardcoding those responses into the codebase as the actual api versions chosen would make more sense

.travis.yml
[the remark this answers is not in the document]
gabriel-tincu: fixed

build_integration.sh
gabriel-tincu: fixed

kafka/conn.py
```diff
@@ -1166,10 +1169,20 @@ def _infer_broker_version_from_api_versions(self, api_versions):
...
+            ((2, 3, 0), FetchRequest[11]),
+            ((2, 2, 0), OffsetRequest[5]),
+            ((2, 1, 0), FetchRequest[10]),
+            ((2, 0, 0), FetchRequest[8]),
+            ((1, 1, 0), FetchRequest[7]),
             ((1, 0, 0), MetadataRequest[5]),
             ((0, 11, 0), MetadataRequest[4]),
             ((0, 10, 2), OffsetFetchRequest[2]),
             ((0, 10, 1), MetadataRequest[2]),
+            # taken from https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_feature.c#L234
```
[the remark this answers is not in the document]
gabriel-tincu: fixed

kafka/conn.py
```diff
@@ -24,9 +24,12 @@
 from kafka.future import Future
 from kafka.metrics.stats import Avg, Count, Max, Rate
 from kafka.oauth.abstract import AbstractTokenProvider
-from kafka.protocol.admin import SaslHandShakeRequest
+from kafka.protocol.admin import SaslHandShakeRequest, DescribeAclsRequest_v2
```
gabriel-tincu, line 27: This plus my last commit is in response to the broker not recognizing the new V2 request even tho the schema is unchanged per 
https://github.com/apache/kafka/blob/2.5/clients/src/main/resources/common/message/DescribeAclsRequest.json
This leads me to believe that objects are serialized and deserialized differently when flexible versions are enabled , but i need some alone time with the java client and / or wireshark to figure out how that actually happens

tvoinarovskyi: @gabriel-tincu Ohhhh, so they changed the protocol specification in version 2.4.0 basically. All new versions after that would require us to have support for Tagged fields on protocol level. For now, we can leave the auto-discovery up to 2.5, that's not a problem, but the change in Admin client above should be reverted if version 2 does not work.

tvoinarovskyi: @gabriel-tincu As for the Compact versions of each type it should not be too hard, as we already have a varint implementation for length in protocol utils. It was used for message V2 parser.

gabriel-tincu: reverted the admin code

tvoinarovskyi:
@jeffwidman FYI. I will merge this PR with 1 minor point - it adds DescribeAclsRequest_v2 and DescribeAclsResponse_v2 that are not fully correct, as they don't abide to this new Protocol format, that supports Tagged fields. We need to add support to it before we can finish this class.

tvoinarovskyi:
@gabriel-tincu Thanks greatly for the contributions!

ofek:
Can this please be released? cc @dpkp

PR #2040, opened 2020-04-24
Feature delete consumergroups

swenzel:
I didn't find any issue requesting this, but I could use it so I just went ahead and implemented it. It's kind of a work in progress, just wanted to get it out for early feedback.
Tests are running fine and it looks good but what's still missing is testing if group coordinator discovery works properly when there are different brokers.
Also I'm not quite sure how to handle errors. Right now they are ignored and it's left for the caller to inspect the result but we could also just raise any errors.
WDYT?

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2040)


kafka/admin/client.py
```diff
@@ -341,6 +343,30 @@ def _find_coordinator_id(self, group_id):
         response = future.value
         return self._find_coordinator_id_process_response(response)
 
+    def _find_many_coordinator_ids(self, group_ids):
```
jeffwidman, line 346: Does Java have this method? When I wrote a bunch of this code two years ago I don't recall seeing anything that handled multiple group_ids at once, but I haven't been spelunking in their code for a while...

swenzel: I don't know, this one is mostly copy pasted from the method above. I went through https://kafka.apache.org/protocol but it doesn't look like there is an API that can provide the coordinator id for more than one group. Not even the DescribeGroups API returns the coordinators.

swenzel:
@jeffwidman I ran this on a cluster with 4 nodes and it deleted 3k groups without an issue... I suppose coordinator discovery works well.  
Considering how much work is required on the test suite to make the tests use multiple brokers I'd just leave it as it is now unless you really want me to extend the tests.  
Any opinion on how to handle failed deletes? Raise exception vs. leave it for the caller to check?

swenzel:
ping @dpkp @jeffwidman

kafka/admin/client.py
```diff
@@ -341,6 +343,30 @@ def _find_coordinator_id(self, group_id):
...
+        Sends a FindCoordinatorRequest message to the cluster for each group_id.
+        Will block until the FindCoordinatorResponse is received for all groups.
+        Any errors are immediately raised.
+
+        :param group_ids: A list of consumer group IDs. This is typically the group
+            name as a string.
+        :return: A list of tuples (group_id, node_id) where node_id is the id
+            of the broker that is the coordinator for the corresponding group.
+        """
+        # Note: Java may change how this is implemented in KAFKA-6791.
```
jeffwidman, line 358: This comment may be outdated, see https://github.com/apache/kafka/pull/4902#issuecomment-390830255

jeffwidman (approved):

jeffwidman:
> Considering how much work is required on the test suite to make the tests use multiple brokers I'd just leave it as it is now unless you really want me to extend the tests.

I am fine with leaving as-is... agreed that doing multiple brokers is a lot of extra work. It's probably better done as a `pytest` fixture that spins up multiple brokers and multiple consumer groups that are guaranteed to have their coordinators spread across the brokers, as we could leverage that for several unit tests...

jeffwidman:
> Any opinion on how to handle failed deletes? Raise exception vs. leave it for the caller to check?

That's really tough... in `list_consumer_groups()` and `describe_consumer_groups()` we immediately raise if _any_ group throws an error...  but those are non-destructive operations, so re-running them is fine. 

Here I'm hesitant to raise since everything is sent async so some groups may complete and others may fail... it is idempotent to simply re-run, but if some will always fail with errors like "group unknown/ doesn't exist" etc, then you want to know what happened to the others in the list... did they delete successfully? So I'd probably lean (slightly) toward just returning the error codes and letting the caller inspect to see which completed and which failed.

jeffwidman:
@swenzel can you rebase to fix conflicts? 

Take a look at my feedback above... the comment and newline are nits but if you're already rebasing be nice to fix these.

kafka/admin/client.py
```diff
@@ -1234,8 +1260,69 @@ def list_consumer_group_offsets(self, group_id, group_coordinator_id=None,
...
+            ]
+
+        self._wait_for_futures(futures)
+
+        results = []
+        for f in futures:
+            results.extend(self._convert_delete_groups_response(f.value))
+        return results
+
+    def _convert_delete_groups_response(self, response):
```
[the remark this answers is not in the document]
swenzel: Yeah, I also thought about this one a while... In the end my reasoning was something like this: the `convert` methods seem to have little to no logic and really do only conversions. The `process` methods involve more computation. So I took the former naming.

kafka/admin/client.py
swenzel: Removed both occurrences of that comment.

kafka/admin/client.py
swenzel: :+1:

test/test_admin_integration.py
```diff
@@ -138,3 +138,67 @@ def test_describe_configs_invalid_broker_id_raises(kafka_admin_client):
...
+    }
+
+    assert delete_results["test2-group1"] == NoError
+    assert delete_results["test2-group2"] == NonEmptyGroupError
+    assert delete_results["test2-group3"] == GroupIdNotFoundError
+
+    consumergroups = {group_id for group_id, _ in kafka_admin_client.list_consumer_groups()}
+    assert "test2-group1" not in consumergroups
+    assert "test2-group2" in consumergroups
+    assert "test2-group3" not in consumergroups
```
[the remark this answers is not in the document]
swenzel: Fixed. Btw. an editorconfig and/or pre-commit config could help with this :wink:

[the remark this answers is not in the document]
jeffwidman: Good call.. PR welcome: https://github.com/dpkp/kafka-python/issues/2125

jeffwidman:
Thanks @swenzel!

PR #2046, opened 2020-05-11
remove unused imports

jameslamb:
Thanks for this great project!

I wanted to take a few minutes and make a small contribution, so I ran `flake8` over this codebase. A lot of the warnings generated are things that should only be done optionally by maintainers (like formatting and whitespace), but I think I found a few non-controversial changes that would be helpful.

In this PR, I'd like to propose removing a few unused imports and one local variable that is created but never used in `ScramClient`.

Thanks for your time and consideration.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2046)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,7 +12,6 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys
 import os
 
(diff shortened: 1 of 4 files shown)


jameslamb:
I've rebased this to `master` to pick up changes since it was first opened.

PR #2048, opened 2020-05-12
Check client version iif not given

jacopofar:
The docstring states the `api_version` and `api_version_auto_timeout_ms` are used to decide if check_version() has to be called and for how long it has to wait, now the function follows that behavior, consistent with the `KafkaClient`

This could help with problems like dpkp#1308

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2048)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -204,11 +204,16 @@ class KafkaAdminClient(object):
         self._client = KafkaClient(metrics=self._metrics,
                                    metric_group_prefix='admin',
                                    **self.config)
-        self._client.check_version()
 
         # Get auto-discovered version from client if necessary
         if self.config['api_version'] is None:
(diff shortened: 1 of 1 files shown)


jacopofar:
I found later that this behavior was introduced in https://github.com/dpkp/kafka-python/pull/1953
because if `check_version()` is not called calls to `get_api_versions()` will fail.

Could it make sense to then remove the `api_version` parameter which is not used and pass the  `api_version_auto_timeout_ms` timeout to it?

jeffwidman:
Possibly related #2107 / #2108 ... as I noted https://github.com/dpkp/kafka-python/pull/2108#pullrequestreview-490178998, we should probably step-back and re-examine the way this gets handled so that it's consistent everywhere.

dpkp:
Addressed in #2496

Issue #1122, opened 2017-06-19
KIP-54: Sticky Partition Assignment Strategy


dpkp:
See: https://cwiki.apache.org/confluence/display/KAFKA/KIP-54+-+Sticky+Partition+Assignment+Strategy

lukekingbru:
Thanks so much, Let me dig into this one.

lukekingbru:
@dpkp, 
May I know how to run or debug kafka-python/test/test_assignors.py ?

dpkp:
if you have `tox` installed, you can run `tox -e py36 -- test/test_assignors.py` (replacing py36 with your local python version in development)

DieterDePaepe:
Any progress on this assignor?

PR #2057, opened 2020-05-26
KIP-54: Implement sticky partition assignment strategy

aynroot:
https://github.com/dpkp/kafka-python/issues/1122

Implementation of a sticky partition assignment strategy for kafka consumers.

General motivation behind sticky assignment:
> In certain circumstances the round robin assignor, [...], fails to produce an optimal and balanced assignment of topic partitions to consumers. [...] In addition, when a reassignment occurs, none of the existing strategies consider what topic partition assignments were before reassignment, as if they are about to perform a fresh assignment. Preserving the existing assignments could reduce some of the overheads of a reassignment. For example, Kafka consumers retain pre-fetched messages for partitions assigned to them before a reassignment. Therefore, preserving the partition assignment could save on the number of messages delivered to consumers after a reassignment. Another advantage would be reducing the need to cleanup local partition state between rebalances.

The assignment algorithm is described here: [KIP-54](https://cwiki.apache.org/confluence/display/KAFKA/KIP-54+-+Sticky+Partition+Assignment+Strategy)
An implementation of the strategy in Java: [AbstractStickyAssignor.java](https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractStickyAssignor.java) + [StickyAssignor.java](https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java). And its tests: [AbstractStickyAssignorTest.java](https://github.com/a0x8o/kafka/blob/de9a5e62076e107824d815c7a26303ee1c57e26f/clients/src/test/java/org/apache/kafka/clients/consumer/internals/AbstractStickyAssignorTest.java) + [StickyAssignorTest.java](https://github.com/a0x8o/kafka/blob/a2110b5aaa9f6cd082e9807124f0d6d3f38d1fcb/clients/src/test/java/org/apache/kafka/clients/consumer/StickyAssignorTest.java)

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2057)

### DIFF
--- a/kafka/coordinator/assignors/abstract.py
+++ b/kafka/coordinator/assignors/abstract.py
@@ -44,7 +44,7 @@ class AbstractPartitionAssignor(object):
         pass
 
     @abc.abstractmethod
-    def on_assignment(self, assignment, generation):
+    def on_assignment(self, assignment):
         """Callback that runs on each assignment.
 
         This method can be used to update internal state, if any, of the
@@ -52,6 +52,5 @@ class AbstractPartitionAssignor(object):
 
         Arguments:
             assignment (MemberAssignment): the member's assignment
-            generation (int): generation id
         """
         pass
--- a/kafka/coordinator/assignors/abstract.py
+++ b/kafka/coordinator/assignors/abstract.py
@@ -44,7 +44,7 @@ class AbstractPartitionAssignor(object):
         pass
 
     @abc.abstractmethod
-    def on_assignment(self, assignment):
+    def on_assignment(self, assignment, generation):
         """Callback that runs on each assignment.
 
         This method can be used to update internal state, if any, of the
@@ -52,5 +52,6 @@ class AbstractPartitionAssignor(object):
 
         Arguments:
             assignment (MemberAssignment): the member's assignment
+            generation (int): generation id
         """
         pass
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -234,7 +234,9 @@ class ConsumerCoordinator(BaseCoordinator):
 
         # give the assignor a chance to update internal state
         # based on the received assignment
-        assignor.on_assignment(assignment, generation)
+        assignor.on_assignment(assignment)
+        if assignor.name == 'sticky':
+            assignor.on_generation_assignment(generation)
 
         # reschedule the auto commit starting from now
         self.next_auto_commit_deadline = time.time() + self.auto_commit_interval
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -6,6 +6,7 @@ import functools
 import logging
 import time
 
+from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor
 from kafka.vendor import six
 
 from kafka.coordinator.base import BaseCoordinator, Generation
@@ -31,7 +32,7 @@ class ConsumerCoordinator(BaseCoordinator):
         'enable_auto_commit': True,
         'auto_commit_interval_ms': 5000,
         'default_offset_commit_callback': None,
-        'assignors': (RangePartitionAssignor, RoundRobinPartitionAssignor),
+        'assignors': (RangePartitionAssignor, RoundRobinPartitionAssignor, StickyPartitionAssignor),
         'session_timeout_ms': 10000,
         'heartbeat_interval_ms': 3000,
         'max_poll_interval_ms': 300000,
@@ -233,7 +234,7 @@ class ConsumerCoordinator(BaseCoordinator):
 
         # give the assignor a chance to update internal state
         # based on the received assignment
-        assignor.on_assignment(assignment)
+        assignor.on_assignment(assignment, generation)
 
         # reschedule the auto commit starting from now
         self.next_auto_commit_deadline = time.time() + self.auto_commit_interval
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -6,12 +6,12 @@ import functools
 import logging
 import time
 
-from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor
 from kafka.vendor import six
 
 from kafka.coordinator.base import BaseCoordinator, Generation
 from kafka.coordinator.assignors.range import RangePartitionAssignor
 from kafka.coordinator.assignors.roundrobin import RoundRobinPartitionAssignor
+from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor
 from kafka.coordinator.protocol import ConsumerProtocol
 import kafka.errors as Errors
 from kafka.future import Future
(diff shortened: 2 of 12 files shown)


aynroot:
CI checks are failing for Python 3.7 due to a pylint issue, looks like it is being addressed already here https://github.com/dpkp/kafka-python/pull/2058

tvoinarovskyi:
@aynroot Yea, you can ignore that CI error for now, I need to debug it properly

aynroot:
@tvoinarovskyi alright, thanks for the info
then this PR is ready for review

kafka/coordinator/assignors/abstract.py
```diff
@@ -44,13 +44,14 @@ def metadata(self, topics):
         pass
 
     @abc.abstractmethod
-    def on_assignment(self, assignment):
+    def on_assignment(self, assignment, generation):
```
dpkp, line 47: This is a breaking change for anyone who may have implemented their own custom assignor, as they will now need to add a generation parameter. Is there a way to structure this so that it doesn't require changes to existing assignors?

kafka/coordinator/consumer.py
```diff
@@ -6,6 +6,7 @@
 import logging
 import time
 
+from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor
```
dpkp, line 9: import order

kafka/coordinator/consumer.py
```diff
@@ -31,7 +32,7 @@ class ConsumerCoordinator(BaseCoordinator):
         'enable_auto_commit': True,
         'auto_commit_interval_ms': 5000,
         'default_offset_commit_callback': None,
-        'assignors': (RangePartitionAssignor, RoundRobinPartitionAssignor),
+        'assignors': (RangePartitionAssignor, RoundRobinPartitionAssignor, StickyPartitionAssignor),
```
dpkp, line 35: would rather not add this to the default list, especially since it requires an external dependency

setup.py
```diff
@@ -58,5 +59,8 @@ def run(cls):
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: Implementation :: PyPy",
         "Topic :: Software Development :: Libraries :: Python Modules",
+    ],
+    install_requires=[
+        "sortedcontainers==2.1.0"
```
dpkp, line 64: kafka-python has to-date not taken on any external dependencies, and we'd like to keep with that philosophy. Is this module necessary / can we refactor to use the standard library instead? If it is necessary, can we make it optional and disable the assignor if it is not installed?

kafka/coordinator/consumer.py
```diff
@@ -233,7 +234,7 @@ def _on_join_complete(self, generation, member_id, protocol,
 
         # give the assignor a chance to update internal state
         # based on the received assignment
-        assignor.on_assignment(assignment)
+        assignor.on_assignment(assignment, generation)
```
dpkp, line 237: this will also crash if the user has configured a custom assignor that doesn't take the additional parameter

dpkp:
Thanks -- I can see that you've put in a lot of work on this, and we'd love to get it merged. I've left a few comments re: compatibility and dependencies that I would like to resolve before adding to kafka-python.

setup.py
jeffwidman: It'd be really nice to see this land without an external dependency...

kafka/coordinator/consumer.py
jeffwidman: This assignment algorithm isn't just sticky, it's also better balanced than either of the other two... and it favors better balancing over stickiness... at least that's what I recall from when I helped drive that KIP forward.

So I think in that case, it'd be nice to add to the default list, and I'd even support moving it up the list to the primary default so that users get the most optimized balancing algorithm.

Be good to know what the Java client does here, I'd check but my wife just said dinner is ready so I need to roll...

jeffwidman:
Does this also include the updates described in [KIP-341](https://cwiki.apache.org/confluence/display/KAFKA/KIP-341%3A+Update+Sticky+Assignor%27s+User+Data+Protocol)?

If you ported the latest Java code, you should be fine, but if you based your implementation off the original KIP-54 it does have a bug that KIP-341 solves...

kafka/coordinator/consumer.py
jeffwidman: I looked into this a little more, and based on the work in [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol), it's probably not a great idea to set this as the preferred option... 

But I do still support including it in the list of default options. Additionally, if this is joining a mixed-language consumer group, it's more likely that it'd be cross-compatible with the Java clients etc if we include this in the default list.

kafka/coordinator/assignors/abstract.py
jeffwidman: I haven't looked deeply into [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol), but if it also leverages generations (and I suspect it might given that it includes incremental shifts), then we may want to consider keeping this and bumping the `kafka-python` version to 3.x for the breaking change... I have a feeling it may make our lives easier down the road. But if they don't need it there, then agree it would be nicer not to add it just for the sticky assignor case...

setup.py
aynroot: Yep, thought about it as well, should not be a rocket science to implement sorted sets by hand. Will do.

aynroot:
> Does this also include the updates described in [KIP-341](https://cwiki.apache.org/confluence/display/KAFKA/KIP-341%3A+Update+Sticky+Assignor%27s+User+Data+Protocol)? I'm sure Vahid also increased the test coverage to cover this case...
> 
> If you ported the latest Java code, you should be fine, but if you based your implementation off the original KIP-54 description, then we'd want to include the KIP-341 bugfix...

Yep, it includes KIP-341

kafka/coordinator/assignors/abstract.py
aynroot: Not sure about [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol), but generations here are crucial for [KIP-341](https://cwiki.apache.org/confluence/display/KAFKA/KIP-341%3A+Update+Sticky+Assignor%27s+User+Data+Protocol) bug fix.

We can allow backwards compatibility with something like this in `consumer.py`
```
assignor.on_assignment(assignment)
if assignor.name == 'sticky':
    assignor.on_generation_assignment(generation)
```

I'll make necessary changes.

aynroot:
@dpkp @jeffwidman I've addressed all the comments, this PR is ready to be re-reviewed.
Please take a look when you have time.

dpkp (approved):
Thanks, this is great and a huge new feature!! Fantastic work

sibiryakov:
@aynroot Well done Valeria!

Issue #2068, opened 2020-06-23
Use setuptools' "extras" feature for installing crc32c


mishas:
Setuptools has a feature called "extras", which allows easy install of "recommended" packages.
In this case, it'll be great to have an extra for installing crc32c, which will look something like: `pip install kafka-python[crc32c]`.

jeffwidman:
Fixed in #2069

PR #2069, opened 2020-06-23
Added crc32c extra requirement as per issue #2068

mishas:
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2069)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ setup(
     version=__version__,
 
     tests_require=test_require,
+    extras_require={"crc32c": ["crc32c"]},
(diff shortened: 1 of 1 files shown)


mishas:
It seems like specifying an issue number in the title doesn't link the PR with the issue.
So - here it is: Issue #2068

mishas:
Friendly ping?

jeffwidman (approved):
👍 

My only question about is about the name... as we're essentially stuck with this long-term due to backwards compatibility concerns... but I couldn't come up with anything better.

PR #2081, opened 2020-07-14
Update example.py

### LINKED ISSUE #1563: Example file (example.py) uses process instead of thread
(https://github.com/dpkp/kafka-python/blob/master/example.py)

Consumer is a subclass of multiprocessing.Process. 

The print statement in the run function will never print to the terminal as processes shares a different memory space.

Mostafa-Elmenbawy:
Use thread with the consumer instead of process.
Closes #1563

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2081)

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,15 +1,15 @@
 #!/usr/bin/env python
 import threading, logging, time
-import multiprocessing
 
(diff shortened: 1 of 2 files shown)


dpkp:
Thanks!

Issue #2078, opened 2020-07-13
Publishing to an ACL protected topic doesn't propagate TopicAuthorizationFailedError in KafkaProducer._wait_on_metadata


mkromer-tc:
In producer/kafka.py in `KafkaProducer._wait_on_metadata`, if an exception was identified in the future, it is not propagated.

The revised code should probably be something like:

```
            future = self._metadata.request_update()
            future.add_both(lambda e, *args: e.set(), metadata_event)
            self._sender.wakeup()
            metadata_event.wait(max_wait - elapsed)
            if future.exception:  # Propagate exception
                raise future.exception
```

dpkp:
Unclear why this isn't sufficient:
```
740             elif topic in self._metadata.unauthorized_topics:
741                 raise Errors.TopicAuthorizationFailedError(set([topic]))
```

PR #2082, opened 2020-07-14
Propagate future.exception in producer/kafka.py _wait_on_metadata

mkromer-tc:
Propagate any exceptions that happened during the future; otherwise _wait_on_metadata will time out even though the error is being continuously reflected in the metadata status fetched from the broker.  Issue #2078

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2082)


dpkp:
Wouldn't this prevent retrying exceptions that are likely to be resolved within the timeout? Are there specific exceptions that you are concerned about here?

mkromer-tc:
Any permission issues are not transient.  I was testing with locked down topics and even though the broker sent the error back the library ignored it.  Also, any attempt to connect with a misconfigured authentication scheme (ssl, sasl, whatever) will also never correct itself.

It’s been a bit since I looked at it, but propagating the exception solved at least two issues I encountered during testing.  Presumably the same issues could exist on the consumer side but I was only testing the producer side.  

I guess if you really felt like it you could split the exceptions into transient and permanent exceptions and only propagate the permanent ones, but it seems like that would be most of them.

dpkp:
It looks like the java client only raises on invalid or unauthorized topics, or on a connection authentication failure. Other failures are not raised and will get retried subject to the timeout.

PR #2101, opened 2020-08-07
docs: note flush behaviour of producer's .close()

hraban:
A little note about the flushing behaviour of KafkaProducer.close.

I don't actually know for a fact if this is true, I just assumed it from the code I read. If anyone can double check that would be a good idea :)

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -450,6 +450,9 @@ class KafkaProducer(object):
     def close(self, timeout=None):
         """Close this producer.
 
+        If a non-zero timeout is passed, this will also attempt to flush the buffer before
+        closing.
+
         Arguments:
             timeout (float, optional): timeout in seconds to wait for completion.
         """


kafka/producer/kafka.py
```diff
@@ -450,6 +450,9 @@ def __del__(self):
     def close(self, timeout=None):
         """Close this producer.
 
+        If a non-zero timeout is passed, this will also attempt to flush the buffer before
```
dpkp, line 453: I think it would be better to say:

Blocks until all internally buffered records are sent and any required acks are received, or until a specified timeout is reached.

hraban: No problems with changing the wording, although the code reads to my (unfamiliar) eye like a zero timeout would instantly skip the entire "graceful shutdown" block. A `0` and a `None` timeout are different: None fetches a default timeout (which could in turn be 0), and `0` just skips the entire graceful shutdown block and immediately closes. That seems slightly at odds with the proposed documentation?

Notably, my suggested documentation does not make the difference between 0 and None clear, either. This is more a bugreport masking as a PR, than a really well thought out proposal :) very open to further suggestions like yours.

PR #2107, opened 2020-08-22
Allow configurable timeouts in admin client check version

sunnyakaxd:
Currently there's no way to pass timeout to check_version if called from admin.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2107)


jeffwidman (approved):
Nice catch!

PR #2108, opened 2020-08-22
Always use config to set default value of timeout in check version

sunnyakaxd:
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2108)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -869,7 +869,7 @@ class KafkaClient(object):
         """
         return self._api_versions
 
-    def check_version(self, node_id=None, timeout=2, strict=False):
+    def check_version(self, node_id=None, timeout=None, strict=False):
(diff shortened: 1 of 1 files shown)


jeffwidman (commented):
I'm a little torn on this one... throughout the code there are a bunch of places we explicitly pass in the config value, and then here we fall back to the config as a default... if we go this route, then we should do it completely and _not_ explicitly pass it in elsewhere and just let them  fallback to this default...

Also, there's another default of `2` in `conn.check_version()` which should probably also default to this value as well for consistency. (although it doesn't look like the default value is ever really used as the only places its called have a passed-in value)

I'd like to hear input though from @dpkp / @tvoinarovskyi on whether they'd rather we get to this value by default or by explicitly passing it in everywhere

PR #2119, opened 2020-09-05
Fix initialization order in KafkaClient

### LINKED ISSUE #2118: Weird error after bad producer initialization
Hi, 
I'm getting this weird error after the GC tries to clean the memory and trigger the `__del__` method of `KafkaClient`. When the producer is not successfully initialized,  the property `_closed` is not ready in the `KafkaClient` object

this snippet if to illustrate the point:
```python
import os
import gc

def test_bad_kafka_initialization():
    connect_str = os.getenv('BROKER_URL')
    try:
        producer = KafkaProducer(bootstrap_servers=connect_str)
    except Exception as e:
        gc.collect()
        assert isinstance(e, TypeError)
        raise e
```
if the `BROKER_URL` env variable is not defined is pretty normal raise a `TypeError` exception but after that, another ignored exception shows up :

```
Exception ignored in: <function KafkaClient.__del__ at 0x10ecc04c0>
  File "..../python3.8/site-packages/kafka/client_async.py", line 443, in __del__
  File "..../python3.8/site-packages/kafka/client_async.py", line 417, in _close
AttributeError: 'KafkaClient' object has no attribute '_closed'
```
this happens because in `KafkaClient.__init__`  are properties defined after too after all initialization:
```        
 self.cluster = ClusterMetadata(**self.config)   -> here the initialization broke
 .....
 self._closed = False
 self._wake_r, self._wake_w = socket.socketpair()
 self._selector = self.config['selector']()

### PR DESCRIPTION
**Pull request by pecalleja**
this resolve #2118

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2119)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -200,11 +200,13 @@ class KafkaClient(object):
         for key in self.config:
             if key in configs:
                 self.config[key] = configs[key]
-
+        self._closed = False
+        self._wake_r, self._wake_w = socket.socketpair()
+        self._selector = self.config['selector']()
         self.cluster = ClusterMetadata(**self.config)
         self._topics = set()  # empty set will fetch all topic metadata
         self._metadata_refresh_in_progress = False
-        self._selector = self.config['selector']()
+
         self._conns = Dict()  # object to support weakrefs
         self._api_versions = None
         self._connecting = set()
@@ -212,7 +214,7 @@ class KafkaClient(object):
         self._refresh_on_disconnects = True
         self._last_bootstrap = 0
         self._bootstrap_fails = 0
-        self._wake_r, self._wake_w = socket.socketpair()
+
         self._wake_r.setblocking(False)
         self._wake_w.settimeout(self.config['wakeup_timeout_ms'] / 1000.0)
         self._wake_lock = threading.Lock()
@@ -226,7 +228,7 @@ class KafkaClient(object):
 
         self._selector.register(self._wake_r, selectors.EVENT_READ)
         self._idle_expiry_manager = IdleConnectionManager(self.config['connections_max_idle_ms'])
-        self._closed = False
+
         self._sensors = None
         if self.config['metrics']:
             self._sensors = KafkaClientMetrics(self.config['metrics'],
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -210,7 +210,6 @@ class KafkaClient(object):
         self.cluster = ClusterMetadata(**self.config)
         self._topics = set()  # empty set will fetch all topic metadata
         self._metadata_refresh_in_progress = False
-
         self._conns = Dict()  # object to support weakrefs
         self._api_versions = None
         self._connecting = set()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: kafka/client_async.py [hunk 1]
```diff
@@ -200,19 +200,21 @@ def __init__(self, **configs):
         for key in self.config:
             if key in configs:
                 self.config[key] = configs[key]
-
+        self._closed = False
+        self._wake_r, self._wake_w = socket.socketpair()
+        self._selector = self.config['selector']()
```
jeffwidman, line 205: It'd be good to have a comment here noting that the above properties should be top so as to not break when `__del__` is called.

jeffwidman (approved):
LGTM, although could you add a comment? It'd be all too easy for this problem to creep back in later from someone who doesn't realize the order matters for these specific attributes...

kafka/client_async.py
pecalleja: sure !!!  done.

kafka/client_async.py
```diff
@@ -201,18 +201,24 @@ def __init__(self, **configs):
...
+        # because they are used when __del__ method is called
+        self._closed = False
+        self._wake_r, self._wake_w = socket.socketpair()
+        self._selector = self.config['selector']()
+
         self.cluster = ClusterMetadata(**self.config)
         self._topics = set()  # empty set will fetch all topic metadata
         self._metadata_refresh_in_progress = False
-        self._selector = self.config['selector']()
+
```
jeffwidman, line 213: Wouldn't you want to delete these new lines?

kafka/client_async.py
```diff
@@ -226,7 +232,7 @@ def __init__(self, **configs):
 
         self._selector.register(self._wake_r, selectors.EVENT_READ)
         self._idle_expiry_manager = IdleConnectionManager(self.config['connections_max_idle_ms'])
-        self._closed = False
+
```
jeffwidman, line 235: Wouldn't you want to delete these new lines?

kafka/client_async.py
```diff
@@ -201,18 +201,24 @@ def __init__(self, **configs):
...
         self._conns = Dict()  # object to support weakrefs
         self._api_versions = None
         self._connecting = set()
         self._sending = set()
         self._refresh_on_disconnects = True
         self._last_bootstrap = 0
         self._bootstrap_fails = 0
-        self._wake_r, self._wake_w = socket.socketpair()
+
```
jeffwidman, line 221: Wouldn't you want to delete these new lines?

kafka/client_async.py
```diff
@@ -201,18 +201,24 @@ def __init__(self, **configs):
             if key in configs:
                 self.config[key] = configs[key]
 
+        # these properties need to be set on top of the initialization pipeline
+        # because they are used when __del__ method is called
+        self._closed = False
+        self._wake_r, self._wake_w = socket.socketpair()
+        self._selector = self.config['selector']()
+
```
jeffwidman, line 209: Keep this newline as part of keeping the "required initialization section" separate

kafka/client_async.py
pecalleja: 👍

kafka/client_async.py
pecalleja: 👍

kafka/client_async.py
pecalleja: 👍

kafka/client_async.py
pecalleja: 👍

jeffwidman (approved):

PR #2123, opened 2020-09-17
Add the optional compression libs to extras_require

### LINKED ISSUE #2122: Add python-snappy / lz4 / python-zstandard to the extras_require
As noted in https://kafka-python.readthedocs.io/en/master/install.html we have multiple optional dependencies for enabling various features. 

Those should be declared in `extras_require`.

jeffwidman:
As noted in https://kafka-python.readthedocs.io/en/master/install.html these are some additional optional libs that we should be specifying in `extras_require` so that callers can specify them if desired.

Fix #2122. 

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2123)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
         "crc32c": ["crc32c"],
         "lz4": ["lz4"],
         "snappy": ["python-snappy"],
-        "zstandard": ["python-zstandard"],
+        "zstd": ["python-zstandard"],
     },
     cmdclass={"test": Tox},
     packages=find_packages(exclude=['test']),
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,11 @@ setup(
     version=__version__,
 
     tests_require=test_require,
-    extras_require={"crc32c": ["crc32c"]},
+    extras_require={
+        "crc32c": ["crc32c"],
+        "lz4": ["lz4"],
+        "snappy": ["python-snappy"],
+    },
     cmdclass={"test": Tox},
     packages=find_packages(exclude=['test']),
     author="Dana Powers",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -35,7 +35,12 @@ def run(cls):
     version=__version__,
 
     tests_require=test_require,
-    extras_require={"crc32c": ["crc32c"]},
+    extras_require={
+        "crc32c": ["crc32c"],
+        "lz4": ["lz4"],
+        "snappy": ["python-snappy"],
+        "zstandard": ["python-zstandard"],
```
jeffwidman, line 42: perhaps this should be `zstd`

setup.py
```diff
@@ -35,7 +35,12 @@ def run(cls):
     version=__version__,
 
     tests_require=test_require,
-    extras_require={"crc32c": ["crc32c"]},
+    extras_require={
+        "crc32c": ["crc32c"],
+        "lz4": ["lz4"],
+        "snappy": ["python-snappy"],
```
jeffwidman, line 41: I wasn't sure if we should take the `"snappy"` namespace or use `"python-snappy"` so as to disambiguate between different implementations if we later have support for a pure-python vs c-based lib... this aplpies to the `"zstandard"` one as well...

jeffwidman: @dpkp @tvoinarovskyi any opinion?

dpkp: I think dropping the `python-` prefix is fine here.

dpkp (approved):

setup.py
```diff
@@ -35,7 +35,12 @@ def run(cls):
...
+        "zstd": ["python-zstandard"],
```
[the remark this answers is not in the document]
jeffwidman: You'd be better off opening a new issue, as comments on old PR's tend to get lost.

PR #2127, opened 2020-09-17
Merge _find_coordinator_id methods

### LINKED ISSUE #2124: Switch describe_consumer_groups() to use _find_many_coordinator_ids()
Once #2040 is merged, it'd be more performant/scalable if we updated `describe_consumer_groups()` to use the new `_find_many_coordinator_ids()` to issue the find coordinator requests async.

jeffwidman:
Previously there were two methods:
* `_find_coordinator_id()`
* `_find_many_coordinator_ids()`

But they do basically the same thing internally. And we need the plural
two places, but the singular only one place.

So merge them, and change the function signature to take a list of
`group_ids` and return a dict of `group_id: coordinator_id`'s.

As a result of this, the `describe_groups()` command should scale better
because the `_find_coordinator_ids()` command issues all the requests
async, instead of sequentially blocking as the `described_groups()` used
to do.

IIRC, this deviates slightly from the Java client, as they only take a singular group ID... mostly because it gets tricky to handle retriable vs non-retriable errors when you send a bunch at once. However, we don't handle errors quite the same way--we just raise if the futures don't complete, and also raise if we encounter any problems in the returned `FindCoordinatorResponse`. Given that, IMO this is more useful/scalable... 

Fix #2124

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2127)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -328,24 +328,8 @@ class KafkaAdminClient(object):
                 .format(response.API_VERSION))
         return response.coordinator_id
 
-    def _find_coordinator_id(self, group_id):
-        """Find the broker node_id of the coordinator of the given group.
-
-        Sends a FindCoordinatorRequest message to the cluster. Will block until
-        the FindCoordinatorResponse is received. Any errors are immediately
-        raised.
-
-        :param group_id: The consumer group ID. This is typically the group
-            name as a string.
-        :return: The node_id of the broker that is the coordinator.
-        """
-        future = self._find_coordinator_id_send_request(group_id)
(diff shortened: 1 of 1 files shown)


jeffwidman:
IIRC, this deviates slightly from the Java client, as they only take a singular group ID... mostly because it gets tricky to handle retriable vs non-retriable errors when you send a bunch at once. However, we don't handle errors quite the same way--we just raise if the futures don't complete, and also raise if we encounter any problems in the returned `FindCoordinatorResponse`. Given that, IMO this is more useful/scalable...

PR #2138, opened 2020-10-10
add .editorconfig file

### LINKED ISSUE #2125: Add a .editorconfig file
This is an easy one if someone's looking for a way to get involved with open source: https://github.com/dpkp/kafka-python/pull/2040#discussion_r490231719

add an `.editorconfig` file to specify newlines at end of file, spaces rather than tabs, etc.

lixiaoqity:
Fix #2125 

This file is used to keep all the files consistent. It is convenient for other people to contribute to the same project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false


.editorconfig
```diff
@@ -0,0 +1,12 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
```
jeffwidman, line 6: python should have 4 spaces for indents... it should probably be specified in a specific subsection so that our JSON files (CI config, etc) are still 2 space indents.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
jeffwidman, line 12: looks like this file is lacking the `insert_final_newline`? 😀

jeffwidman (commented):
Thanks for putting this together... Did you look around at other python-based projects? IMO this is missing some specs...

.editorconfig
lixiaoqity: Yes, python files usually have 4 spaces. I have fixed it and add the final newline style for the .md file. Thanks.

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.py]
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
+insert_final_newline = false
```
jeffwidman, line 16: Why are these both false for markdown? Why not make them true (or more concisely, simply drop the markdown specific section since it is identical to the global config)

jeffwidman:
Thanks for updating! Can you clarify why the markdown section has these configs? The other stuff looks good...

.editorconfig
lixiaoqity: These are only common writing formats for some special files. I think it is fine not to add blank line or remove any whitespace characters preceding newline characters in Markdown file. You can set what you want. Just keep all the files consistent. Thanks.

PR #2144, opened 2020-10-20
client: allow injection of custom kafka clients

gabriel-tincu:
Provide the consumer, producer and admin client with the option to
create the kafka client from a custom callable, thus allowing more
flexibility in handling certain low level errors

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2144)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -146,6 +146,7 @@ class KafkaAdminClient(object):
             sasl mechanism handshake. Default: one of bootstrap servers
         sasl_oauth_token_provider (AbstractTokenProvider): OAuthBearer token provider
             instance. (See kafka.oauth.abstract). Default: None
+        kafka_client (callable): Custom class / callable for creating KafkaClient instances
 
     """
     DEFAULT_CONFIG = {
@@ -186,6 +187,7 @@ class KafkaAdminClient(object):
         'metric_reporters': [],
         'metrics_num_samples': 2,
         'metrics_sample_window_ms': 30000,
+        'kafka_client': KafkaClient,
     }
 
     def __init__(self, **configs):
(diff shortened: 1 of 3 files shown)


jeffwidman:
Related: #735

gabriel-tincu:
> Related: #735

did not realize there was another PR dealing with the same issue. Should i just close this ? Will the other one get merged?

jeffwidman:
I think this should be kept open, at least until one of them is merged... I think both myself and @dpkp are in favor of allowing injection of custom clients, it was just never pushed across the line.

kafka/admin/client.py
```diff
@@ -186,6 +188,7 @@ class KafkaAdminClient(object):
         'metric_reporters': [],
         'metrics_num_samples': 2,
         'metrics_sample_window_ms': 30000,
+        'client_factory': None,
```
dpkp, line 191: Could we simplify by making this default to `kafka.client_async.KafkaClient` ?

I also think we could probably just call this `'client'` . Note that we don't use 'factory' to describe similar configs like 'selector' and 'metrics_reporters'

kafka/admin/client.py
```diff
@@ -205,9 +208,11 @@ def __init__(self, **configs):
         reporters = [reporter() for reporter in self.config['metric_reporters']]
         self._metrics = Metrics(metric_config, reporters)
 
-        self._client = KafkaClient(metrics=self._metrics,
-                                   metric_group_prefix='admin',
-                                   **self.config)
+        self._client = get_client_factory(self.config)(
```
dpkp, line 211: I don't love the `get_client_factory` indirection. Could we skip that and just use the config lookup directly?

kafka/admin/client.py
gabriel-tincu: updated

kafka/admin/client.py
gabriel-tincu: removed

gabriel-tincu:
Did away with the factory function, and opted for the `kafka_client` name, as simple `client` caused a conflict in one of the constructors downstream

gabriel-tincu:
@dpkp this is ready for another look, whenever you have the time

dpkp:
Looks great -- thanks!!

PR #2151, opened 2020-10-27
KIP-482: add support for flexible versions / tagged fields

gabriel-tincu:
- Add support for varints
- Add support for compact collections (byte array, string, array)
- Add support for new request and response headers, supporting flexible
versions
- Add list of supported flexible versions, plus extraction script
- Add List / Alter partition reassignments apis 

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2151)

### DIFF
--- a/kafka/protocol/api.py
+++ b/kafka/protocol/api.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 import abc
 
 from kafka.protocol.struct import Struct
-from kafka.protocol.types import Int16, Int32, String, Schema, Array
+from kafka.protocol.types import Int16, Int32, String, Schema, Array, TaggedFields
 
 
 class RequestHeader(Struct):
@@ -20,8 +20,39 @@ class RequestHeader(Struct):
         )
 
 
+class RequestHeaderV2(Struct):
+    # Flexible response / request headers end in field buffer
+    SCHEMA = Schema(
+        ('api_key', Int16),
+        ('api_version', Int16),
+        ('correlation_id', Int32),
+        ('client_id', String('utf-8')),
+        ('tags', TaggedFields),
+    )
+
+    def __init__(self, request, correlation_id=0, client_id='kafka-python', tags=None):
+        super(RequestHeaderV2, self).__init__(
+            request.API_KEY, request.API_VERSION, correlation_id, client_id, tags or {}
+        )
+
+
+class ResponseHeader(Struct):
+    SCHEMA = Schema(
+        ('correlation_id', Int32),
+    )
+
+
+class ResponseHeaderV2(Struct):
+    SCHEMA = Schema(
+        ('correlation_id', Int32),
+        ('tags', TaggedFields),
+    )
+
+
 class Request(Struct):
     __metaclass__ = abc.ABCMeta
+
+    FLEXIBLE_VERSION = False
 
     @abc.abstractproperty
     def API_KEY(self):
@@ -49,6 +80,16 @@ class Request(Struct):
 
     def to_object(self):
         return _to_object(self.SCHEMA, self)
+
+    def build_request_header(self, correlation_id, client_id):
+        if self.FLEXIBLE_VERSION:
+            return RequestHeaderV2(self, correlation_id=correlation_id, client_id=client_id)
+        return RequestHeader(self, correlation_id=correlation_id, client_id=client_id)
+
+    def parse_response_header(self, read_buffer):
+        if self.FLEXIBLE_VERSION:
+            return ResponseHeaderV2.decode(read_buffer)
+        return ResponseHeader.decode(read_buffer)
 
 
 class Response(Struct):
--- a/kafka/protocol/parser.py
+++ b/kafka/protocol/parser.py
@@ -4,10 +4,9 @@ import collections
 import logging
 
 import kafka.errors as Errors
-from kafka.protocol.api import RequestHeader
 from kafka.protocol.commit import GroupCoordinatorResponse
 from kafka.protocol.frame import KafkaBytes
-from kafka.protocol.types import Int32
+from kafka.protocol.types import Int32, TaggedFields
 from kafka.version import __version__
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


gabriel-tincu:
The way the request / response headers are handled needs an update, as apparently it's possible to send requests that have an api version with flexible versions enabled with both flexible versions and without. So if someone ends up defining a new requests object but using the old collections, introducing the empty tag buffer will mess up the entire payload. The java code base uses the json files to generate the classes so maybe we should look into that as well?

jeffwidman:
Is there a KIP associated with this change?

gabriel-tincu:
> Is there a KIP associated with this change? I switched teams and only work infrequently with Kafka now, so haven't kept up with the latest news.

@jeffwidman 
New types as described by the protocol docs: https://kafka.apache.org/protocol.html#protocol_types
Tagged fields KIP: https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields
Jira issue requiring version bump to use both tagged fields and compact data types for apis: https://issues.apache.org/jira/browse/KAFKA-9420

kafka/admin/client.py
```diff
@@ -460,6 +460,28 @@ def create_topics(self, new_topics, timeout_ms=None, validate_only=False):
         # TODO raise exceptions if errors
         return self._send_request_to_controller(request)
 
+    def list_partition_reassignments(self, topics_partitions, timeout_ms=None):
```
dpkp, line 463: Given the complexity, I might suggest splitting this into a PR that adds support for KIP-482 at the protocol level and another PR that adds the new admin APIs.

kafka/protocol/api.py
```diff
@@ -20,6 +20,22 @@ def __init__(self, request, correlation_id=0, client_id='kafka-python'):
...
 
 
+class RequestHeaderV2(Struct):
+    # Flexible response / request headers end in field buffer
+    SCHEMA = Schema(
+        ('api_key', Int16),
+        ('api_version', Int16),
+        ('correlation_id', Int32),
+        ('client_id', String('utf-8')),
+        ('_tag_buffer', TaggedFields)
```
dpkp, line 30: Can we name this simply `tags` ?

kafka/protocol/parser.py
```diff
@@ -181,3 +190,11 @@ def _reset_buffer(self):
         self._receiving = False
         self._header.seek(0)
         self._rbuffer = None
+
+    @staticmethod
+    def is_flexible_version_request(request):
```
dpkp, line 195: I think we should move this logic to the Request class as it seems like it's primary function is to determine which version of request and response headers we should be using for a particular api+version.

kafka/protocol/types.py
```diff
@@ -196,3 +196,156 @@ def repr(self, list_of_items):
...
+class TaggedFields(AbstractType):
+    @classmethod
+    def decode(cls, data):
+        num_fields = UnsignedVarInt32.decode(data)
+        ret = {}
+        if not num_fields:
+            return ret
+        prev_tag = -1
+        for i in range(num_fields):
+            tag = UnsignedVarInt32.decode(data)
```
dpkp, line 296: Is it expected that tag numbers are globally unique? Is the apache kafka project planning to publish a list of supported / known tags + numbers, and value serialization formats?

dpkp (commented):
> The way the request / response headers are handled needs an update, as apparently it's possible to send requests that have an api version with flexible versions enabled with both flexible versions and without.

Can you explain this in more detail? Do you mean that it is possible to send a "flexible version" api request but not include any tags in the header (as I understand it this takes up 1 byte)? Or do you mean not even include the tag structure in the header (0 bytes)? I would assume that any "flexible version" request would require at least a 1-byte empty tag buffer and would generate a response that also has at least a 1-byte empty tag buffer?

kafka/protocol/api.py
```diff
@@ -20,6 +20,22 @@ def __init__(self, request, correlation_id=0, client_id='kafka-python'):
...
+    )
+
+    def __init__(self, request, correlation_id=0, client_id='kafka-python'):
```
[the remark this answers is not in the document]
gabriel-tincu: I'm adding constructor support for the tags, but for now they will still be initialized to empty i guess

kafka/protocol/parser.py
```diff
@@ -59,9 +60,14 @@ def send_request(self, request, correlation_id=None):
         log.debug('Sending request %s', request)
         if correlation_id is None:
             correlation_id = self._next_correlation_id()
-        header = RequestHeader(request,
-                               correlation_id=correlation_id,
-                               client_id=self._client_id)
+        if not self.is_flexible_version_request(request):
```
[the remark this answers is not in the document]
gabriel-tincu: Moved this and the response header parsing to the request class

kafka/protocol/parser.py
```diff
@@ -162,6 +168,9 @@ def _process_response(self, read_buffer):
                 'Correlation IDs do not match: sent %d, recv %d'
                 % (correlation_id, recv_correlation_id))
 
+        # Flexible response / request headers end in field buffer
+        if self.is_flexible_version_request(request):
```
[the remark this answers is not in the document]
gabriel-tincu: Just took your hint and added a class property to indicate this

kafka/protocol/parser.py
```diff
@@ -162,6 +168,9 @@ def _process_response(self, read_buffer):
...
+            _ = TaggedFields.decode(read_buffer)
```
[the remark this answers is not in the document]
gabriel-tincu: Added ResponseHeader classes

kafka/admin/client.py
gabriel-tincu: removed, will add in another PR

kafka/protocol/api.py
gabriel-tincu: renamed

kafka/protocol/parser.py
gabriel-tincu: done

kafka/protocol/types.py
gabriel-tincu: The documentation leads me to believe that, at least initially, there will not be any structure to these tags, other than what the client implementation may wish to impart on them (their example of using a trace id comes to mind)

gabriel-tincu:
@dpkp this is ready for another look, whenever you have the time

dpkp:
Thanks a lot. I really like this design. We will probably have to convert the default for FLEXIBLE_VERSIONS from false -> true eventually, but I think we can wait for this feature to bake upstream a bit before we do that.

tvoinarovskyi:
Wow, this is awesome work! Good job! 👍

PR #2152, opened 2020-10-30
consumer: Exit poll if consumer is closed

rikonen:
Caller may invoke poll with long timeout and then end up closing the
consumer from another thread e.g. to stop the application. Previously
poll forcibly waited for the timeout before existing even though it
could do nothing but spin in busy loop for the remainder of the time.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2152)

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -651,7 +651,7 @@ class KafkaConsumer(six.Iterator):
         # Poll for new data until the timeout expires
         start = time.time()
         remaining = timeout_ms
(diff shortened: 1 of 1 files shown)


gabriel-tincu (approved):

jeffwidman (approved):
This makes sense to me, but I'm going to let @dpkp take a look as he knows this section of the code far better than me.

dpkp:
This makes sense to me -- thanks!

Issue #2153, opened 2020-10-30
Can't add a consumer to an existing group with a StickyPartitionAssignor


joein:
It is not possible to add a new consumer to an existing group with a `StickyPartitionAssignor` due to the lack of `__len__` method in `dict_itemiterator`.

Kafka version 1.0.1 (https://archive.apache.org/dist/kafka/1.0.1/kafka_2.11-1.0.1.tgz)
Python 3.8
kafka-python 2.0.2

Suppose we have `first_consumer.py`:
```python
from kafka import KafkaConsumer
from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor 

addr = "127.0.0.1:9092"
group_id = "some_group_id"
topic = ["first_topic"]
consumer = KafkaConsumer( *topic,  
                         partition_assignment_strategy=(StickyPartitionAssignor,), 
                         bootstrap_servers=addr,   
                         group_id=group_id)
while True:
    consumer.poll(timeout_ms=4000, max_records=1)
    print(consumer.assignment())
```
And `second_consumer.py`
```python
from kafka import KafkaConsumer
from kafka.coordinator.assignors.sticky.sticky_assignor import StickyPartitionAssignor 

addr = "127.0.0.1:9092"
group_id = "some_group_id"
topic = ["second_topic"]
consumer = KafkaConsumer( *topic,  
                         partition_assignment_strategy=(StickyPartitionAssignor,), 
                         bootstrap_servers=addr,   
                         group_id=group_id)
while True:
    consumer.poll(timeout_ms=4000, max_records=1)
    print(consumer.assignment())
```

Then we launch our script `first_consumer.py` and wait him to get his assignment and then launch the second script `second_consumer.py`. 
This sequence leads to  `join_group_request` during the second consumer connection, which in turn calls `assignor.metadata()`. Eventually, in `assignor.metadata` we are trying to encode `user_data`:
 ```python
return b''.join(
            [Int32.encode(len(items))] +
            [self.array_of.encode(item) for item in items]
        )
```
But `items` is type of `dict_iteritem` (which is six.iteritems) and has no `__len__` defined. And this is the reason for the following exception:

```python
File "first_consumer.py", line 12, in consume
    records = consumer.poll(timeout_ms=4000, max_records=1)
  File "/venv/lib/python3.8/site-packages/kafka/consumer/group.py", line 655, in poll
    records = self._poll_once(remaining, max_records, update_offsets=update_offsets)
[...]

PR #2154, opened 2020-10-30
Use six.viewitems instead of six.iteritems to avoid encoding problems

aynroot:
Should solve https://github.com/dpkp/kafka-python/issues/2153

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2154)


jeffwidman:
I'm going to go ahead and merge so `master` is fixed, but do you want to add a test for the underlying issue? IIRC, you had great test coverage originally, so I'm surprised this wasn't caught earlier...

aynroot:
yes, I'll take care of it this week, wanted to unlock the issue as fast as possible last week without having much time

the problem I think is that the `metadata` method is _the_ one currently which is not covered and is almost duplicated in the tests with a helper method :D

PR #2169, opened 2020-11-16
Support connections through SOCKS5 proxies

hnousiainen:
Implement support for SOCKS5 proxies. Implement a new proxy wrapper that handles SOCKS5 connection, authentication and requesting connections to the actual Kafka broker endpoints.

The proxy can be configured via a new keyword argument `socks5_proxy` to consumers, producers or admin client. The value is URL with optional username and password. E.g. `socks5://user:secret@proxy.example.com:10800`

The implementation is done in state machine that makes progress on repeated calls to connect_ex. The rationale with this bit strange design is to minimize amount of changes on the actual BrokerConnection object.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2169)


markbrock301:
Is there any reason this hasn't been merged?  It would be a handy feature.

PR #2172, opened 2020-11-25
feat: logged cancelled exception as info messages

Arfey:
we manual interrupt our connection so all inflight request which was canceled by this reason shouldn't log it as an error

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2172)


dpkp:
Merged in #2467

Arfey:
thx so much

PR #2185, opened 2020-12-23
using context manager to acquire/release lock

hackaugusto:
This change makes the code a tiny little bit easier to read, and maybe more robust.

There are some exceptions that may be raised which are not handled, and would result in the lock not being release. For example, `self._maybe_connect(try_node)` calls `BrokerConnection.connect` which may raise a `KafkaConnectionError` on DNS errors. In that case the lock would not be released.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2185)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -888,42 +888,37 @@ class KafkaClient(object):
             UnrecognizedBrokerVersion: please file bug if seen!
             AssertionError (if strict=True): please file bug if seen!
         """
-        self._lock.acquire()
-        end = time.time() + timeout
-        while time.time() < end:
+        with self._lock:
+            end = time.time() + timeout
+            while time.time() < end:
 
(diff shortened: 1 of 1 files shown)


kafka/client_async.py
```diff
@@ -888,42 +888,37 @@ def check_version(self, node_id=None, timeout=2, strict=False):
...
-            except Errors.NodeNotReadyError:
-                # Only raise to user if this is a node-specific request
-                if node_id is not None:
-                    self._lock.release()
-                    raise
-            finally:
-                self._refresh_on_disconnects = True
-
-        # Timeout
-        else:
```
hackaugusto, line 925: This is the only other change in the code. There is no `break` in the loop, so the else clause is always executed, which is the same as having the code directly after the loop. Removing the `else` is a tiny little simpler and makes `pylint` happier

Issue #2188, opened 2021-01-05
KafkaClient.maybe_connect and ClusterMetadata.update_metadata are not synchronized which leads to broken asserts


hackaugusto:
I hit this problem in production:

step | **thread1** | **thread2**
--- | --- | ---
1 | sends MetadataRequest_v1 |
2 | calls poll() |
3 |  | calls `maybe_connect` with node id `n`
4 | process MetadataResponse_v1 which removes `n` from the metadata |

When step 3 happens, the id `n` is added to `_connecting`. Once step 4 happens, the corresponding metadata for node `n` is removed from `ClusterMetadata`. The result is that all subsequent calls to `poll` will fails on this assertion:

https://github.com/dpkp/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/client_async.py#L372

Here is a somewhat hacked together script to reproduce the problem:

```python
#!/bin/env python
from contextlib import contextmanager
import subprocess
import threading
import logging
from time import sleep
from kafka.client_async import KafkaClient
from kafka.admin.client import KafkaAdminClient
from kafka.admin.new_topic import NewTopic

PORT_START = 2000
PORT_END = 2020
ZK_PORT = 2181


def create_pod():
    args = [
        "podman",
        "pod",
        "create",
        "--publish",
        f"{PORT_START}-{PORT_END}:{PORT_START}-{PORT_END}",
        "--name",
        "kafka",
    ]
    subprocess.run(args)
    print("kafka pod created")


def start_zk():
    args = [
        "podman",
        "run",
        "--detach",
        "--tty",
        "--rm",
        "--pod",
        "kafka",
        "--name",
        "zk",
        "--env",
        f"ZOOKEEPER_CLIENT_PORT={ZK_PORT}",
        "confluentinc/cp-zookeeper",
    ]
    subprocess.run(args)
    print("zk started")


def start_kafka(num: int):
    port = PORT_START + num * 2
    port_host = port + 1

    args = [
        "podman",
        "run",
        "--detach",
        "--tty",
        "--rm",
        "--pod",
        "kafka",
        "--name",
        f"kafka-{num}",
        "--env",
        "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT",
        "--env",
        f"KAFKA_ZOOKEEPER_CONNECT=localhost:{ZK_PORT}",
        "--env",
[...]

PR #2189, opened 2021-01-05
bugfix: race among _connecting and cluster metadata

hackaugusto:
Fixes https://github.com/dpkp/kafka-python/issues/2188

A call to `maybe_connect` can be performed while the cluster metadata is
being updated. If that happens, the assumption that every entry in
`_connecting` has metadata won't hold. The existing assert will then
raise on every subsequent call to `poll` driving the client instance
unusable.

This fixes the issue by ignoring connetion request to nodes that do not
have the metadata available anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2189)

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -368,18 +368,26 @@ class KafkaClient(object):
             conn = self._conns.get(node_id)
 
             if conn is None:
-                broker = self.cluster.broker_metadata(node_id)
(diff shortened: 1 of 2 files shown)


hackaugusto:
I'm not 100% sure if the fix is good enough, there are multiple paths that lead to calls to `maybe_connect`, and I'm not sure if they can recover from the node disappearing.

PR #2202, opened 2021-02-08
bugfix: don't allow None in String, Bytes

ivanyu:
This commit makes `String`, `Bytes` method `encode` not allow `None` as input. This aligns it with how Kafka Java library handles this (https://github.com/apache/kafka/blob/57cef765f55fa1b9bab99cf4c71cf40f64533656/clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java#L406, https://github.com/apache/kafka/blob/57cef765f55fa1b9bab99cf4c71cf40f64533656/clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java#L636).

Please note that `Array` and the compact versions of these types are not altered, since in those cases nullability is valid.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2202)

### DIFF
--- a/kafka/protocol/__init__.py
+++ b/kafka/protocol/__init__.py
@@ -43,5 +43,7 @@ API_KEYS = {
     40: 'ExpireDelegationToken',
     41: 'DescribeDelegationToken',
     42: 'DeleteGroups',
+    45: 'AlterPartitionReassignments',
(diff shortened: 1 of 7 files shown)


ivanyu:
Recreated in https://github.com/dpkp/kafka-python/pull/2203

PR #2203, opened 2021-02-09
bugfix: don't allow encoding None in String, Bytes

ivanyu:
This commit makes `String`, `Bytes` method `encode` not allow `None` as input. This aligns it with how Kafka Java library handles this (https://github.com/apache/kafka/blob/57cef765f55fa1b9bab99cf4c71cf40f64533656/clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java#L406, https://github.com/apache/kafka/blob/57cef765f55fa1b9bab99cf4c71cf40f64533656/clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java#L636).

Please note that `Array` and the compact versions of these types are not altered, since in those cases nullability is valid.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2203)

### DIFF
--- a/kafka/protocol/types.py
+++ b/kafka/protocol/types.py
@@ -96,7 +96,7 @@ class String(AbstractType):
 
     def encode(self, value):
         if value is None:
-            return Int16.encode(-1)
+            raise ValueError("value cannot be None")
         value = str(value).encode(self.encoding)
         return Int16.encode(len(value)) + value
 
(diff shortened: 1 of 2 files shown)


ivanyu:
The problem: some strings are nullable and in Kafka code base differentiates them into `STRING` and `NULLABLE_STRING`. Currently, all strings are considered nullable by the library, even in places where the protocol excepts non-nullable. This PR makes all string non-nullable, but there are places where they should be nullable. I'm closing the PR, because the problem needs to be addressed from a different angle: by separating the types.

PR #2206, opened 2021-02-17
Fix DescribeConfigsResponse_v1 config_source vs is_default

ryarnyah:
When i tried to use `DescribeConfigsResponse_v1` i realised that it contain `is_default` field but Kafka protocol changed for this version to replace `is_default` with `config_source`.

This fix the response.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2206)


wbarnha (approved):
Double-checked https://kafka.apache.org/11/protocol.html, the change here is correct! Good catch, LGTM!

Also, according to the linked protocol definition, `config_names` should be `config_name`. Weird.

dpkp:
Landed as #2464

PR #2213, opened 2021-03-17
Flush before close

moshez:
If there are any messages pending, flush them before doing anything else. This makes it more likely that callbacks have a chance to run before the producer is closed, which can help in corner cases (like when callbacks want to send another message)

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2213)


dpkp:
Closing in favor of #2537

PR #2226, opened 2021-04-21
Fix producerconfigs documentation link

MandarJKulkarni:
Documentation link fix

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2226)


jeffwidman (approved):
Thanks!

I suspect we have the same problem in consumer docs too if you want to take a quick check.

PR #2230, opened 2021-05-12
Do not list or describe non-consumer groups

mborst:
When listing or describing groups via a `ListGroups` or `DescribeGroups`
request, Kafka also returns non-consumer groups, e.g. Kafka Connect
groups. Those should be omitted when listing or describing consumer
groups, just as in the official implementation [1].

[1] https://github.com/apache/kafka/blob/13ffebe2f1012da473189c4fe86fd14b83809962/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3428

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2230)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -1018,24 +1018,23 @@ class KafkaAdminClient(object):
                     described_groups_field_schema = response_field.array_of
                     described_group = response.__dict__[response_name][0]
                     described_group_information_list = []
-                    protocol_type_is_consumer = False
                     for (described_group_information, group_information_name, group_information_field) in zip(described_group, described_groups_field_schema.names, described_groups_field_schema.fields):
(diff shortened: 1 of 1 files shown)


mborst:
Wouldn't that be covered by [this](https://github.com/dpkp/kafka-python/pull/2230/files#diff-1f78bd289614d00208876a775765ea4cb5e2d319167ec8931423ff7a07e6968fR1025), and in the same way as in the [reference implementation](https://github.com/apache/kafka/blob/13ffebe2f1012da473189c4fe86fd14b83809962/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3430)?

mborst:
Is there anything I can do to get this merged?

PR #2231, opened 2021-05-13
Implemented the return type of the send function;

JGabrielGruber:
It would be nice to have an accessible and easy way to know the returns of the send functions, as the Documentation doesn't have a index for RecordMetadata;

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2231)

### DIFF
--- a/kafka/producer/kafka.py
+++ b/kafka/producer/kafka.py
@@ -15,7 +15,7 @@ from kafka.client_async import KafkaClient, selectors
 from kafka.codec import has_gzip, has_snappy, has_lz4, has_zstd
 from kafka.metrics import MetricConfig, Metrics
 from kafka.partitioner.default import DefaultPartitioner
-from kafka.producer.future import FutureRecordMetadata, FutureProduceResult
+from kafka.producer.future import FutureRecordMetadata, FutureProduceResult, RecordMetadata
 from kafka.producer.record_accumulator import AtomicInteger, RecordAccumulator
 from kafka.producer.sender import Sender
 from kafka.record.default_records import DefaultRecordBatchBuilder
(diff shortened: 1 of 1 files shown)


kafka/producer/kafka.py
```diff
@@ -538,7 +538,7 @@ def _estimate_size_in_bytes(self, key, value, headers=[]):
             return LegacyRecordBatchBuilder.estimate_size_in_bytes(
                 magic, self.config['compression_type'], key, value)
 
-    def send(self, topic, value=None, key=None, headers=None, partition=None, timestamp_ms=None):
+    def send(self, topic, value=None, key=None, headers=None, partition=None, timestamp_ms=None) -> RecordMetadata:
```
jeffwidman, line 541: This is just adding the type annotation, right? Does it do anything else?

I don't currently work in python much, so haven't kept up with all the latest language developments.

Congee: It is just type annotation and users need it badly :sweat_smile: 

related to this ticket https://github.com/dpkp/kafka-python/issues/2245

Congee: Since this library supports Python 2.7, we can use type comments to annotate stuff. It would be `# type: (...) -> RecordMetadata` to the next line of `def`.

See https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code

JGabrielGruber: Done, fixed it! Thank you for the feedback

JGabrielGruber: > This is just adding the type annotation, right? Does it do anything else?
> 
> I don't currently work in python much, so haven't kept up with all the latest language developments.

Yes @jeffwidman , it allows users to quickly trough the IDE's knows whats the return type of the function, pretty useful.
I'm sorry for the delay, and I've fixed it for compatibility with Python 2.7 as suggested by @Congee

dpkp:
The docstring pretty clearly shows the input and output types.

PR #2237, opened 2021-05-26
Support loading cas from memory.

abergmeier:
Passes along cadata to SSL library.
This removes the need to unnecessarily persist CA files when already
present in memory.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2237)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -107,6 +107,8 @@ class KafkaAdminClient(object):
         ssl_check_hostname (bool): Flag to configure whether SSL handshake
             should verify that the certificate matches the broker's hostname.
             Default: True.
+        ssl_cadata (str): optional ca filecontent to use in certificate
+            verification. default: None.
         ssl_cafile (str): Optional filename of CA file to use in certificate
             verification. Default: None.
         ssl_certfile (str): Optional filename of file in PEM format containing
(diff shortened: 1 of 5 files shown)


abergmeier:
Seems like something is wrong with CI?

kafka/admin/client.py
```diff
@@ -107,6 +107,8 @@ class KafkaAdminClient(object):
         ssl_check_hostname (bool): Flag to configure whether SSL handshake
             should verify that the certificate matches the broker's hostname.
             Default: True.
+        ssl_cadata (str): optional ca filecontent to use in certificate
+            verification. default: None.
```
jeffwidman, line 111: Can you capitalize properly?

```suggestion
        ssl_cadata (str): Optional CA filecontent to use in certificate
            verification. Default: None.
```

This applies several other places in the PR too, since this text is sorta copy/pasta.

jeffwidman (changes requested):
This looks good to me other than the docstring capitalization.

I don't have a way to test, but I'll take your word for it that the argument is supported by the underlying SSL lib and just needs to be plumbed through. Looks straightforward.

jeffwidman:
We need to migrate from Travis to some other system.

abergmeier:
We don't use Kafka anymore.

PR #2285, opened 2021-12-11
Fix base class of DescribeClientQuotasResponse_v0

ods:
Wrong base class causes error when checking code with mypy:
```
kafka/protocol/admin.py:928: error: Class kafka.protocol.admin.DescribeClientQuotasResponse_v0 has abstract attributes "RESPONSE_TYPE"
kafka/protocol/admin.py:928: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
```

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2285)


wbarnha (approved):

dpkp:
Rebased as #2465

PR #2296, opened 2022-01-18
Fix a link to snappy

kurtmckee:
The current link leads to HTTP 404.

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2296)

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -58,7 +58,7 @@ Optional Snappy install
 Install Development Libraries
 =============================
 
-Download and build Snappy from https://google.github.io/snappy/
+Download and build Snappy from https://github.com/google/snappy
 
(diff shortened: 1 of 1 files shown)


jeffwidman (dismissed):

docs/install.rst
```diff
@@ -58,7 +58,7 @@ Optional Snappy install
 Install Development Libraries
 =============================
 
-Download and build Snappy from https://google.github.io/snappy/
```
jeffwidman, line 61: This doesn't 404 for me, are you sure it's a bad link?

kurtmckee: Huh. It's live now. Thanks for double-checking!

PR #2300, opened 2022-01-27
Feat: FNV1a 32-Bit Partitioner

kemalty:
**Summary**

This PR introduces FNV1-a 32-bit hasher as a new partitioner. The tests confirmed that this implementation yields the same hash and partition selection as Goka/Sarama.

**Why**

Kafka leaves the partition selection up to the producer. This creates a potential discrepancy issue when different libraries are used. As different libraries across different programming languages may use different hashers, or different configurations for them, the messages may arrive at partitions different than expected. When different hashers or different configurations are used, the message with the same key may arrive at an unexpected partition.

This was the case for us at [Beat](https://thebeat.co/). We realized that messages that are produced by Python services using Kafka-python arrive at partitions that are different than what consumer services expect. We realized that this is because Kafka-python uses murmur2 as the default partitioner, while Goka library uses Sarama which uses FNV1a-32.

**What** 

This PR introduces a new partitioner that is based on FNV1a 32-bit. It is a separate and isolated implementation. Therefore users can select using the default one, or this one easily.

The partitioner is designed to match the partitioner of Goka/Sarama. It uses the "a" variant of FNV based on 32-bit. It also utilizes twos-complement decimal conversions, and the same parameters used for mentioned libraries. As a result, with this partitioner, python-kafka calculates the same hash and selects the same partition as Goka/Sarama does. In our tests, we experimentally proved that this implementation outputs the same hash and partitioner as Goka/Sarama does.

**Who**

I worked together on this change with [Evgenia Martynova](https://www.linkedin.com/in/evgenia-martynova-24b49315/) at Beat. Beat is a ride-hailing company that has engineering hubs in Athens and Amsterdam. We are always on the look for great engineers! [We are hiring!](https://thebeat.co/en/careers/?intl=1)

---
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2300)


kemalty:
Hey @dpkp - How shall we proceed with this? Any directions?

wbarnha (approved):
Let's give it a try. I'm not willing to make it the default partitioner, but I'm all for giving users the choice to try other partitioning algorithms.

wbarnha:
> > Let's give it a try. I'm not willing to make it the default partitioner, but I'm all for giving users the choice to try other partitioning algorithms.
> 
> Thanks @wbarnha - Could you please guide me to the next steps? I think someone needs to mark it as "reviewed" in the Reviewable tool.

I'll include it after the next release of kafka-python. I have a very specific release plan for 2.0.3/2.1.0. This will probably be included in 2.2.0/2.1.0.

dpkp:
kafka-python's default partitioner is intended to be consistent with the official java client. The interface allows providing a custom partitioner for special cases like this. But I'm not inclined to add this custom class to kafka-python (and maintain it). It seems like you could distribute this separately if needed. I'm confused why go/sarama decided to implement a different partitioning scheme.

PR #2321, opened 2022-08-05
Add zstd support on legacy record and ensure no variable is referred before definition 

ax-ale:
It solves situations where some protocol is not supported, and some variables would be referred before the assignment 
Also, it's possible to be necessary to use the zstandard with legacy records, so I copy the same lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2321)


dpkp:
zstd is not available for legacy records (message v0 and v1). And currently there is no known codec bit set that is not accounted for, so this is only an issue in theory. Thanks for putting up a PR, but I'm going to close rather than attempt to rebase / fixup / clean.

Issue #2286, opened 2021-12-15
"Timeout to send to wakeup socket!" when trying to describe consumer group


lvicentesanchez:
This is happening inside a docker container based on:

apache/airflow:2.2.2-python3.9

The version of python inside the container is 3.9.9. We are trying to run a datahub recipe that reads information about consumer groups and it runs just fine on my macbook pro m1 (latest model) using python 3.9.7.

I wanted to increase the wakeup timeout, but it's not possible to do that using the admin client as it's not one of the valid options.

This error happens after we get the list of all consumer groups. When we try to describe all of them, we get the timeout:

```
    354  def _send_request_to_node(self, node_id, request):
 (...)
    364      while not self._client.ready(node_id):
    365          # poll until the connection to broker is ready, otherwise send()
    366          # will fail with NodeNotReadyError
    367          self._client.poll()
--> 368      return self._client.send(node_id, request)
    ..................................................
     self = <kafka.admin.client.KafkaAdminClient object at 0x4015388310>
     node_id = 6
     request = GroupCoordinatorRequest_v0(consumer_group='some-group')
     self._client.ready = <method 'KafkaClient.ready' of <kafka.client_async.KafkaClient object at 0x4103564fa0> client_async.py:395>
     self._client.poll = <method 'KafkaClient.poll' of <kafka.client_async.KafkaClient object at 0x4103564fa0> client_async.py:550>
     self._client.send = <method 'KafkaClient.send' of <kafka.client_async.KafkaClient object at 0x4103564fa0> client_async.py:515>
    ..................................................

File "/home/airflow/.local/lib/python3.9/site-packages/kafka/client_async.py", line 546, in send
    515  def send(self, node_id, request, wakeup=True):
 (...)
    542      # Wakeup signal is useful in case another thread is
    543      # blocked waiting for incoming network traffic while holding
    544      # the client lock in poll().
    545      if wakeup:
--> 546          self.wakeup()
    547
    ..................................................
     self = <kafka.client_async.KafkaClient object at 0x4103564fa0>
     node_id = 6
     request = GroupCoordinatorRequest_v0(consumer_group='some-group')
     wakeup = True
     self.wakeup = <method 'KafkaClient.wakeup' of <kafka.client_async.KafkaClient object at 0x4103564fa0> client_async.py:929>
    ..................................................

File "/home/airflow/.local/lib/python3.9/site-packages/kafka/client_async.py", line 935, in wakeup
    929  def wakeup(self):
 (...)
[...]

Issue #2290, opened 2022-01-05
is kafka-python deprecated?


RJ3:
No release in 1.5 yrs. Just wondering if eventually changes on Kafka broker will become incompatible with what we have. Is the project dead?

sparklecast1:
Damn, It would be painful.

dpkp:
Im not doing new feature work currently. I am not aware of serious
compatibility issues, apart from new broker features that are not supported.

Bigger picture, there were 2 of us actively maintaining this project for
several years. Both have since moved to companies that do not use this
library. If there are others who are actively using and would like to join
the maintainer team, please reach out!

fenchu:
Others are not able to get wheels for 3.10 out (https://github.com/confluentinc/confluent-kafka-python), so I really appreciate You keep up the work. I'll try and contribute if I can. I have a pretty complex ssl rig that triggers lots some bugs.

akira-kurogane:
Please, be honest and declare if this project is deprecated.

I was building another project which uses the current committed code in the main branch. It's requirements install step pulled the latest released version 2.0.2, but then I was debugging 'Unrecognized configs: .... KafkaClient' messages for quite a while before I had the idea to check if the erroring code line, committed ~2 years ago, was in fact not part of the latest release.

Is another package recommend in place of this one?

grofte:
Aiven is recommending the use of this package: https://docs.aiven.io/docs/products/kafka/howto/connect-with-python.html

They should really put some resources into maintaining it.

hrichards:
@dpkp Let me know if you're still in need of maintainers (email in profile).  My company uses Kafka and Kafka-Python extensively and has need of some of the unreleased and in-PR functionality for this project, and so could surely come up with a maintainer or two.

Alternately, have you considered moving this project to Jazzband?  https://github.com/jazzband

jeffwidman:
@hrichards your best bet is probably to directly email @dpkp (you can pull his email off one of his old commits).

wbarnha:
I reached out to @dpkp, if I don't hear back, I'm probably going to start working on building workflows in https://github.com/wbarnha/kafka-python-ng and automating creating pull requests from the existing ones here.

wbarnha:
Got push access, hopefully I'll get to revive this project shortly!

wbarnha:
We're back!

wbarnha:
I'm still waiting on getting PyPi workflows ready, in case anyone revisits this.

[...]

PR #2335, opened 2022-11-04
Allow disabling thread wakeup in send_request_to_node

yzhan289:
The Datadog kafka_consumer check uses the `kafka-python` client library to collect metrics. On very large clusters, it sometimes gets a `KafkaTimeoutError` similar to the one [here](https://github.com/dpkp/kafka-python/issues/2286). We've been able to [work around this](https://github.com/DataDog/integrations-core/pull/13221) by disabling the socket wakeup since our check runs in a single thread (after looking at [this old Github issue](https://github.com/dpkp/kafka-python/pull/1761) that was meant to fix the issue). However, we needed to re-implement some parts of the `_send_request_to_node()` function since the library's version always runs a socket wakeup. 

This PR just adds the `wakeup` argument to `_send_request_to_node()` to make it easier to skip socket wakeup. 

I saw that there's a few other places in the code that calls `_client.send()` where adding a `wakeup` argument can be possible (examples: [1](https://github.com/dpkp/kafka-python/blob/f19e4238fb47ae2619f18731f0e0e9a3762cfa11/kafka/coordinator/consumer.py#L627), [2](https://github.com/dpkp/kafka-python/blob/f19e4238fb47ae2619f18731f0e0e9a3762cfa11/kafka/coordinator/consumer.py#L747)), but wasn't sure if those would necessarily be useful. 

Let me know if I should do anything else for this PR!

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2335)

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -355,13 +355,14 @@ class KafkaAdminClient(object):
         }
         return groups_coordinators
 
-    def _send_request_to_node(self, node_id, request):
+    def _send_request_to_node(self, node_id, request, wakeup=True):
         """Send a Kafka protocol message to a specific broker.
 
         Returns a future that may be polled for status and results.
 
         :param node_id: The broker id to which to send the message.
         :param request: The message to send.
+        :param wakeup: Optional flag to disable thread-wakeup.
         :return: A future object that may be polled for status and results.
         :exception: The exception if the message could not be sent.
         """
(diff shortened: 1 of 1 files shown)


yzhan289:
Closing this PR since it's unlikely to be reviewed due to https://github.com/dpkp/kafka-python/issues/2290

jeffwidman:
Sorry, I missed this previously. I'm 👍 on this, and since it's in the admin client it's not too intrusive into core. I heavily refactored a ton of the admin client and the datadog check, so I am pretty comfortable with this section of code. If you re-open/rebase I can merge.

That said, I probably won't be cutting a release since that's up to @dpkp but at least you can land on master and then pin to a sha upstream in datadog.

yzhan289:
@jeffwidman sounds good, I can reopen and rebase the PR!

jeffwidman (approved):
Flying blind due to https://github.com/dpkp/kafka-python/issues/2356, but this looks like a pretty low-risk change.

Issue #2356, opened 2023-03-02
CI is broken


jeffwidman:
No CI checks are currently running for any PRs.

We used to use Travis, and probably just need to migrate them to Actions.

wbarnha:
This will be closed by https://github.com/dpkp/kafka-python/issues/2290!

PR #2357, opened 2023-03-04
Migrate from Travis to GitHub Actions.

vitaliyf:
@jeffwidman, this is a first cut at addressing #2356.

I attempted to convert from Travis to GitHub actions as-is. The only exceptions is that old Python 3.x are no longer supported by GitHub, so I dropped them. I also dropped pypy2.7.

There are failures, but I didn't investigate too much into why and whether they are because of GitHut itself. I don't have too much expertise in kafka-python itself, but I'm happy to provide a bit of GitHub Actions help.

You can see results of the current workflow here: https://github.com/vitaliyf/kafka-python/actions/runs/4332483559

If you want to me to continue pursuing this, the questions I have are:
- Should Python 2.7 be dropped?
- Should we add newer Python 3.x all the way to 3.11?
  - Do you also want pypy 3.8 and 3.9?
-  Should we drop antique Kafka versions, and if so - how far?
- Should we add latest Kafka versions, all the way to 3.4?
  - Java 8 was deprecated in Kafka 3.0+, should we switch to 11 or 17?
- Do we want to continue using tox or just run pytest/pylint directly?

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2357)


jeffwidman:
Thanks! 

However, right now I'm not investing much time actively maintaining this project... that may change down the road, but at least right now it's not something I can spend much time on.

Perhaps you'd be interested in helping out as an active maintainer? 
See: https://github.com/dpkp/kafka-python/issues/2290.

Issue #2344, opened 2023-01-11
Support for admin method: deleterecords (kip-107)


michaelandrepearce:
This is an ask if support for delete records admin function can be exposed in the admin client as like is available in java admin client

dpkp:
Added in #2535

PR #2062, opened 2020-06-01
Add delete records method support for kafka admin api

10101010:
Hello. Sometimes it might be useful to delete messages from kafka topic without a trick with setting 1 second retention time.

I tried to implement this like in java version of client 
https://github.com/apache/kafka/blob/f98e176746d663fadedbcd3c18312a7f476a20c8/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L2279

For me it solves the problem, hope it can be helpful for somebody.

This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2062)

PR #2359, opened 2023-03-19
feat AdminClient: support delete_records

alfaix:
Hey!

This PR is based on https://github.com/dpkp/kafka-python/pull/2062
It implements delete_records for AdminClient. Thanks to @10101010 for the initial PR.

This PR contains the following improvements on that version:
1. It fixes review comments: adds tests and removes formatting peculiarities
2. It contains an optimization that is important to me: if multiple partitions live on the same broker, it sends one request, instead of one per partition. It is important because I use delete_records in integration tests, and this significantly improves test startup/teardown speed.
3. It checks for errors as much as it can.

I tested it locally with 2.6.0 + py3.9, I'm assuming CI will check for everything else.
Additionally, I would like some direction from maintainers on whether it is possible/required to test this with a multi-broker setup - I couldn't find an example in the tests that uses multiple brokers.

Resolves: #2344

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2359)


dpkp:
Rebased in #2535

PR #2374, opened 2023-07-12
Strip trailing dot off hostname used for SSL validation.

voutilad:
When using FQDN's with trailing dots to connect to a broker, the trailing dot in the hostname should be stripped before using the hostname for validation against the server's certificate.

For instance, this code:

```python
producer = KafkaProducer(
        bootstrap_servers="redpanda.redpanda.svc.cluster.local.:9093",
        security_protocol="SSL",
)
```

Will cause certificate validation to fail when the subject name doesn't contain the trailing dot.

For some context, this is also a problem with librdkafka: https://github.com/confluentinc/librdkafka/issues/4348

The current stance from the OpenSSL team is this is an application layer issue: https://github.com/openssl/openssl/issues/11560

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2374)


voutilad:
For additional context, this was raised to the Python project and was decided that it, too, is a problem for the application layer: https://github.com/python/cpython/issues/76178

wbarnha (commented):
I'm cautious about adding an invocation of `rstrip(".")` to each hostname attached to the socket. I think this should be the responsibility of users to add this logic. Sure, it makes sense to remove dots from the end of all hostnames, but I don't think it should be the library's responsibility to do so. The library assumes an arbitrary hostname is passed in, not specifically FQDNs.

voutilad:
> Sure, it makes sense to remove dots from the end of all hostnames, but I don't think it should be the library's responsibility to do so. 

The issue here is `self.host` is used in 2 different capacities with different semantics:

1. passed to `socket.getaddrinfo` by the `dns_lookup` function
2. passed to the `wrap_socket` method of `ssl.SSLContext`

There are reasons to an FQDN (terminated with `.`) to the DNS resolver via the `getaddrinfo` call. One reason may be for [performance](https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html).

As for `SSLContext`, the trailing dot shouldn't be used. See RFC 6066's comment in [Section 3: Server Name Indication](https://datatracker.ietf.org/doc/html/rfc6066#section-3):

> "HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client.  The hostname is represented as a byte
   string using ASCII encoding without a trailing dot.

A real-world example is seen in the .NET Runtime: https://github.com/dotnet/runtime/pull/62540

My stance is for `getaddrinfo`, a dot is valid. For `wrap_socket` and passing it to `SSLContext`, it's invalid. Since there's only 1 input from the application to the library, the library must do the right thing because it's impossible for the application to do so.

wbarnha (approved):
I see now, thank you for the clarification! LGTM.

wbarnha:
Just a note: I'm waiting to get v2.0.3 released before merging this.

PR #2383, opened 2023-08-06
Set SSLContext minimum version to TLSv1_2

wbarnha:
This change should enforce usage of stronger cryptographic protocols. More information is available at https://cwe.mitre.org/data/definitions/327.html.

This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2383)


wbarnha:
Looks like this will break support for Kafka v0.8.2.2, the test is stuck running for reasons I need to look into.

dpkp:
We provide `ssl_context` configurations for users who want to enforce this minimum.

Issue #1583, opened 2018-09-04
Handling "Fetch to node %i failed: Cancelled: ..." error


xpyy:
Hello.

I have kafka with 3 brokers on different servers. I use kafka-python to read messages from one topic until i get timeout, like so:
```python
def handle(self):
    consumer = KafkaConsumer(topic, **settings)
    while True:
        message = consumer.poll(timeout)
        if not message:
            break

        process_message(message)

    consumer.close()
```

I'm getting a lot of errors like "Fetch to node %i failed: Cancelled: ..." on 1 or 2 brokers.  Debug log tells me that they actually have in flight requests with type "FetchRequest_v3".

My questions:

1. Am i do something wrong?
2. If not, how can i fix it?

pavan538:
when ever the the socket is closed for a Node , All the pending in-flight-request will be failed with this exception. Default: kafka.errors.KafkaConnectionError.
for more insight please refer - https://github.com/dpkp/kafka-python/blob/master/kafka/conn.py BrokerConnection class clcose() - method.

`def close(self, error=None):
        """Close socket and fail all in-flight-requests.
        Arguments:
            error (Exception, optional): pending in-flight-requests
                will be failed with this exception.
                Default: kafka.errors.KafkaConnectionError.
        """
        if self.state is ConnectionStates.DISCONNECTED:
            if error is not None:
                log.warning('%s: Duplicate close() with error: %s', self, error)
            return
        log.info('%s: Closing connection. %s', self, error or '')
        self.state = ConnectionStates.DISCONNECTING
        self.config['state_change_callback'](self)
        self._update_reconnect_backoff()
        self._close_socket()
        self.state = ConnectionStates.DISCONNECTED
        self._sasl_auth_future = None
        self._protocol = KafkaProtocol(
            client_id=self.config['client_id'],
            api_version=self.config['api_version'])
        if error is None:
            error = Errors.Cancelled(str(self))
        while self.in_flight_requests:
            (_, future, _) = self.in_flight_requests.popleft()
            future.failure(error)
        self.config['state_change_callback'](self)`

joekohlsdorf:
So what is the way to handle this? Ignoring an error isn't optimal
There is no straightforward way to catch these exceptions because this is an asynchronous call and they are raised by futures.

The documentation on `consumer.close()` is also misleading:
`Close the consumer, waiting indefinitely for any needed cleanup.`
[...]

PR #2402, opened 2023-10-16
Change loglevel of cancelled errors to info

Laityned:
See #1583 

When closing the connection on purpose by calling `consumer.close()`, each ongoing Fetch request will be cancelled, by setting a Cancelled KafkaException on that particular request.

Due to the error callback on this object, which is nothing else except a logging statement, the cancellation will be logged with level ERROR, instead as INFO, while it is not actually an error, but a way of stopping the ongoing request

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2402)


Laityned:
@wbarnha

ajaynegi-et (approved):

dpkp:
Merged in #2467

PR #2415, opened 2023-11-18
Update license_file to license_files

micwoj92:
The former has been deprecated since setuptools 56

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2415)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,7 +40,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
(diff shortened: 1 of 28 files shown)


50-Course (changes requested):
Hello,

Thank you for making this patch. I think it would be best if old diff is maintained since, at the moment, `kafka-python` package is not using multiple Licenses.

micwoj92:
It does not matter.
https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#metadata

wbarnha:
We'll merge this in after the next release.

wbarnha (approved):

dpkp:
Rebased: #2462

PR #2428, opened 2024-02-23
Franco.AMVP-89449.Use OffsetFetchRequest_v3 in _send_offset_fetch_request

royfranco:
Due to a bug in kafka, empty response can return instead of an error for OffsetFetchRequest_v1, since v1 does not support top-level errors. This issue resulted in an incident where the consumer interpret the bad response as "the consumer group does not have offsets for the topic partition", and therefore it reset the offsets according to the auto-reset-policy which was set to earliest.

In order to avoid the problem, this fix makes sure to use v3, and to check for top-level errors in the response.

The bug is described extensively [here](https://armis-security.slack.com/archives/C02FGA4D85V/p1708585371380439).
The issue was discovered during the deployment of v191, and was managed in the slack channel [#incident-st-kafka-consumer-offset-reset](https://armis-security.slack.com/archives/C06LDPKJ8GG/p1708585041128199).
To prevent the issue, the 191 change was reverted [here](https://github.com/ArmisSecurity/armis/pull/72646).

Following merging that PR, we will move the armis repo to use this custom fork, and revert https://github.com/ArmisSecurity/armis/pull/72646.

### DIFF
--- a/kafka/coordinator/consumer.py
+++ b/kafka/coordinator/consumer.py
@@ -731,7 +731,12 @@ class ConsumerCoordinator(BaseCoordinator):
         for tp in partitions:
             topic_partitions[tp.topic].add(tp.partition)
 
-        if self.config['api_version'] >= (0, 8, 2):
+        if self.config['api_version'] >= (0, 10, 2):
+            request = OffsetFetchRequest[3](
+                self.group_id,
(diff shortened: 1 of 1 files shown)


royfranco:
FYI, This PR was created by accident in that repo.

PR #2449, opened 2024-10-22
added group.instance.id to consumer group

TeisNP:
Added group.instance.id to consumer group config parameters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/dpkp/kafka-python/2449)


dpkp:
It looks like this PR only adds a configuration value, but it doesn't use it anywhere?

PR #2466, opened 2025-02-10
Fix crc32c's __main__ for Python 3

dpkp:
Rebase of #2298


dpkp:
We don't use the `__main__` function in this code, and I'm not comfortable breaking python2 right now. So going to close this.

PR #2479, opened 2025-02-14
Remove unreachable raise in __cmp__ method

daniloarodrigues:
Motivation:
The raise statement was unreachable code and never executed after the return statement, causing confusion and redundancy.

Modifications:
Removed the `raise NotImplementedError` call in the `__cmp__` method.

Result:
The `__cmp__` method is now clearer, avoiding any unused or unreachable code.


dpkp:
This is a vendored file; not accepting bespoke modifications.

Issue #2400, opened 2023-10-09
CPU spikes upon broker replacement


antonakospanos:
Hello @dpkp @wbarnha,

We have used `kafka-python` to interact with our Kafka brokers since 2020. 

However, during the last Kafka Broker upgrades (i.e. 3.3 => 3.5) we experienced spikes in the CPU utilization of the broker's nodes. The same happens upon hardware failure in the kafka cluster that results in replacing a broker with another node.
![production](https://github.com/dpkp/kafka-python/assets/42936184/b57a41fd-103d-4c87-acee-dddf2a050a52)

The issue is mitigated by restarting the consumers in our python apps. Aiven responded that this is a known issue on apps relying on `kafka-python` and we should switch to https://github.com/confluentinc/confluent-kafka-python. 

Here is their whole response:
> We also noticed that the impacted consumers are those based on client library [kafka-python-2.0.2](https://github.com/dpkp/kafka-python) , which one was last updated 3 years ago, and is based on an older version of Kafka protocol. In Kafka 3.3.1, we are aware that this older protocol has lost in efficiency, why generally leads to more latency and consumer rebalances. The short-term solution is to relax your configurations as follow:

>* Increase broker value for [group.initial.rebalance.delay.ms](http://group.initial.rebalance.delay.ms/) (default: 3s).
> * Increase consumer value for [session.timeout.ms](http://session.timeout.ms/) (default: 45s).
> * Make sure that the consumer value of [heartbeat.interval.ms](http://heartbeat.interval.ms/) (default: 3s) is lower than [session.timeout.ms](http://session.timeout.ms/), but not higher than 1/3 of that value.

> The long term solution is to migrate these consumers to client library [confuent-kafka-python-2.2.0](https://github.com/confluentinc/confluent-kafka-python) which one offers a similar SDK, and is up-to-date.

Is there any development in progress so that we tackle this issue? 

If not we will have to switch to another kafka library.

wbarnha:
Sorry to hear that you're having issues. Unfortunately I'm not in a position to do new releases, so even if I did have a solution, I couldn't get it pushed out into the mainstream.

I do acknowledge the older version of the Kafka Protocol is still implemented in this project. To be honest, `aiokafka` and `confluent-kafka-python` are more up-to-date with the current state of the Kafka Protocol and have more underlying optimizations. I'm not surprised that `kafka-python` is not handling the newer versions of Kafka so well, since this project is currently geared towards supporting older implementations.

[...]

PR #2480, opened 2025-02-14
Improve client networking backoff / retry

dpkp:
The primary goal of this PR is to reduce busy-polling during broker and network failures. Several incremental changes are included:

* Merge BrokerConnection blacked_out() and connection_delay() logic and eliminate delays / blackouts when cycling through entries from a single dns lookup.
* Improve request-timeout handling in client.poll() by checking all connections for the next potential in-flight-request timeout.
* Drop 100ms timeout override when there are no in-flight requests (added prior to full async connection support)
* Do not mark connection as sending if future immediately resolves (error).
* Remove flat 100ms timeout for unfinished futures (added prior to full async connection support)
* Respect connection delays during metadata refresh when all brokers are unavailable.
* Honor reconnect backoff in connection_delay when connecting. This is primarily used by the producer send thread to defer sending accumulated batches. It was originally 0ms but was changed to `float('inf')` to avoid busy polling. Really this should just be the reconnect backoff delay so that we both avoid busy polling and also avoid starving the sender thread.
* Increase default `reconnect_backoff_max_ms` to 30000 (30 secs). This should make defaults more appropriate for production use, especially in larger fleets.

### DIFF
--- a/kafka/admin/client.py
+++ b/kafka/admin/client.py
@@ -72,7 +72,7 @@ class KafkaAdminClient(object):
             reconnection attempts will continue periodically with this fixed
             rate. To avoid connection storms, a randomization factor of 0.2
             will be applied to the backoff resulting in a random range between
-            20% below and 20% above the computed value. Default: 1000.
+            20% below and 20% above the computed value. Default: 30000.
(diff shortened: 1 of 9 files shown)


dpkp:
This should help w/ #2400

PR #2493, opened 2025-02-24
Call ApiVersionsRequest during connection, prior to Sasl Handshake

dpkp:
Moves ApiVersionsRequest into pre-connection state handling, prior to SaslHandshake. Use v2 for initial request and check for UnsupportedVersionError.

Skips ApiVersionsRequest if `api_version` has been configured by user.

### DIFF
--- a/kafka/client_async.py
+++ b/kafka/client_async.py
@@ -303,7 +303,7 @@ class KafkaClient(object):
 
     def _conn_state_change(self, node_id, sock, conn):
         with self._lock:
-            if conn.connecting():
+            if conn.state is ConnectionStates.CONNECTING:
                 # SSL connections can enter this state 2x (second during Handshake)
                 if node_id not in self._connecting:
                     self._connecting.add(node_id)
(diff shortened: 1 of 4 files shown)


ijuma:
Nice to see all the protocol support improvements. Are we also planning to support [KIP-511](https://cwiki.apache.org/confluence/display/KAFKA/KIP-511%3A+Collect+and+Expose+Client%27s+Name+and+Version+in+the+Brokers)? That's very helpful for users to understand which clients and versions are being used within their applications.

dpkp:
> Nice to see all the protocol support improvements. Are we also planning to support [KIP-511](https://cwiki.apache.org/confluence/display/KAFKA/KIP-511%3A+Collect+and+Expose+Client%27s+Name+and+Version+in+the+Brokers)? That's very helpful for users to understand which clients and versions are being used within their applications.

Yes, but as I understand it that change requires flexible version support / tags and so I'll likely tackle after getting to 2.1 baseline.

PR #2516, opened 2025-03-08
Remove the "mock" dependency.

Romain-Geissler-1A:
"mock" is included inside the standard "unittest" package in Python >= 3.3.


dpkp:
Thanks, but we haven't officially dropped python 2.7 support yet.

PR #2518, opened 2025-03-11
Make the "mock" dependency optional (only used in Python 2.7).

Romain-Geissler-1A:
"mock" is included inside the standard "unittest" package in Python >= 3.3.

(This is an alternatve version of what was proposed (and declined) earlier in #2516. The idea is to allow distributions to package kafka-python without having to patch it explicitly to remove the "mock" dependency).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ crc32c = ["crc32c"]
 lz4 = ["lz4"]
 snappy = ["python-snappy"]
 zstd = ["zstandard"]
-testing = ["pytest", "mock", "pytest-mock"]
+testing = ["pytest", "mock; python_version < '3.3'", "pytest-mock"]
 
 [tool.setuptools]
 include-package-data = false
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ crc32c
 docker-py
 flake8
 lz4
-mock
+mock; python_version < '3.3'
 py
 pylint
 pytest
--- a/test/record/test_default_records.py
+++ b/test/record/test_default_records.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 import pytest
-from mock import patch
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch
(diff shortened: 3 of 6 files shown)


dpkp:
I think I can merge this, thanks for the update. Can you show me what commands you are using to package that break on the current master? Our intent was to make testing dependencies / requirements-dev optional and wouldn't expect them to be included generally. The are not included in wheels uploaded to pypi, for example.

Romain-Geissler-1A:
Actually we have upgraded python-kafka in Fedora + EPEL 10 this week-end, including the diff of the closed pull request #2516: https://src.fedoraproject.org/rpms/python-kafka/pull-request/4#request_diff They way most fedora python packages are built are via some `%pyproject_` macros, which honestly I don't really know what they do. See the spec file here: https://src.fedoraproject.org/rpms/python-kafka/blob/rawhide/f/python-kafka.spec The most basic use case is to read the `pyproject.toml` and look for dependencies there, but I have seen other package defining their dependencies either a `tox.ini` file or various `requirements.txt` files. These macro will try to discover dependencies this way, and "convert" these to actual RPM packages dependencies. In Fedora, `python-mock` has been removed already, no one is meant to depend on it anymore, and thus right now trying to package `python-kafka` without any patch results in a missing `mock` dependency. Note that usually fedora packages try to run tests (which often catches some breakages when something changes).

pyproject.toml
```diff
@@ -37,7 +37,7 @@ crc32c = ["crc32c"]
 lz4 = ["lz4"]
 snappy = ["python-snappy"]
 zstd = ["zstandard"]
-testing = ["pytest", "mock", "pytest-mock"]
+testing = ["pytest", "mock; python_version < '2.7'", "pytest-mock"]
```
dpkp, line 40: can we make this `< '3.3'` ?

Romain-Geissler-1A:
I have update the version constraint to ` < '3.3'` everywhere as suggested.

dpkp:
sorry, i also removed `tox.ini` and you'll need to rebase against that change before I can merge :(

dpkp (approved):

dpkp:
Thanks!

PR #2560, opened 2025-03-22
KIP-70: Auto-commit offsets on consumer.unsubscribe(), defer assignment changes to rejoin 

dpkp:


kafka/consumer/subscription_state.py
```diff
@@ -1,6 +1,10 @@
 from __future__ import absolute_import
 
 import abc
+try:
+    from collections import Sequence
+except ImportError:
+    from collections.abc import Sequence
```
millerdev, line 7: This causes `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working`

I think the imports should be reversed here:
```suggestion
    from collections.abc import Sequence
except ImportError:
    from collections import Sequence
```

dpkp: thanks, yea we still support very old pythons (2.7). I did `try old, except new` but that does cause deprecation warnings for interim pythons. I can see how you might prefer `try new, except old`

PR #2563, opened 2025-03-24
Fix maximum version to send an OffsetFetchRequest

garnaud:
Max version has been increased to 5 but the method raises an NotImplementedError error due to a wrong condition.


dpkp:
Thanks!

Issue #2632, opened 2025-05-22
SaslMechanism uses empty host-config if sasl_kerberos_domain_name is not set


SeppMe:
According to the manual, `sasl_kerberos_domain_name` should default to one of the bootstrap servers, if not set. However, this does not currently work. The lines
```python
kerberos_domain_name = config.get('sasl_kerberos_domain_name', '') or config.get('host', '')
```
in the respective `__init__` methods of the various SaslMechanisms try to access `config['host']`, which is never set in the `BrokerConnection` object. Thus, the `get` defaults to an empty string, which is effectively the local host when a connection is attempted down the line. Which in general is wrong and fails. The `SaslMechanismAwsMskIam` should not even work at all, as it makes an assertion that `host` is in the config, but I cannot test that right now, because my environment is not set up for it.

I think, a simple `self.config['host'] = host` in the `BrokerConnection`'s `__init__` before `self._init_sasl_mechanism()` should do. As far as I can tell, config['host'] gets used nowhere else, so there should not be unexpected side-effects. I am preparing a pull request for this + tests.

PR #2633, opened 2025-05-22
Set the current host in the SASL configs

SeppMe:
This fixes issue #2632


kafka/conn.py
```diff
@@ -256,6 +256,7 @@ def __init__(self, host, port, afi, **configs):
         for key in self.config:
             if key in configs:
                 self.config[key] = configs[key]
+        self.config['host'] = host
```
dpkp, line 259: Since this is only used in sasl mechanisms, I think it would be better to limit the scope. How about adding `host=self.host` on L316 instead?

PR #2640, opened 2025-06-04
Do not ignore metadata response for single topic with error

dpkp:
Improve behavior when single topic is unauthorized.


jp-hammer (commented):
@jd-hammer I really need to update my crypto please been struggling

PR #2660, opened 2025-07-01
fix flush_in_progress not visible to sender

llk89:
Call site

https://github.com/dpkp/kafka-python/blob/2abf2cd6458dcb069940c04da2711aa41683f4f4/kafka/producer/sender.py#L285

This is called in Sender._maybe_send_transactional_request(), but over there it's called without `_` prefix. Since this method is accessed by other classes I assume we should remove the prefix, instead of making the caller call `_flush_in_progress` instead


dpkp:
thanks for finding these!

PR #2665, opened 2025-07-15
Add missing argument in debug log

afeblot:
The change is self explanatory and fixes a TypeError due to missing partition argument in a debug log.


dpkp:
Thanks!

PR #2666, opened 2025-07-23
WIP: support looking up remote name resolving with SOCKS5

jschwartzenberg:
Hi all, I'm dealing with a Kafka cluster behind a SOCKS proxy where the broker names cannot locally be resolved. I saw that the current SOCKS5 support doesn't handle remote look-ups, so I started looking into implementing it. Unfortunately I'm not that familiar with either Kafka or Python, so if anybody could look along that would be great.

Here is my first attempt. I see that it either fails to initialize the socket, `socket.socket(afi, sock_type)` returning none on line 78 in `socks5_wrapper.py` or getting errno 10035 when trying to initialize things. It is not even hitting my new part yet, so it's probably something silly.

If anybody happens to spot what's going wrong, that would be great, otherwise I'll look further these days.


kafka/conn.py
```diff
@@ -366,35 +366,46 @@ def connect_blocking(self, timeout=float('inf')):
...
                         self._sock = self._socks5_proxy.socket(self._sock_afi, socket.SOCK_STREAM)
                     else:
                         self._sock = socket.socket(self._sock_afi, socket.SOCK_STREAM)
                 except (socket.error, OSError) as e:
                     self.close(e)
                     return self.state
+            else:
+                self._sock = self._socks5_proxy.socket(None, socket.SOCK_STREAM)
+                self._sock_afi = None
+                self._sock_addr = [self.host.encode('ascii'), self.port]
```
dpkp, line 395: recommend you defer `str.encode()` until needed by socks code.

kafka/conn.py
```diff
@@ -366,35 +366,46 @@ def connect_blocking(self, timeout=float('inf')):
...
         if self.state is ConnectionStates.DISCONNECTED and not self.blacked_out():
             self.state = ConnectionStates.CONNECTING
             self.last_attempt = time.time()
-            next_lookup = self._next_afi_sockaddr()
-            if not next_lookup:
-                self.close(Errors.KafkaConnectionError('DNS failure'))
-                return self.state
-            else:
+            if self.config["socks5_proxy"] is None or not self._socks5_proxy.use_remote_lookup():
+                next_lookup = self._next_afi_sockaddr()
```
dpkp, line 376: Would it be simpler to move the `self._socks5_proxy.use_remote_lookup()` inside of `_next_afi_sockaddr()` ? For socks proxy w/ remote_lookup, it should return `(socket.AF_UNSPEC, (self.host, self.port))`. With that change I think most of these other changes would be unnecessary.

kafka/conn.py
```diff
@@ -862,7 +873,7 @@ def connection_delay(self):
         large number to handle slow/stalled connections.
         """
         if self.disconnected() or self.connecting():
-            if len(self._gai) > 0:
+            if len(self._gai) > 0 or self._socks5_proxy.use_remote_lookup():
```
dpkp, line 876: won't this crash if/when self._socks5_proxy is None and self._gai is empty?

kafka/conn.py
```diff
@@ -366,35 +366,46 @@ def connect_blocking(self, timeout=float('inf')):
...
+                self._sock_afi = None
+                self._sock_addr = [self.host.encode('ascii'), self.port]
 
             for option in self.config['socket_options']:
                 log.debug('%s: setting socket option %s', self, option)
                 self._sock.setsockopt(*option)
 
             self._sock.setblocking(False)
             self.config['state_change_callback'](self.node_id, self._sock, self)
+            if self._sock_afi != None:
```
dpkp, line 403: generally should check `is` or `is not` for None, not equality.

kafka/conn.py
```diff
@@ -1291,9 +1302,13 @@ def check_version(self, timeout=2, **kwargs):
             return self._api_version
 
     def __str__(self):
+        if self._sock_afi != None:
```
dpkp, line 1305: implied suggestion above is to use `socket.AF_UNSPEC` instead of None

jschwartzenberg:
Many thanks for your review! I'll rework and push a better commit!

kafka/conn.py
jschwartzenberg: Thanks, that indeed reduces the changes quite a bit! Let me know what you think of the current version!

kafka/conn.py
jschwartzenberg: You're completely right. I added some extra logic now to check whether `self._socks5_proxy` is set.

kafka/socks5_wrapper.py
```diff
@@ -200,14 +210,28 @@ def connect_ex(self, addr):
...
+                    addr_len,
+                    addr[0].encode('ascii'),
+                )
+            else:
+                # either 4 (type 1) or 16 (type 4) bytes of actual address
+                self._buffer_out += struct.pack(
+                    "!{}s".format(addr_len),
+                    socket.inet_pton(self._target_afi, addr[0]),
+                )
+            self._buffer_out += struct.pack("!H", addr[1])  # port
```
dpkp: I changed the pack format from `h` to `H`. Unsigned int is required for ports >32k

PR #2670, opened 2025-08-19
Fix thread not waking up when there is still data to be sent

gqmelo:
When producing messages quickly without waiting for the future of previous requests, there could  be some situations when the last batch was not sent.

That seemed to be more frequent with larger messages (~100KiB), but apparently it could happen to any message when `linger_ms` is 0. Not sure if it could happen when it is non-zero though.

The reason is that `BrokerConnection.send_pending_requests_v2` would fill the internal buffer with the bytes from a request and try to send it.

https://github.com/dpkp/kafka-python/blob/512d0a0b8d71cf7f34f1b23f8a42d52c28af3266/kafka/conn.py#L1071

If it couldn't send it completely for some reason, it would try to send again in the next call to `send_pending_requests_v2`.

But if between those 2 calls, `BrokerConnection.send` was called, new data would be appended to self._protocol: KafkaProtocol:

https://github.com/dpkp/kafka-python/blob/512d0a0b8d71cf7f34f1b23f8a42d52c28af3266/kafka/conn.py#L1015

but the second call to `send_pending_requests_v2` wouldn't check if any new data was available and would return False:

https://github.com/dpkp/kafka-python/blob/512d0a0b8d71cf7f34f1b23f8a42d52c28af3266/kafka/conn.py#L1070

This would tell `KafkaClient._poll` that all pending data was sent, which would make the client not listen to socked write readiness anymore:

https://github.com/dpkp/kafka-python/blob/512d0a0b8d71cf7f34f1b23f8a42d52c28af3266/kafka/client_async.py#L744-L748

### DIFF
--- a/test/test_conn.py
+++ b/test/test_conn.py
@@ -12,6 +12,9 @@ import pytest
 
 from kafka.conn import BrokerConnection, ConnectionStates
 from kafka.future import Future
+from kafka.conn import BrokerConnection, ConnectionStates, SSLWantWriteError
+from kafka.metrics.metrics import Metrics
+from kafka.metrics.stats.sensor import Sensor
 from kafka.protocol.api import RequestHeader
 from kafka.protocol.group import HeartbeatResponse
 from kafka.protocol.metadata import MetadataRequest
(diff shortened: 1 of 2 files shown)


test/test_conn.py
```diff
@@ -228,6 +240,46 @@ def test_send_response(_socket, conn):
...
+    # This won't send any bytes because of the SSL exception and the request bytes will be kept in
+    # the buffer.
+    assert conn.send_pending_requests_v2() is False
+    assert bytes_sent_sensor.record.call_args_list[0].args == (0,)
+
+    conn.send(req2, blocking=False)
+    # This will send the remaining bytes in the buffer from the first request, but should notice
+    # that the second request was queued, therefore it should return False.
+    bytes_sent_sensor.record.reset_mock()
+    assert conn.send_pending_requests_v2() is False
```
gqmelo, line 271: This is where the test would fail before the fix.

dpkp:
Yes, thank you!

Issue #2667, opened 2025-07-30
Consumer randomly stucks and timeouts when communicating with confluent cloud


roman-bartusiak-yohana:
Issue: 

We observer random timeouts of the consumer. We have contacted confluent on that - and they do not see anything wrong on broker side. CPU/MEM of the consumer is not high and limited during those events. The only correlation we observe is when we stat a lot (50) consumers per pod - it can happen more often then. Increasing timeout is rather a not-go as it is already 5m.  It is not corellating with high load.  

Sample logs
```

warn | Jul 24 20:35:17.924 | i-094c46e288d48e5ce | aidp-api | Marking the coordinator dead (node coordinator-5) for group aidp-chat_completion-request-consumer: [Error 7] RequestTimedOutError: Request timed out after 305000.0 ms.
-- | -- | -- | -- | --
error | Jul 24 20:35:17.924 | i-094c46e288d48e5ce | aidp-api | Error sending JoinGroupRequest_v4 to node coordinator-5 [[Error 7] RequestTimedOutError: Request timed out after 305000.0 ms]
error | Jul 24 20:35:17.924 | i-094c46e288d48e5ce | aidp-api | [IPv4 ('34.211.165.150', 9092)]>: Closing connection. [Error 7] RequestTimedOutError: Request timed out after 305000.0 ms
warn | Jul 24 20:35:17.924 | i-094c46e288d48e5ce | aidp-api | [IPv4 ('34.211.165.150', 9092)]> timed out after 305000.0 ms. Closing connection.
```

1. Stack
   * kafka-pytho: 2.2.9
   * python  3.12
   * agains confluent cloud
   * multiple consumers in separate threads - not sharing consumer instances, every thread has a separate consumer
   * 200 partitions on the topic 
   * up to 50 listener threads 
   * poll interface with timeout
   * consumer config: 
[...]

PR #2672, opened 2025-08-21
Add a check to ensure broker connection is ready during poll if auto commit is disabled

TribuneX:
Potentially fixes https://github.com/dpkp/kafka-python/issues/2667

If `enable_auto_commit=False` the CPU usage of a kafka consumer increases to 100% if the broker connection is lost. If `enable_auto_commit=True`, the connection loss is detected by the periodic auto commit, which acts as a sort of health check in this case.

I added a similar check to the `poll` method. I am not sure if this the best location, thats why I also did not add any test yet. 
We might also only check the readiness in case `enabled_auto_commit=False`. 

Looking forward to your feedback.

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -726,6 +726,10 @@ class KafkaConsumer(six.Iterator):
             log.debug('poll: timeout during coordinator.poll(); returning early')
             return {}
 
+        node_id = self._coordinator.coordinator()
+        if node_id is None or not self._client.ready(node_id, metadata_priority=False):
+            return {}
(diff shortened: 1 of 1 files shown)


roman-bartusiak-yohana:
To confirm, we have added connection liveness check and it fixed all the issues - with both CPU and wired UnknownMembeID exceptions

TribuneX:
Any updates @dpkp since we would love to integrate this into an official release?

dpkp:
After looking at this more closely I think the best fix is #2695

PR #2714, opened 2026-02-10
Use time.monotonic() instead of time.time() for elapsed time calculations

### LINKED ISSUE #1546: Using non monotonic functions to calculate elapsed time
on master (commit 9ac3cb1ec220ff9968a8b003b02e98dd11cc486b) https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/group.py , 
method `def poll(self, timeout_ms=0, max_records=None)` uses `time.time()` to calculate elapsed time.
`time.time()` does not return a monotonic clock , and therefore the elapsed time calculation could yield unexpected results.

`>>> time.get_clock_info('time')
namespace(adjustable=True, implementation='clock_gettime(CLOCK_REALTIME)', monotonic=False, resolution=1.0000000000000002e-06)`

the `time` module has several monotonic times available, but I don't believe they're available for python 2, which this library supports

veeceey:
## Summary

- Replaces all 92 uses of `time.time()` with `time.monotonic()` across 18 source files for measuring elapsed time, timeouts, and latency
- Updates 7 test files to mock `time.monotonic` instead of `time.time`, ensuring all unit tests continue to pass
- Preserves `time.time()` in `kafka/record/legacy_records.py` where wall clock time is needed for Kafka message timestamps

## Motivation

`time.time()` returns wall clock time which is **adjustable** -- it can jump forward or backward due to NTP synchronization, DST changes, or manual clock adjustments. Using it for elapsed time calculations can cause:

- **Premature timeouts**: if the clock jumps forward, operations may time out before their actual deadline
- **Infinite hangs**: if the clock jumps backward, timeout conditions may never be satisfied
- **Incorrect latency metrics**: recorded latencies can be negative or wildly inflated

`time.monotonic()` provides a clock that is guaranteed to never go backward, making it the correct choice for all interval/elapsed time measurements.

```python
>>> import time
>>> time.get_clock_info('time')
namespace(adjustable=True, implementation='clock_gettime(CLOCK_REALTIME)', monotonic=False, ...)
>>> time.get_clock_info('monotonic')
namespace(adjustable=False, implementation='clock_gettime(CLOCK_MONOTONIC)', monotonic=True, ...)
```

Fixes #1546

## Test plan

- [x] All existing unit tests pass (1656 passed, 28 skipped -- the 1 pre-existing failure in `test_crc32c[None]` is unrelated)
- [x] Updated test mocks in `test_metrics.py`, `test_client_async.py`, `test_conn.py`, `test_coordinator.py`, `test_sender.py`, `test_fetcher.py`, and `testutil.py` to patch `time.monotonic` instead of `time.time`
- [x] Verified `kafka/record/legacy_records.py` still uses `time.time()` for Kafka message timestamps (wall clock time is correct there)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/integration/test_consumer_group.py
+++ b/test/integration/test_consumer_group.py
@@ -73,9 +73,9 @@ def test_group(kafka_broker, topic):
         threads[i] = t
 
     try:
-        timeout = time.time() + 15
+        timeout = time.monotonic() + 15
         while True:
-            assert time.time() < timeout, "timeout waiting for assignments"
+            assert time.monotonic() < timeout, "timeout waiting for assignments"
             # Verify all consumers have been created
             missing_consumers = set(consumers.keys()) - set(range(num_consumers))
             if missing_consumers:
@@ -128,7 +128,7 @@ def test_group(kafka_broker, topic):
         while True:
             for c in range(num_consumers):
                 heartbeat = consumers[c]._coordinator.heartbeat
-                last_hb = time.time() - 0.5
+                last_hb = time.monotonic() - 0.5
                 if (heartbeat.heartbeat_failed or
                     heartbeat.last_receive < last_hb or
                     heartbeat.last_reset > last_hb):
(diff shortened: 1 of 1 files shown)


dpkp:
Looks like `test/integration/test_consumer_group.py::test_heartbeat_thread` is consistently failing on gha. You said this test passes locally? Any ideas what might be happening?

PR #2725, opened 2026-03-09
Fix TaggedFields value encoding; add test coverage

dpkp:
TaggedFields is mostly unused in the current client code, so it makes sense that this bug slipped through. Fixing and adding test coverage. Specifically not testing tag order because I do not think that is part of the protocol spec (tags can be encoded in any order).


vmaurin:
@dpkp We were talking about this in an aiokafka PR and it is supposed to be encoded in order according the KIP (not sure why though) https://github.com/aio-libs/aiokafka/pull/1139#pullrequestreview-4176880280

PR #2840, opened 2026-04-09
Drop the mutable default in _estimate_size_in_bytes

basheer-cloud:
I ran into this while reading through the code path around kafka/producer/kafka.py. Drop the mutable default in _estimate_size_in_bytes. Mutable defaults are shared across calls and usually turn into state leaks. I kept the patch small and re-ran syntax checks after applying it.

### DIFF
--- a/kafka/producer/future.py
+++ b/kafka/producer/future.py
@@ -3,6 +3,7 @@ import threading
 
 from kafka import errors as Errors
 from kafka.future import Future
+from kafka.util import Timer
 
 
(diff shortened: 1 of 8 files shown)


dpkp:
Thanks - it looks like your PR inadvertently includes another commit from a different PR that recently landed. Can you rebase -i so that this PR only includes your changes?

dpkp:
Resolved conflicts w/ cherry-pick in #2865

Issue #3076, opened 2026-06-18
NetworkSelector.unschedule() leaks cancelled tasks in _pending_tasks


cozifyaki:
### Summary

`NetworkSelector.unschedule()` appears to leak cancelled scheduled tasks in kafka-python 3.0.0.

`NetworkSelector` keeps scheduled tasks strongly referenced in `_pending_tasks`, but `NetworkSelector.unschedule()` does not remove cancelled/unscheduled tasks from that set.

As a result, every scheduled task that is cancelled before firing remains strongly referenced forever.

### Why this leaks

In `kafka/net/selector.py`, scheduled tasks are added to `_pending_tasks` by `NetworkSelector.call_at()`.

However, `NetworkSelector.unschedule()` removes the task from `_scheduled`, but does not remove it from `_pending_tasks`.

That means this sequence leaks one task:

1. Schedule timeout task.
2. Request completes successfully.
3. Timeout task is cancelled / unscheduled.
4. Task stays in `_pending_tasks` forever.

For Kafka traffic this can happen continuously because successful requests cancel their response-timeout callbacks.

### Impact

Long-running consumers can leak memory steadily while polling/fetching, even when fetches return no records.

The leak rate depends on request rate. It is especially visible with low `fetch_max_wait_ms` / frequent empty fetches, but it is not limited to that configuration.

### Expected behavior

Cancelled or unscheduled tasks should be removed from `_pending_tasks`, so they can be garbage-collected.

### Actual behavior

Cancelled tasks remain in `NetworkSelector._pending_tasks` indefinitely.

### Minimal reproducer

```python
import time
from kafka.net.selector import NetworkSelector

net = NetworkSelector()

for _ in range(1000):
    task = net.call_at(time.monotonic() + 3600, lambda: None)
    net.unschedule(task)

assert len(net._scheduled) == 0
assert len(net._pending_tasks) == 1000
```

### Suggested fix

```python
def unschedule(self, task):
    if task.scheduled_at is not None:
        self._scheduled.remove((task.scheduled_at, task))
        task.scheduled_at = None
        self._pending_tasks.discard(task)
```

The important part is that tasks which will never run must not remain strongly referenced by `_pending_tasks`.

### Version

Observed by AI with kafka-python 3.0.0.

dpkp:
Thanks -- fixed in 3.0.2

PR #3077, opened 2026-06-18
fix memory leak of cancelled timeout tasks

azdobylak:
## Issue
The issue is already described in #3076. Cancelled/unscheduled tasks were not garbage collected due to strong reference kept in `_pending_tasks`. Tasks were only removed from `_pending_tasks` on completion in `_poll_once`, so `unschedule()` left them pinned by that strong reference (with their underlying coroutines left unclosed).
To confirm that issue exists and is resolved by this PR I am attaching memray screenshots. Leaking version is `kafka-python==3.0.0`.
leak
fixed
## Fix Approach:
This PR introduces `_retire_task` to drop the strong reference from `_pending_tasks` and explicitly closes the coroutine stack on cancellation/shutdown to release the frame's local variables.

### DIFF
--- a/kafka/net/selector.py
+++ b/kafka/net/selector.py
@@ -122,6 +122,21 @@ class Task:
             raise RuntimeError('Task exception is already set!')
         self._exc = exc
 
+    def close(self):
+        stack = self._stack
+        while stack:
+            coro, stack = stack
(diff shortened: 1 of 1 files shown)


dpkp:
Thanks!

PR #3078, opened 2026-06-18
Fix/net selector task already done

azdobylak:
I missed one corner case in my last PR (#3077). If the task was already in the ready queue (no longer scheduled), we must not clean it up. It shouldn't have caused any major issues thanks to the safeguards, but it still produced a misleading error message. Sorry about that! I generated a test case to cover this path and implemented the fix.

### DIFF
--- a/kafka/net/selector.py
+++ b/kafka/net/selector.py
@@ -396,8 +396,10 @@ class NetworkSelector:
         task.close()
 
     def unschedule(self, task):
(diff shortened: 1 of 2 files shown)


dpkp:
Thanks for digging in on this one; I think I'll probably rename `net.unschedule(task)` -> `net.cancel(task)` and make a few additional changes before releasing 3.0.3.

Issue #3102, opened 2026-06-24
IndexError in UnsignedVarInt32.encode_into for large records


dpkp:
See #3101 (originally filed as a PR)

dpkp:
Fixed in 3.0.5 release

PR #3103, opened 2026-06-24
Reserve buffer capacity before every encode_into write

dpkp:
The optimized encode path writes primitives directly into a growable
EncodeBuffer (initial 64KB). Only Bytes reserved space via out.ensure();
every other writer assumed room existed. When a payload forces ensure()
to grow the buffer to exactly fit (zero trailing slack, i.e. when
pos + payload > 2 * capacity), the next write ran off the end:

  - UnsignedVarInt32 / tagged-field count -> IndexError (the reported bug)
  - fixed fields (e.g. the next partition's int32 index) -> struct.error
  - arrays of fixed primitives ([]int32) overflow at 64KB with no Bytes
    or tagged fields involved at all
  - codegen kept a stale `buf` local after a delegated tagged-field
    encode reallocated out.buf (silent data loss / crash); adding
    ensure() to the varint made this fire routinely

Make capacity reservation uniform:

  - runtime codecs (types.py, tagged_fields.py) call out.ensure(N) before
    writing N bytes (fixed/uuid/bitfield/varint, String prefix+payload,
    Bytes length prefix, tagged size byte)
  - codegen gains CodegenContext.emit_reserve(): an inline
    `if pos + N > _cap` check that only syncs/grows/re-reads on overflow,
    keeping buf/pos/_cap in sync
  - ArrayField bulk-reserves fixed-element arrays once, then writes the
    run with a tight pack_into loop (keeps the hot path fast)
  - struct.py / struct_array.py re-bind buf/_cap after every delegated
    encode_into that can reallocate the buffer

Document the EncodeBuffer capacity contract and the reallocation hazard
(callers must re-read out.buf after a grow).

Adds test/protocol/schemas/test_encode_large.py: boundary unit tests for
every codec plus byte-for-byte parity and round-trip for large records
(Produce v9/v3, single and multi partition), large []int32 arrays, the
64KB/128KB grow boundaries, and pooled-buffer reuse. 22 of these fail on
the prior tree.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/kafka/protocol/schemas/fields/array.py
+++ b/kafka/protocol/schemas/fields/array.py
@@ -72,9 +72,10 @@ class ArrayField(BaseField):
         if compact:
             an = ctx.next_var('an')
             ctx.emit(indent, '%s = len(%s) + 1 if %s is not None else 0' % (an, val_expr, val_expr))
-            UnsignedVarInt32.emit_encode_into(ctx, an, indent)
+            UnsignedVarInt32.emit_encode_into(ctx, an, indent)  # reserves the length varint
             ctx.emit(indent, 'if %s is not None:' % val_expr)
         else:
+            ctx.emit_reserve(indent, 4)
             ctx.emit(indent, 'if %s is None:' % val_expr)
             ctx.emit(indent, "    pack_into('>i', buf, pos, -1)")
             ctx.emit(indent, '    pos += 4')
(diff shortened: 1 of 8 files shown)


dpkp:
Based on original report / PR #3101 (Fixes #3102)

PR #3109, opened 2026-06-28
Fix leaking io coroutines

azdobylak:
I found one more leak in the transport layer cleanup. Previously, I/O tasks were only being reclaimed when the peer (broker) closed the connection and even then, only as a side effect of the read-error path. When the broker disconnected, the socket became readable, returned an EOF error, and forced the task to abort and exit.

However, on locally-initiated closes (like idle-connection reaping, request timeouts, or rebalances), transport._close() would close the socket and unregister it without ever cancelling the parked I/O waiters. Because there was no incoming I/O event to trigger an error and wake them up, the reader and writer coroutines stayed permanently stranded in `net._pending_tasks` in a `WAIT_IO` state.

### DIFF
--- a/kafka/net/transport.py
+++ b/kafka/net/transport.py
@@ -19,6 +19,8 @@ class KafkaTCPTransport:
         self._closed = False
         self._write_buffer = deque()
         self._writing = False
+        self._read_task = None
+        self._write_task = None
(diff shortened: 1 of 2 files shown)


dpkp:
thanks!!

PR #3119, opened 2026-07-12
Fix OverflowError in offsets_for_times() when timestamp is non-finite float

koteshyelamati:
## Bug

`offsets_for_times()` converts each timestamp value with `int(ts)`. If `ts` is `float('inf')` or `float('nan')`, this raises an `OverflowError` (or `ValueError` for nan) instead of the expected `ValueError` with a clear message.

**Reproducer:**
```python
consumer = KafkaConsumer(bootstrap_servers='localhost:9092')
tp = TopicPartition('my-topic', 0)
consumer.offsets_for_times({tp: float('inf')})
# OverflowError: cannot convert float infinity to integer
```

## Fix

Wrap `int(ts)` in a try/except that catches `ValueError`, `TypeError`, and `OverflowError`, and re-raises as `ValueError` with a descriptive message — consistent with the existing negative-timestamp check.

### DIFF
--- a/kafka/consumer/group.py
+++ b/kafka/consumer/group.py
@@ -1228,7 +1228,12 @@ class KafkaConsumer:
         """
         timeout_ms = self.config['request_timeout_ms'] if timeout_ms is None else timeout_ms
         for tp, ts in timestamps.items():
-            timestamps[tp] = int(ts)
+            try:
+                timestamps[tp] = int(ts)
+            except (ValueError, TypeError, OverflowError):
+                raise ValueError(
+                    "The target timestamp for partition {} is {}. Timestamps "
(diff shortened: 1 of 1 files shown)


dpkp:
Why did or do you expect a ValueError in this case? The docstring only claims to raise ValueError for negative timestamps. I'm not sure how changing OverflowError to ValueError improves this api contract?

koteshyelamati:
Fair point, @dpkp — there's no documented contract that non-finite floats should raise `ValueError`, and `OverflowError` from `int(float('inf'))` is a perfectly reasonable Python error. I was motivated by consistency (negative timestamps raise `ValueError`, so non-finite floats raising a different exception type felt surprising), but you're right that the docstring doesn't promise that. Happy to close this if you'd prefer to leave the behavior as-is.

PR #3150, opened 2026-08-11
Fix JsonSerializer returning None and raising TypeError on deserialize

davexat:
## Problem

`JsonSerializer` is broken since the 3.0 serializer refactor (#3046):

- `JsonSerializer().serialize(topic, headers, data)` always returns `None`
- `JsonSerializer().deserialize(topic, headers, data)` raises
  `TypeError: the JSON object must be str, bytes or bytearray, not NoneType`

This silently corrupts messages produced with `value_serializer=JsonSerializer()` and crashes consumers using `value_deserializer=JsonSerializer()`.

## Root cause

`kafka/serializer/json.py` inherits from `Serializer, Deserializer`, whose abstract methods in `kafka/serializer/abstract.py` are no-ops (`pass`). The `super().serialize(...)` / `super().deserialize(...)` calls in `JsonSerializer` therefore return `None`. Before the 3.0 refactor, `JsonSerializer` inherited from `DefaultSerializer`, which actually performs the UTF-8 encode/decode.

## Fix

Restore delegation to `DefaultSerializer` by inheriting from it:

```python
class JsonSerializer(DefaultSerializer):
```

`serialize` now returns UTF-8 bytes and `deserialize` decodes bytes before `json.loads`.

## Tests

Added `test/test_serializer.py` (no serializer unit tests existed):

- `DefaultSerializer`: roundtrip (utf-8 / utf-16), bytes-like passthrough (bytes / bytearray / memoryview / None), `deserialize(None) -> None`, rejection of non-bytes input
- `JsonSerializer`: roundtrip dict -> bytes -> dict, `None` handling, deserializing plain JSON bytes

All tests pass with the fix; the `JsonSerializer` tests fail against the previous code (reproducing the bug).

## Verification

```
python -m pytest test/test_serializer.py -v
# 11 passed
```

### DIFF
--- a/kafka/serializer/json.py
+++ b/kafka/serializer/json.py
@@ -1,9 +1,9 @@
 import json
 
-from .abstract import Serializer, Deserializer
+from .default import DefaultSerializer
 
 
-class JsonSerializer(Serializer, Deserializer):
+class JsonSerializer(DefaultSerializer):
     def serialize(self, topic, headers, data):
         if data is None:
             return None
--- a/test/test_serializer.py
+++ b/test/test_serializer.py
@@ -1,0 +1,47 @@
+# pylint: skip-file
+
+import pytest
+
+from kafka.serializer import DefaultSerializer, JsonSerializer
+
+
+@pytest.mark.parametrize('encoding', ['utf-8', 'utf-16'])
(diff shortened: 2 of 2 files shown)


dpkp:
Thanks!

